Issue #14, opened 2023-05-26
I24 SSX : tidy up and update code


noemifrisina:
- Update to newer python version (some still using py2.7 syntax)
- A few functions are doubled ie. there's two versions of them doing the same thing with different variables
- Linting should be fixed

PR #16, opened 2023-05-31
Tidy up I24 serial code

noemifrisina:
- [x] Remove unused code from `setup_beamline` 
       (small note: `modechange` seems to be the change between goniometers. As I understand it, it's supposed to control a touch screen in the hutch that is currently broken. Keeping it in for the moment as beamline staff wants to get it working again. )
- [x] Remove any legacy code that is not used anymore
- [x] Remove unused edms/pvar files for unused chips
- [x] General tidying up - remove/fix paths to avoid breaking things on the beamline
- [x] Add script to set visit directory from the command line 
- [x] Add a couple more tests where possible

Closes #13 
Closes #14  (...hopefully)

PR #17, opened 2023-06-05
I24 ssx: determine detector

noemifrisina:
Add a small function to find which detector is in use.
Employing roughly the same logic used in GDA to check the detector and move it to correct positioning. There's probably a better way, but for now it does the job. 

Part of #14


DominicOram (changes requested):
Thanks, few suggestions:
1. Can you add the changes where you're using these functions into this PR too?
1. Can we return the detector class rather than a string? If we need the string put it as a param on the class
1. Given 2 can we add a return type to `get_detector_type`

noemifrisina:
Made the suggested changes, merging now because needed in #16

Issue #21, opened 2023-06-16
I24 SSX: avoid having to manually change the scripts anywhere.


noemifrisina:
Other than the visit setup and detector change, there are some other places where values are manually changed. While these are not modified as often, it should still be avoided. 

 - scalex, scaley, scalez and skew in chip manager. Occasionally we recalibrate the step size of the stages. When this is done we update the motor record in epics AND update these values. I guess scalex, scaley and scalez could be obtained from a PV (they would be 1/ (AXIS-PV.MRES) ). skew is the deviation from how perfectly orthogonal x and y are and this is not a PV.
 - More rarely: sign of Sx, Sy, Sz in chip manager. Only required if a different set of stages are mounted which are physically connected up in a different orientation
 - If different stages are mounted / the directionality of the stages is changed then this has to be reflected in (param_path)/motor_directions.txt

noemifrisina:
Add on to this: datasetsizei24() in chip_Collect

noemifrisina:
Detail comment from line 939 in cs_maker...

skew = -0.189  # same as scalex/y/z skew=th_z1 - th_z2

PR #22, opened 2023-06-22
I24 SSX tidy up: remove Sacla

### LINKED ISSUE #19: I24 ssx tidy up: remove sacla part of code
Only keep diamond relevant parts, remove anything specific to sacla. 

Temporarily keep using old scripts for sacla experiments, they will be added back in once there's a working deployment

noemifrisina:
Closes #19 

Remove the parts of the code and edms related to collections at sacla.


noemifrisina:
Sacla-related comment in `cs_maker` in Chip_Manager have been left as it has been listed by beamline staff as one of the functions that they modify manually at times - need to double check the values first and it's good to have the full comment. This change will be part of another PR (see #21 ).

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -891,7 +769,7 @@ def main(args):
         "-l",
         "--loc",
         type=str,
-        choices=["i24", "SACLA"],
+        choices=["i24"],
```
Tom-Willemsen, line 772: I guess this is what you meant when you said you've "left in a parser which isn't currently useful".
My instinct is remove this. It's not complicated so very easy to add back in later if it turns out to be needed. In my experience lots of code which "might" be useful later turns out to never be used.

If you're strongly in favour of keeping it then feel free to skip this comment.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -1164,8 +1151,6 @@ def block_check():
                 block_start_list = scrape_pvar_file("minichip_oxford.pvar")
             if chip_type == 10:
                 block_start_list = scrape_pvar_file("oxford6x6.pvar")
-            else:
-                block_start_list = scrape_pvar_file("sacla3_oxford.pvar")
             for entry in block_start_list:
```
Tom-Willemsen, line 1154: This is a change in behaviour (but the previous behaviour looks like it may have been a bug to me...)

- _Previously_ if chip type was anything other than `10`, the block_start_list would have been set to `sacla3_...`. 
- _With this change_ if chip type is 9 it'll be set to `minichip_oxford`, and in any other case it'll crash with an undefined variable.

I think instead of 

```python
if chip_type == 9:
    block_start_list = ...
if chip_type == 10:
    block_start_list = ...
```

we want:

```python
if chip_type == 9:
    block_start_list = ...
elif chip_type == 10:
    block_start_list = ...
else:
    raise ValueError("invalid chip type")  # Or whatever other error-handling is suitable.
```

In general: do we know which chip types correspond to SACLA experiments? If so can we remove all references to those chip types too? Or is that part of a follow-up ticket?

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: I don't have strong feelings about this, that's why I wanted a second opinion :) 
It's now been removed

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: I think you're right and it might have been a bug. Anyway, change has been made.
In general, at the moment they use the same chips here and at sacla. I have a list of chips in use, but which numbers they correspond to I'm still looking into. I guess it's part of the next issue to tackle (see #20 ) which will remove the unused chips left.

Tom-Willemsen (approved):

PR #25, opened 2023-07-13
Add dodal 

noemifrisina:
- Add dodal to dependencies 
- Install dodal in .venv
- Set up vscode workspace

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -156,7 +156,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           build-args: |
-            PIP_OPTIONS=-r dist/*.whl
+            PIP_OPTIONS=dist/*.whl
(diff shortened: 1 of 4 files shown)


.github/workflows/code.yml
```diff
@@ -156,7 +156,8 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           build-args: |
-            PIP_OPTIONS=-r lockfiles/requirements.txt dist/*.whl
+            PIP_OPTIONS=dist/*.whl
+          #  PIP_OPTIONS=-r lockfiles/requirements.txt dist/*.whl
```
DominicOram, line 160: Should: If we're going to leave the code in a comment can you put a link to the issue here?

DominicOram (changes requested):
Minor comment in code, tested by installing venv and confirming that `dodal.__file__` was pointed to local folder

DominicOram (approved):
Great, thanks!

PR #26, opened 2023-07-13
I24 SSX: Remove all manual changes

### LINKED ISSUE #21: I24 SSX: avoid having to manually change the scripts anywhere.

noemifrisina:
Closes #21 

- Motor positions are updated in the file
- Use a json file to set the scale x/y/z and skew in the coordinate system maker
- Add parameter files to .gitignore
- Use a nicer parser in chip_manager


Tom-Willemsen (commented):
Is there any parameter validation we can do? e.g. `Sz_x` I think should be either `-1` or `1` if I've understood it correctly, can we error if that's not the case?

Similarly type validation, what happens if a user accidentally puts their value as a string in the JSON rather than a number (easy mistake to make) - can we catch this case and give a sensible error?

Tom-Willemsen (approved):

PR #30, opened 2023-07-18
Tidy up logging

noemifrisina:


src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Mapping_py3v1.py
```diff
@@ -130,36 +126,24 @@ def convert_chip_to_hex(fid, chip_type):
...
-                        + 4 * "0",
+                        + 4 * "0"
                     )
                 )
-                logger.info("%s %s %s \n" % (name, i, right_list))
-                logger.info("%s %s\n" % (name, "".join(str(x) for x in right_list)))
-                logger.info("%s %s \n" % (name, line))
+                logger.info("%s %s \n" % (i, right_list))
+                logger.info("%s\n" % ("".join(str(x) for x in right_list)))
+                logger.info("%s \n" % line)
```
Tom-Willemsen, line 140: We shouldn't need trailing newlines on logging statements...?

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
```diff
@@ -147,9 +148,10 @@ def get_format(chip_type):
         b2b_vert = 0
         chip_format = [1, 1, 20, 20]
     else:
-        logger.warning("%s Unknown chip_type, %s, in fiducials" % (name, chip_type))
-        print("unknown chip type")
+        logger.error("Unknown chip_type, %s" % chip_type)
+        raise ValueError("Unknown chip_type")
```
Tom-Willemsen, line 152: Similar to a previous comment - pull out the common string and then use it in both the logging statement and the valueerror so that the valueerror has the context of which chip type was invalid.

No further comments on similar instances, please check for similar cases throughout PR.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
```diff
@@ -216,11 +218,9 @@ def pathli(l_in=[], way="typewriter", reverse=False):
                 for rep in range(25):
                     long_list.append(entry)
         else:
-            logger.warning("%s no known path, way =  %s" % (name, way))
-            print("no known path")
+            logger.warning("%s No known path, way =  %s" % (name, way))
```
Tom-Willemsen, line 221: In most places we got rid of the `name` prefix, should we get rid of it here too?

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
```diff
@@ -323,8 +322,7 @@ def write_file(
             prepumpexptime,
         ) = scrape_parameter_file(param_file_path)
     else:
-        logger.warning("%s Unknown location, %s" % (name, location))
-        print("Unknown location in write_file")
+        logger.warning("Unknown location, %s" % location)
```
Tom-Willemsen, line 325: Since it'll immediately fall over on the line below if it hits this case, this should probably be `logger.error` and maybe also a `raise ValueError(...)`.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
```diff
@@ -348,7 +346,7 @@ def write_file(
             line = "\t".join([xtal_name, str(x), str(y), "0.0", pres]) + "\n"
             g.write(line)
 
-    logger.info("%s" % name)
+    logger.info("Write file done")
```
Tom-Willemsen, line 349: Might be beneficial to add the context of _which_ file was written.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
```diff
@@ -388,10 +386,9 @@ def check_files(
         if full_fid.is_file():
             time_str = time.strftime("%Y%m%d_%H%M%S_")
             timestamp_fid = full_fid.parent / f"{time_str}_{chip_name}{full_fid.suffix}"
-            print("FIX ME")
-            # hack / fix
-            print("Already exists ... moving old file:", timestamp_fid)
-            logger.info("%s File Already Exists\n -->%s" % (name, timestamp_fid))
+            # FIXME hack / fix
+            logger.info("File Already Exists\n -->%s" % timestamp_fid)
```
Tom-Willemsen, line 390: Being a bit more explicit here would be nice. Something like:

> File /some/path/to/file already exists, Writing new file as /some/other/path/to/file instead

would be more helpful to users, I think.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -17,10 +20,10 @@
 # Register clicks and move chip stages
 def onMouse(event, x, y, flags, param):
     if event == cv.EVENT_LBUTTONUP:
-        print("Clicked X and Y", x, y)
+        logger.info("Clicked X and Y %s %s" % x, y)
```
Tom-Willemsen, line 23:
```suggestion
        logger.info("Clicked X and Y %s %s" % (x, y))
```

pretty sure this is required, or at least used by convention. Please check throughout PR, no further comments on similar cases.

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
```diff
@@ -192,18 +198,18 @@ def modechange(action):
...
     else:
-        print("Unknown action", action)
+        logger.debug("Unknown action: %s" % action)
     return 1
 
 
 def beamline(action, args_list=None):
-    print("\n***** Entering Beamline")
-    print("beamline - ", action)
+    logger.debug("\n***** Entering Beamline")
```
Tom-Willemsen, line 208: We probably don't need a leading newline.

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
```diff
@@ -216,9 +222,9 @@ def beamline(action, args_list=None):
     elif action == "quickshot":
         det_dist = args_list[0]
         caput(pv.det_z, det_dist)
-        print("Waiting on detector")
-        print(str(int(float(det_dist))))
-        print(str(int(float(caget(pv.det_z + ".RBV")))))
+        logger.info("Waiting on detector")
+        logger.debug("Detector distance: %s" % int(float(det_dist)))
```
Tom-Willemsen, line 226: doing an implicit `str(int(float(x)))` is a bit much for my taste.

Can't we just log the detector distance as-is?

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
```diff
@@ -230,36 +236,36 @@ def beamline(action, args_list=None):
...
     if args_list:
         for arg in args_list:
-            print(arg)
+            logger.debug("Argument: %s" % arg)
 
     # caput(pv.pilat_wavelength, caget(pv.dcm_lambda))
     caput(pv.pilat_detdist, caget(pv.det_z))
     caput(pv.pilat_filtertrasm, caget(pv.attn_match))
-    print("WARNING: Have you set beam X and Y?")
+    logger.warning("WARNING: Have you set beam X and Y?")
```
Tom-Willemsen, line 268: The logger will prefix this with `[WARN]` anyway so do we need the manual `WARNING: `?

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
```diff
@@ -363,21 +369,20 @@ def pilatus(action, args_list=None):
...
+    logger.debug("\n***** Entering Eiger")
+    logger.info("eiger - %s" % action)
     if args_list:
         for arg in args_list:
-            print(arg)
+            logger.debug("Argument: %s" % arg)
     # caput(pv.eiger_wavelength, caget(pv.dcm_lambda))
     caput(pv.eiger_detdist, str(float(caget(pv.det_z)) / 1000))
-    print("WARNING: Have you set header info?")
+    logger.warning("WARNING: Have you set header info?")
```
Tom-Willemsen, line 385: Similar to above. No further comments on similar cases, please check throughout.

Tom-Willemsen (changes requested):
I may have been a little heavy-handed with the review but logging is something that is quite important for debuggability!

Feel free to grab me if any comments don't make sense or you feel the scope is creeping too much on some of these comments.

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -62,19 +62,19 @@ def initialise_extruderi24(args=None):
...
-    logger.info("%s Initialsation complete" % name)
+    logger.info(4 * "DONE ")
+    logger.info("Initialsation complete")
 
 
 def moveto(args):
     place = args.place
     name = inspect.stack()[0][3]
-    logger.info("%s Move to %s" % (name, place))
+    logger.debug("Running %s " % name)
```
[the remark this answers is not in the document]
noemifrisina: Thanks, this looks good. I looked for a way to reduce duplication but after a few failed attempts I just left it, I'll try this

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -248,40 +234,38 @@ def run_extruderi24(args=None):
...
-        logger.info("%s Pilatus quickshot setup: filepath %s" % (name, filepath))
-        logger.info("%s Pilatus quickshot setup: filepath %s" % (name, filename))
-        logger.info(
-            "%s Pilatus quickshot setup: number of images %s" % (name, num_imgs)
-        )
-        logger.info("%s Pilatus quickshot setup: exposure time %s" % (name, exp_time))
+        logger.info("Pilatus quickshot setup: filepath %s" % filepath)
+        logger.info("Pilatus quickshot setup: filepath %s" % filename)
+        logger.info("Pilatus quickshot setup: number of images %s" % num_imgs)
+        logger.info("Pilatus quickshot setup: exposure time %s" % exp_time)
```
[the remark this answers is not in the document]
noemifrisina: Not really, you're right, changing them with %d and %f which are more correct. (For some reason it does work anyway - I guess there may be some casting done in the background)

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -387,67 +364,57 @@ def run_extruderi24(args=None):
                 i += 1
                 if int(caget(pv.ioc12_gp8)) != 0:
                     aborted = True
-                    logger.warning("%s Data Collection Aborted" % (name))
-                    print(50 * "ABORTED ")
+                    logger.warning("Data Collection Aborted")
+                    logger.info(20 * "ABORTED ")
```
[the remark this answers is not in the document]
noemifrisina: Not really an error condition, it's just logging that the user pressed the abort button. I agree it's awful and at some point the behaviour will change, but for the moment I just want some logging telling me that things were stopped on purpose by pressing abort

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -975,24 +908,15 @@ def check_dir(val):
...
     sqfact2 = np.sqrt(x2factor**2 + y2factor**2 + z2factor**2) / scaley
     sqfact3 = np.sqrt(x3factor**2 + y3factor**2 + z3factor**2) / scalez
-    print(sqfact1)
-    print(sqfact2)
-    print(sqfact3)
-    logger.info("%s %1.4f \n %1.4f \n %1.4f" % (name, sqfact1, sqfact2, sqfact3))
-    print("Long wait, please be patient")
-    logger.info("%s Long wait, please be patient" % (name))
+    logger.info("%1.4f \n %1.4f \n %1.4f" % (sqfact1, sqfact2, sqfact3))
+    logger.info("Long wait, please be patient")
```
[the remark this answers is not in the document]
noemifrisina: I have no idea, my best guess would be a few minutes from the one time I saw this done. However, from what I've been told this is not something a user would run, just the beamline staff once in awhile.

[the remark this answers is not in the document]
Tom-Willemsen: ok - if you have a good guess put that in (we can always correct it later if it's wrong), if not then feel free to leave this comment.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_StartUp_py3v1.py
noemifrisina: Fair enough. 
We found out while trying to test it that this actually does nothing when the file exists - which is why I removed the "moving old file" comment. However, it is misleading that I'm printing out the supposedly new filename so I changed that. There's a FIXME as a reminder to actually change the file name. 
As an aside, right now it's always called and checks/writes a bunch of unused files. Since it seems it the only use these files have is if the full mapping gets fixed, it's not something the users will look for. (It is in the list of things to fix and I'll need to set up a way so that this part of the code only gets called when needed)

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
noemifrisina: Ah yes, definitely a typo, good spot.

noemifrisina:
> I may have been a little heavy-handed with the review but logging is something that is quite important for debuggability!
> 
> Feel free to grab me if any comments don't make sense or you feel the scope is creeping too much on some of these comments.

The point of this was to at least try to make logging a little more sensible so a lot of comments are definitely appreciated, thank you! 
I haven't fully followed a couple and tried to explain my reasoning - let me know if they don't make sense

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
noemifrisina: The logger won't print the logging level on the terminal because I didn;t want to change how things were printed out too much from what they're used to. It will be in the .log file (along with timestamp and some other useful info) if something goes wrong and we need it for debugging. 

I admit I'm not super sure what to do with this particular warning message. Ideally, once deployment is tested we'll get the beam position from somewhere else (this is the very last bit of manual changes but it hasn't been changed for more than a year) and this message will be removed completely. For the moment I'm tempted to leave it with the extra "WARNING" just to make it more visible on the terminal. A second opinion is definitely welcomed!

PR #31, opened 2023-07-20
Set up a first deployment for testing I24 ssx

### LINKED ISSUE #27: Change edm to look at new scripts
edm screens should now run the scripts in mx_bluesky (using either relative paths or entry points)

### LINKED ISSUE #29: Set up deployment for I24
- Make sure all the edm screens point to the correct scripts/screens
- Make a first release and set up deployment to go in i24/software.

noemifrisina:
Closes #27 and closes #29  

- Have the edm point to the correct scripts/edms
- Have a bash script to run extruder/fixed target experiment

Set up two options for running, one using a bash script and the other entry points, one to be removed after testing and discussion with beamline staff.

### DIFF
--- a/deploy/deploy_edm_for_ssx.sh
+++ b/deploy/deploy_edm_for_ssx.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+
+# Deploy EDM screens for serial crystallography
+# Make a copy of them in edm/ and replace paths depending if in dev or beamline mode
+
+base=$( dirname "$( dirname "$0" )" )
+
+edm_build="$base/edm"
+
+if [ -d $edm_build ]; then
+    rm -rf $edm_build
+fi
+
+mkdir -p $edm_build
+
+edm_placeholder="EDM_LOCATION"
+ex_edm="$edm_build/EX-gui"
+ft_edm="$edm_build/FT-gui"
+
+mkdir $ex_edm
+mkdir $ft_edm
+
+scripts_placeholder="SCRIPTS_LOCATION"
+scripts_loc="$base/src/mx_bluesky/I24/serial"
+
+# Copy extruder
+cp $scripts_loc/extruder/EX-gui-edm/*.edl $ex_edm
+# Copy fixed target
+cp $scripts_loc/fixed_target/FT-gui-edm/*.edl $ft_edm
+# Different places because of edm with common name from microdrop alignment. Might change once checked.
+
+# Fix both edm and scripts paths
+for filename in $ex_edm/*.edl; do
+    sed -i "s+${edm_placeholder}+${ex_edm}+g" $filename     # Fix edm paths
+    sed -i "s+${scripts_placeholder}+${scripts_loc}+g" $filename    # Fix scripts paths
+done
+for filename in $ft_edm/*.edl; do
+    sed -i "s+${edm_placeholder}+${ft_edm}+g" $filename     # Fix edm paths
+    sed -i "s+${scripts_placeholder}+${scripts_loc}+g" $filename    # Fix scripts paths
+done
--- a/deploy/deploy_edm_for_ssx.sh
+++ b/deploy/deploy_edm_for_ssx.sh
@@ -3,7 +3,9 @@
 # Deploy EDM screens for serial crystallography
 # Make a copy of them in edm/ and replace paths depending if in dev or beamline mode
 
-base=$( dirname "$( dirname "$0" )" )
+current=$( realpath "$( dirname "$0" )" )
+base=$(dirname $current)
+echo $base
 
 if [[ -n "${BEAMLINE}" ]]; then
     edm_build="/dls_sw/$BEAMLINE/software/bluesky/edm_serial"
--- a/deploy/deploy_edm_for_ssx.sh
+++ b/deploy/deploy_edm_for_ssx.sh
@@ -37,13 +37,13 @@ cp $scripts_loc/fixed_target/FT-gui-edm/*.edl $ft_edm
 # Different places because of edm with common name from microdrop alignment. Might change once checked.
 
 # Fix both edm and scripts paths
+echo "Setting up screen for extruder"
 for filename in $ex_edm/*.edl; do
-    echo "Setting up screen for extruder"
     sed -i "s+${edm_placeholder}+${ex_edm}+g" $filename     # Fix edm paths
     sed -i "s+${scripts_placeholder}+${scripts_loc}+g" $filename    # Fix scripts paths
 done
+echo "Setting up screens for fixed target"
 for filename in $ft_edm/*.edl; do
-    echo "Setting up screens for fixed target"
     sed -i "s+${edm_placeholder}+${ft_edm}+g" $filename     # Fix edm paths
     sed -i "s+${scripts_placeholder}+${scripts_loc}+g" $filename    # Fix scripts paths
 done
--- a/deploy/deploy_edm_for_ssx.sh
+++ b/deploy/deploy_edm_for_ssx.sh
@@ -42,7 +42,7 @@ for filename in $ex_edm/*.edl; do
     sed -i "s+${edm_placeholder}+${ex_edm}+g" $filename     # Fix edm paths
     sed -i "s+${scripts_placeholder}+${scripts_loc}+g" $filename    # Fix scripts paths
 done
-echo "Setting up screens for fixed target"g
+echo "Setting up screens for fixed target"
 for filename in $ft_edm/*.edl; do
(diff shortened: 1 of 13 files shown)


src/mx_bluesky/I24/serial/run_ssx.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Start the edm screen relative to requested serial collection
+expt_type=${1:-FT}
+
+echo "Activate python environment before starting edm screen"
```
DominicOram, line 6: Should: This makes it sound like something the user needs to do, rather than something this script is doing. Would be better to print something like `Activating python environment...`

DominicOram (changes requested):
This looks like it should work but to call it done it needs some user facing documentation somewhere about how to run it and how to change visit etc. Also, because we have fixed paths in the EDMs it will make testing this really hard. I think https://github.com/DiamondLightSource/mx_bluesky/issues/32 would be a better solution, what do you think?

noemifrisina:
> This looks like it should work but to call it done it needs some user facing documentation somewhere about how to run it and how to change visit etc. Also, because we have fixed paths in the EDMs it will make testing this really hard. I think #32 would be a better solution, what do you think?


That sounds good, although I'm not sure right now how to get a placehoder location to work in the EDM - a question for controls I guess. In any case, it definitely would work better than the set up I have right now - which is a separate branch where I have the EDMs pointing somewhere locally. 

As far as documentation goes, would adding a section for serial to the wiki here work or should it go somewhere else?

DominicOram:
> > This looks like it should work but to call it done it needs some user facing documentation somewhere about how to run it and how to change visit etc. Also, because we have fixed paths in the EDMs it will make testing this really hard. I think #32 would be a better solution, what do you think?
> 
> That sounds good, although I'm not sure right now how to get a placehoder location to work in the EDM - a question for controls I guess. In any case, it definitely would work better than the set up I have right now - which is a separate branch where I have the EDMs pointing somewhere locally.
> 
> As far as documentation goes, would adding a section for serial to the wiki here work or should it go somewhere else?

I don't mean that the placeholder should work, the idea would be to never run the EDMs with the placeholder values but only the deployed ones that have the placeholders changed to dev/prod locations depending on where you deployed them. Yh, docs on here is fine

src/mx_bluesky/I24/serial/run_serial.py
```diff
@@ -0,0 +1,34 @@
...
+    return logger
+
+
+def run_extruder():
+    logger = setup_logging()
+    logger.info("Starting extruder edm screen.")
+    subprocess.run(
+        [
+            "edm",
+            "/dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/extruder/EX-gui-edm/DiamondExtruder-I24-py3v1.edl",
```
DominicOram, line 19: Should: This can just be a relative path now, right?

src/mx_bluesky/I24/serial/run_serial.py
```diff
@@ -0,0 +1,34 @@
...
+    logger.info("Edm screen closed.")
+
+
+def run_fixed_target():
+    logger = setup_logging()
+    logger.info("Starting extruder edm screen.")
+    subprocess.run(
+        [
+            "edm",
+            "/dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/DiamondChipI24-py3v1.edl",
```
DominicOram, line 31: Should: As above, should be relative

src/mx_bluesky/I24/serial/run_ssx.sh
```diff
@@ -0,0 +1,24 @@
...
+shopt -s nocasematch
+
+if [[ $expt_type == "FT" ]] || [[ $expt_type == "fixed-target" ]]
+then
+    echo "Starting fixed target edm screen."
+    edm /dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/DiamondChipI24-py3v1.edl
+elif [[ $expt_type == "EX" ]] || [[ $expt_type == "extruder" ]]
+then
+    echo "Starting extruder edm screen."
+    edm /dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/extruder/EX-gui-edm/DiamondExtruder-I24-py3v1.edl
```
DominicOram, line 18: Should: As above, paths can be relative

deploy/deploy_edm_for_ssx.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+
+# Deploy EDM screens for serial crystallography
+# Make a copy of them in edm/ and replace paths depending if in dev or beamline mode
+
+base=$( dirname "$( dirname "$0" )" )
```
DominicOram, line 6: Should: This means it's deploying relative to where the script is run from, which could vary. Better to deploy from where this script lives

DominicOram (changes requested):
Good start, thanks! Some comments in code. Additionally:
* Could you log a bit more in the deploy script, it wasn't clear if it passed/failed
* The paths being copied into the EDM are relative e.g. I've got `python ./src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py write_parameter_file` in one. I think we need to replace with absolute paths?
* In addition to documenting the user facing `run_serial.sh` scripts we now need to also document this deploy script, just a line or two in the wiki about how to run it and any gotchas

deploy/deploy_edm_for_ssx.sh
noemifrisina: Done

src/mx_bluesky/I24/serial/run_ssx.sh
DominicOram: Can we not do something like:
```
#!/bin/bash

# Start the edm screen relative to requested serial collection
expt_type=${1:-FT}

current=$( realpath "$( dirname "$0" )" )
cd $current
cd ../../../..

echo "Activating python environment..."
source .venv/bin/activate

shopt -s nocasematch

if [[ $expt_type == "FT" ]] || [[ $expt_type == "fixed-target" ]]
then
    echo "Starting fixed target edm screen."
    edm ./src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/DiamondChipI24-py3v1.edl
elif [[ $expt_type == "EX" ]] || [[ $expt_type == "extruder" ]]
then
    echo "Starting extruder edm screen."
    edm ./src/mx_bluesky/I24/serial/extruder/EX-gui-edm/DiamondExtruder-I24-py3v1.edl
else
    echo "No edm found for $expt_type."
fi

echo "Edm screen closed, deactivate python environment"
deactivate
```

src/mx_bluesky/I24/serial/run_ssx.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "Activating python environment..."
+source /dls_sw/i24/software/bluesky/mx_bluesky/.venv/bin/activate
+
+shopt -s nocasematch
+
+if [[ $expt_type == "FT" ]] || [[ $expt_type == "fixed-target" ]]
+then
+    echo "Starting fixed target edm screen."
+    edm /dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/DiamondChipI24-py3v1.edl
```
DominicOram, line 14: Should: This opens them in an editor, presumably we want `-x` to run them?

DominicOram: Should: The paths are also using the src version of the edms, shouldn't we be running the built version from the deploy script?

src/mx_bluesky/I24/serial/run_serial.py
```diff
@@ -0,0 +1,50 @@
...
+def run_fixed_target():
+    logger = setup_logging()
+    loc = get_location()
+    logger.info(f"Running on {loc}.")
+    edm_path = get_edm_path()
+    logger.info("Starting extruder edm screen...")
+    subprocess.run(
+        [
+            "edm",
+            edm_path / "FT-gui/DiamondChipI24-py3v1.edl",
```
DominicOram, line 47: Should: Shouldn't we running these rather than open them in an editor?

DominicOram (changes requested):
Thanks, paths look correct in deployed version now. Some comments in code

DominicOram (approved):
Great, thank you!

PR #38, opened 2023-08-24
Read current visit from file

### LINKED ISSUE #35: I24 ssx: read visit from a file
`set_visit_directory.sh` should read the visit from a set .txt file in  `/dls_sw/i24/etc`

noemifrisina:
Closes #35 

Instead of passing value from the command line, have a file that can be modified by beamline staff and read visit from there.


Tom-Willemsen (approved):
I assume that the file currently not existing on the filesystem is expected.

noemifrisina:
Yep, I'll create it when testing

PR #40, opened 2023-09-05
Add a timeout to the data collection

noemifrisina:
Temporarily add a timeout to the data collection in case something goes wrong and it gets stuck. During the last testing there were a few issues that caused the data collection to fail, but the script was not getting out of the while loop and just kept going because conditions on the PVs were not met. It also could not be easily killed from the terminal.
Needs a better solution for the future but this should allow for a slightly safer testing of the deployment. 

Timeout is calculated from the number of images to be collected and the exposure time plus a constant in seconds. Probably overdoing it with the chip, but better be safe since I'm not considering the pump_repeat in it.

### DIFF
--- a/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
+++ b/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
@@ -343,48 +343,46 @@ def run_extruderi24(args=None):
 
     aborted = False
     timeout_time = time.time() + int(num_imgs) * float(exp_time) + 10
-    while True:
-        if int(caget(pv.ioc12_gp8)) == 0:  # ioc12_gp8 is the ABORT button
-            caput(pv.zebra1_pc_arm, 1)
-            sleep(gate_start)
-            i = 0
-            text_list = ["|", "/", "-", "\\"]
-            while True:
-                line_of_text = "\r\t\t\t Waiting   " + 30 * ("%s" % text_list[i % 4])
-                flush_print(line_of_text)
-                sleep(0.5)
-                i += 1
-                if int(caget(pv.ioc12_gp8)) != 0:
-                    aborted = True
-                    logger.warning("Data Collection Aborted")
-                    if det_type == "pilatus":
-                        caput(pv.pilat_acquire, 0)
-                    elif det_type == "eiger":
-                        caput(pv.eiger_acquire, 0)
-                    sleep(1.0)
-                    break
-                elif int(caget(pv.zebra1_pc_arm_out)) != 1:
-                    # As soon as the zebra1_pc_arm_out is not 1 anymore, exit.
-                    # Epics checks the geobrick and updates this PV once the collection is done.
-                    logger.info("----> Zebra disarmed  <----")
-                    break
-                elif time.time() >= timeout_time:
-                    logger.warning(
-                        """
-                        Something went wrong and data collection timed out. Aborting.
+
+    if int(caget(pv.ioc12_gp8)) == 0:  # ioc12_gp8 is the ABORT button
+        caput(pv.zebra1_pc_arm, 1)
+        sleep(gate_start)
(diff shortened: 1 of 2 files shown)


src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -342,6 +342,7 @@ def run_extruderi24(args=None):
         call_nexgen(None, start_time, param_file_tuple, "extruder")
 
     aborted = False
+    timeout = time.time() + int(num_imgs) * float(exp_time) + 1
```
Tom-Willemsen, line 345: As discussed in person: Not sure the `+1` on the end is sufficient here given that just the sleeps in the inner loop account for `1.5`s, before you even count any time for acquisition or `caput`s or ...

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -365,6 +366,18 @@ def run_extruderi24(args=None):
...
+                        """
+                        Something went wrong and data collection timed out. Aborting.
+                    """
+                    )
+                    if det_type == "pilatus":
+                        caput(pv.pilat_acquire, 0)
+                    elif det_type == "eiger":
+                        caput(pv.eiger_acquire, 0)
+                    sleep(1.0)
+                    break
```
Tom-Willemsen, line 380: As discussed in person: needs to break out of outer loop too

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
Tom-Willemsen: Can you add a comment on the appropriate line in the inner loop which *actually* waits for the acquisition to complete (in a way that presumably depends on `num_imgs` and `exp_time`)?

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -365,6 +366,18 @@ def run_extruderi24(args=None):
...
+                    logger.warning(
+                        """
+                        Something went wrong and data collection timed out. Aborting.
+                    """
+                    )
+                    if det_type == "pilatus":
+                        caput(pv.pilat_acquire, 0)
+                    elif det_type == "eiger":
+                        caput(pv.eiger_acquire, 0)
+                    sleep(1.0)
```
Tom-Willemsen, line 379: I don't understand why this sleep would be necessary if we're aborting anyway? If it is necessary can you add a comment explaining why, if it isn't then can you remove it?

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
Tom-Willemsen: `timeout` sounds like a duration rather than an absolute time. Maybe `timeout_time` would be clearer?

Tom-Willemsen (changes requested):
Most comments as discussed in person and likely apply to both files (only commented on the first as they're very similar).

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: This is in the list of things to ask Robin once he's back because I do not know. For that matter I also don't know why the abort button that controls that PV shouldn't be used so I'm not touching that part for now...

Tom-Willemsen (approved):
There is still some extra duplication that's been added here between the `timeout` and `abort` cases. But if you want to fix that in a later pull request that's fine.

PR #41, opened 2023-09-12
I24 hotfixes

noemifrisina:
Fixes made during/in the aftermath of testing on 12/09/23

- [x] Commit missing mkdir
- [x] Change dodal to dls-dodal
- [x] Update scale/skew values for cs_maker
- [x] Add parameter files to gitignore
- [x] ~~Fix nudges in the chip edm screen, they are currently upside down~~ (do in separate PR)
- [x] ~~Beam X/Y in moveonclick is hardcoded and needs to be changed every time, grab values from GDA so they are current~~ (do in separate PR)
- [x] ~~Get all zoom levels to work properly~~ (do in separate PR)
- [x] Have startup.run in wrirte_parameter_file to only run id map_type == "2"
- [x] Add new Zebra Status box to chip edm
- [x] ~~Fix permission issue for logfiles (right now users don't have permission to write them)~~ (do in separate PR)


src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -166,8 +166,9 @@ def write_parameter_file(param_path: Path | str = PARAM_FILE_PATH_FT):
     logger.info("prepumpexptime: %s" % prepumpexptime)
     logger.info("detector type: %s" % det_type)
 
-    logger.debug("Running start up now.")
-    startup.run()
+    if map_type == "2":
+        logger.debug("Running start up now.")
+        startup.run()
```
DominicOram, line 171: Should: Why do we only want this run when `map_type == 2`? Can we add a comment?

DominicOram (changes requested):
Minor comment in code. Can you also add issues for:
* Fixing the nudge in the EDM
* Fix permissions issue for log files
* Pulling in any updates from the live version

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: Sure. This step creates files only needed when using full mapping, which is map_type 2. 
Comment added.

DominicOram (approved):
Great, thank you!

Issue #46, opened 2023-09-15
I24: Fix nudges/windows edm screen


noemifrisina:
The nudge/windows buttons on the edm screen for fixed target look to be upside down and should be fixed.

Once that's done, it would be useful to find an alternative to them talking directly to the PMAC_STRING PV.

noemifrisina:
Tested today, now in the correct directions.

PR #48, opened 2023-09-18
Pull in latest updates

### LINKED ISSUE #47: I24: Update to include any recent live edits
Compare and pull in any updates on the live version of the scripts/edms.

noemifrisina:
Closes #47 
Check differences with live script and update everything.

Summary of live updates:
- For some reason the temporary eiger/pilatus hack was missing from extruder script (probably because it hadn't been used in a long time and never before with eiger). Added it now.
- In the EDMs for the chip, the nudges/windows were upside down ( See #46 ) . Live version had updated values in what should be the correct directions. Adopting those.

Other changes:
- Add back missing caput for pump_repeat values in PumpProbe
- Tidy up pilatus filename hack
- Update .gitignore with correct names


src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -278,6 +279,24 @@ def run_extruderi24(args=None):
...
+
+        logger.warning(
+            """TEMPORARY HACK!
+            Running a Single image pilatus data collection to create directory."""
+        )
+        num_imgs = 1
+        sup.pilatus("quickshot-internaltrig", [filepath, filename, num_imgs, exp_time])
+        logger.debug("Sleep 2s waiting for pilatus to arm")
+        sleep(2)
+        sleep(0.5)
```
Tom-Willemsen, line 292: !?

Either:
```suggestion
        sleep(2.5)
```
Or
```suggestion
        # Some comment explaining a justification for a 2s wait
        sleep(2)
        # Some comment explaining a different justifiction for a further 0.5s wait
        sleep(0.5)
```

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -278,6 +279,24 @@ def run_extruderi24(args=None):
             sup.zebra1("quickshot", [gate_start, gate_width])
 
     elif det_type == "eiger":
+        logger.info("Using Eiger detector")
+
+        logger.warning(
+            """TEMPORARY HACK!
+            Running a Single image pilatus data collection to create directory."""
+        )
+        num_imgs = 1
```
Tom-Willemsen, line 288: Setting `num_imgs` to `1` here means that the `eiger` collection later on will only ever take one image, I think. Was that intended? It seems wrong. If it *was* intended it needs a comment.

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -278,6 +279,24 @@ def run_extruderi24(args=None):
             sup.zebra1("quickshot", [gate_start, gate_width])
 
     elif det_type == "eiger":
+        logger.info("Using Eiger detector")
+
+        logger.warning(
+            """TEMPORARY HACK!
+            Running a Single image pilatus data collection to create directory."""
```
Tom-Willemsen, line 286: Can we link https://github.com/DiamondLightSource/mx_bluesky/issues/45 , either from this logging statement or in a source code comment? So that we know what this hack refers to and what the issue is to get it fixed properly.

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
```diff
@@ -14,7 +14,7 @@ class UnknownDetectorType(Exception):
 
 def get_detector_type() -> Detector:
     det_y = caget(pv.det_y)
-    if float(det_y) > Eiger.det_y_threshold:
+    if float(det_y) < Eiger.det_y_threshold:
```
Tom-Willemsen, line 17: To avoid needing this kind of swap in future, maybe it would be better to define a `tolerance` around the target position, rather than absolute thresholds? (feel free to kick this into a separate issue if you prefer)

tests/I24/serial/fixed-target/test_chip_manager.py
```diff
@@ -168,3 +169,12 @@ def test_arg_parser_runs_define_current_chip_function_as_expected(
...
     mock_define_current_chip.assert_called_once_with("chip_id")
+
+
+@patch("mx_bluesky.I24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.caput")
+@patch("mx_bluesky.I24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.caget")
+def test_pumpprobe_calc(fake_caget, fake_caput):
+    fake_caget.side_effect = [0.01, 0.005]
+    pumpprobe_calc()
+    assert fake_caget.call_count == 2
+    assert fake_caput.call_count == 5
```
Tom-Willemsen, line 180: Are there any more useful assertions we could do here?

You can assert that calls had specific arguments for example if you're trying to prove that certain PVs were read and written to.

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Yeah, good spot. No idea why two separate sleeps, probably simply because they were added at different times. I didn't even notice it tbh, just grabbed the missing code.

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Not intended, they were supposed to have different names but it probably got lost somewhere in previous changes. I just renamed the variable.

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Added a comment

src/mx_bluesky/I24/serial/setup_beamline/setup_beamline.py
noemifrisina: Not changing this at this time as it's doing the same as GDA and still haven't worked out the details of what to do for the detector (see #50 ).  Also, I honestly have no idea what a useful tolerance for this case should be... 
I had to swap because I accidentally got the wrong values from GDA when I set this up and had the detectors positions the wrong way around, hopefully I shouldn't need to swap it again before moving on to #50

tests/I24/serial/fixed-target/test_chip_manager.py
noemifrisina: I mostly wanted to check that the number of caput was correct since this will show on the screen. In any case, I'd be happy to improve on the test, but not sure how to do that. Do you have any suggestions?

Tom-Willemsen: I was thinking you should be able to assert that the `caput`s have useful values.

Something along the lines of:

```python
assert fake_caput.has_calls(
    Call("some-pv-name", value=123),
    Call("some-other-pv-name", value=456),
)
```

To not only check the *number* of caputs but also that they have useful values calculated. But happy for you to leave this for now.

Tom-Willemsen (approved):

PR #55, opened 2023-10-27
I24: get the beam centre from the 

### LINKED ISSUE #42: I24 ssx: get the beam centre from file
When setting the fiducials (`i24ssx_moveonclick.py`), beam X/Y hardcoded and needs to be changed every time. Instead of this, we should be grabbing the values from GDA so they are current

noemifrisina:
Fixes #42 

Instead of using hard coded values that need to be updated manually, get the beam centre from `display.configurations` in GDA using dodal devices.
 
Link to dodal PR: [PR220](https://github.com/DiamondLightSource/dodal/pull/220)


src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -3,24 +3,62 @@
...
+    oav_params = OAVParameters("xrayCentring", **oav_config)
+
+    return _get_beam_centre(oav, oav_params)
+
+
 # Register clicks and move chip stages
 def onMouse(event, x, y, flags, param):
-    if event == cv.EVENT_LBUTTONUP:
+    beamX, beamY = get_beam_centre_from_oav()
+    if event == cv.EVENT_LBUTTONUP:  # cv.EVENT_LBUTTONUP is an int! (=4)
```
DominicOram, line 61: Should: What does this comment mean? I'm not sure what it adds?

DominicOram (changes requested):
See comment on https://github.com/DiamondLightSource/dodal/pull/220#pullrequestreview-1710895959. I also think that the tests are testing slightly the wrong things. We don't need tests that confirm `OAVParameters` is reading the files and doing the calculations correctly, it's up to `dodal` to test that. We instead should mock out the function on `OAVParamaters` to give us a specific beam position and test that the ellipse gets drawn in the right place and the move on click gets calculated correctly.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
noemifrisina: To the code nothing, a note from trying to test this I forgot to remove

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -31,9 +68,92 @@ def onMouse(event, x, y, flags, param):
...
+        1,
+    )
+    cv.imshow("OAV1view", frame)
+
+
+def start_viewer(oav1: str = OAV1_CAM):
     # Create a video caputure from OAV1
-    cap = cv.VideoCapture("http://bl24i-di-serv-01.diamond.ac.uk:8080/OAV1.mjpg.mjpg")
+    cap = cv.VideoCapture(oav1)
+    # cap = cv.VideoCapture("http://bl24i-di-serv-01.diamond.ac.uk:8080/OAV1.mjpg.mjpg")
```
DominicOram, line 156: Should: Can you remove commented out code please?

tests/I24/serial/fixed_target/test_moveonclick.py
```diff
@@ -0,0 +1,80 @@
...
+def test_onMouse__gets_beam_position_and_sends_correct_str(fake_beam_pos, fake_caput):
+    fake_beam_pos.side_effect = [(15, 10)]
+    onMouse(cv.EVENT_LBUTTONUP, 0, 0, "", "")
+    assert fake_caput.call_count == 2
+    fake_caput.assert_has_calls(
+        [
+            call(ANY, "#1J:-90"),
+            call(ANY, "#2J:60"),
+        ]
+    )
```
DominicOram, line 44: Should: I think it would be good to parametrize this into `(beam_pos, expected_J1, expected_J2)` and have half a dozen or so values

DominicOram (changes requested):
Nice, great to see things incrementally getting cleaner and better tested. Some comments in code.

DominicOram (approved):
Great, thank you!

PR #56, opened 2023-10-30
Remove python 3.8

noemifrisina:


.github/workflows/code.yml
```diff
@@ -35,12 +35,12 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-latest"] # can add windows-latest, macos-latest
-        python: ["3.9", "3.10", "3.11"]
+        python: ["3.10", "3.11"]
```
DominicOram, line 38: Should: I thought this was remove 3.8, why is this removing 3.9?

DominicOram (changes requested):
Looks good, other than removing too much

.github/workflows/code.yml
noemifrisina: It's not really, 3.9 has taken the place of 3.8 as the non-editable version for testing (a few lines down) so it shouldn't be also in the list of editable installs.

DominicOram: You're right, sorry

DominicOram (approved):

PR #58, opened 2023-11-10
Choose the detector to use and move the detector stage

### LINKED ISSUE #50: I24: Choose detector in use
Right now the detector in use is determined by reading the position of the detector stage (PV `BL24I-EA-DET-01:Y`), but the actual motion is done somewhere else - usually from gda. It would be good if the beamline scientist was able to change it directly from here.

This could be done with a small plan using the device in https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/i24/I24_detector_motion.py ?

noemifrisina:
Closes #50 

Adds a small plan to move the detector stage to the correct position depending on the requested detector to use. 
Also add the edm screens for fixed-target and extruder to trigger it.


src/mx_bluesky/I24/serial/setup_beamline/setup_detector.py
```diff
@@ -0,0 +1,85 @@
...
+        detector_stage.y,
+        target,
+        wait=True,
+    )
+
+
+def setup_detector_stage(detector_stage: DetectorMotion, expt_type: str):
+    # Grab the correct PV depending on experiment
+    # Its value is set with MUX on edm screen
+    det_type = pv.me14e_gp101 if expt_type == "fixed-target" else pv.ioc12_gp15
```
dperl-dls, line 60: Should: can we use the SSXType enum for this?

src/mx_bluesky/I24/serial/setup_beamline/setup_detector.py
```diff
@@ -0,0 +1,85 @@
...
+    log.config(logfile)
+
+
+class UnknownDetectorType(Exception):
+    pass
+
+
+def get_detector_type() -> Detector:
+    det_y = caget(pv.det_y)
+    # Note to self, I should also be able to use detmotion for this too!
```
dperl-dls, line 36: Should: please expand into an issue and link it so that someone else could fix this, or remove

src/mx_bluesky/I24/serial/setup_beamline/setup_detector.py
```diff
@@ -0,0 +1,85 @@
...
+    setup_logging()
+    RE = RunEngine()
+    # Use dodal device for move
+    detector_stage = i24.detector_motion()
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "expt",
+        type=str,
+        choices=["extruder", "fixed-target"],
```
dperl-dls, line 80: you should be able to parse directly to SSXType enum here if the string choices match the enum values

dperl-dls (changes requested):
Looks fine, just a couple comments

dperl-dls (approved):
Great, thank you! I realise my comment wasn't very clear, so I made a small change to show what I meant. Please review it and make sure you're ok with it before merging

Issue #44, opened 2023-09-14
I24 SSX: Take into account zoom in click to move


DominicOram:
Currently the click to move has a hardcoded scale factor for pixels to microns. We should instead get the scale from the OAVParameters and we should make sure it's updated on zoom level

DominicOram:
Should probably do https://github.com/DiamondLightSource/mx_bluesky/issues/42 first

DominicOram:
We may also need to do https://github.com/DiamondLightSource/mx_bluesky/issues/54 first

noemifrisina:
OAVParameters related issue from testing on 14/11/23

Viewer opens fine but it gets shut off after a bit with the following error:

```Showing camera feed. Press escape to close
Clicked X and Y 827 266
Moving X and Y 1134 756
Clicked X and Y 684 467
Moving X and Y 276 -450
Clicked X and Y 528 596
Moving X and Y -660 -1224
Traceback (most recent call last):
  File "/dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py", line 59, in onMouse
  File "/dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py", line 54, in get_beam_centre_from_oav
  File "/dls_sw/i24/software/bluesky/mx_bluesky/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py", line 38, in _get_beam_centre
  File "/dls_sw/i24/software/bluesky/mx_bluesky/.venv/lib/python3.10/site-packages/bluesky/run_engine.py", line 364, in __init__
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/asyncio/events.py", line 783, in new_event_loop
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/asyncio/events.py", line 673, in new_event_loop
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/asyncio/unix_events.py", line 64, in __init__
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.10/lib/python3.10/asyncio/selector_events.py", line 56, in __init__
[...]

PR #59, opened 2023-11-14
I24 ssx test hotfixes

noemifrisina:
- [x] Wrong PV name
- [x] Set up and enum or something to look up what detector is requested. The choice on the edm will set the value to 0/1 instead of eiger/pilatus ( -> follow up to this: set up something similar for map_type and chip_type)
- [x] Temporary fix for log permission issue
- [x] OAV only uses crosshair from zoom 1.0, on I24 this is not updated beyond this line, so simply set that level for moveonclick

For Viewer/OAVParams issue see #44


dperl-dls (approved):
Looks good to me

PR #60, opened 2023-11-23
Use new OAV from dodal for moveonclick

noemifrisina:
Use the new OAV parameters definition, which has been changed so that the ones inherent to the OAV are now decoupled from the context. 

See also [Dodal#224](https://github.com/DiamondLightSource/dodal/pull/239)

### DIFF
--- a/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
+++ b/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
@@ -3,13 +3,15 @@ Move on click gui for fixed targets at I24
 Robin Owen 12 Jan 2021
 """
 import logging
-from typing import Dict
 
+import bluesky.plan_stubs as bps
 import cv2 as cv
-from dodal.devices.oav.oav_parameters import OAVParameters
+from bluesky.run_engine import RunEngine
+from dodal.beamlines import i24
+from dodal.devices.oav.oav_detector import OAV
 
 from mx_bluesky.I24.serial.fixed_target import i24ssx_Chip_Manager_py3v1 as manager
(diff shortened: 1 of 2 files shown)


src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -19,26 +21,28 @@
...
 
     Args:
-        oav_params (OAVParamters): the OAV parameters.
+        oav (OAV): the OAV device.
     """
     # Set to 1.0, as this is the only value that is updated in display config
     # on the beamline (all other beam positions will be invalid)
-    beamX, beamY = oav_params.get_beam_position_from_zoom(1.0)
-    return beamX, beamY
+    yield from bps.abs_set(oav.zoom_controller, "1.0", wait=True)
```
DominicOram, line 32: Should: Do we want t change zoom here? I thought we just wanted to get the beam position whatever the current zoom is?

DominicOram (changes requested):
Thanks, comment in code.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
noemifrisina: Not really?
Or well, we may want to change the zoom at a later time (we'll probably need to when running the fiducials bit) but for getting the beam position we need it at 1.0. It seems that on I24 they only use the beam positions for zoom level 1.0, whatever the actual zoom level used, and all the other values in the display_config file haven't been updated or checked in a long time...

DominicOram (approved):
Looks good, thank you!

PR #63, opened 2023-11-24
Tidy up MUX options

### LINKED ISSUE #61: I24 ssx: Tidy up mux options from edm
Use Enums to define the MUX options in the fixed target edm screens. 
These options are put to ME14E general purpose PVs for Map Type, Chip Type and Pump Probe as int values, which makes the code a bit confusing for those that don't know which number corresponds to which option.

Also, adding the setup_detector button means having this MUX option on that variable too.

noemifrisina:
When some option is chosen from a MUX on the edm screen, the relative PV value is set to an enum, for which we need to remember the type of chip/map/pp setting it corresponds to all through the code. This is just trying to tidy it up so that it's a bit more clear in which case we are operating.

Also, these general purpose have in some places been set to strings, which can causes confusion - in particular the ones relative to the detector in use, so right now instead of the PV, the detector stage position is used to get the value.

Closes #61


src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -71,21 +71,21 @@ def moveto(args):
     place = args.place
     logger.info("Move to: %s" % place)
 
-    det_type = caget(pv.ioc12_gp15)
+    det_type = get_detector_type()
 
     if place == "laseron":
-        if det_type == "pilatus":
+        if det_type.name == "pilatus":
```
DominicOram, line 77: Should: Rather than check names here could we not do:
```python
from mx_bluesky.I24.serial.setup_beamline.pv_abstract import Pilatus
if det_type == Pilatus:
   ....
```

src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
```diff
@@ -0,0 +1,25 @@
...
+
+class PumpProbeSetting(IntEnum):
+    NoPP = 0
+    Short1 = 1
+    Short2 = 2
+    Repeat1 = 3
+    Repeat2 = 4
+    Repeat3 = 5
+    Repeat5 = 6
+    Repeat10 = 7
```
DominicOram, line 25: Should: The order on these has to match the order on the EDM/IOC, correct? We should have a comment here saying this otherwise people may change it without realising

DominicOram (changes requested):
Some minor comments in code, I didn't check that you were correct on which ones you've replaced them with for all of them but they generally look about right

src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
```diff
@@ -1,3 +1,10 @@
+"""
+Define a mapping between the edm screens/IOC enum inputs fro the general purpose PVs and
```
DominicOram, line 2: Nit: typo on `for`

DominicOram (approved):
Minor typo in code, otherwise good, thank you!

PR #66, opened 2023-12-01
Add checker pattern to pump probe for I24 serial

### LINKED ISSUE #65: I24ssx: Add Checker Pattern
Add "Checker pattern" choice to edm screen for pump probe in FT and the bit of code that looks it up and send information to the pmac in `load_motion_program_data` if pump_repeat is not None. 

```
if int(caget(pv.me14e_gp111)) == 1:
            print('CHECKER PATTERN IS ENABLED')
            lg.info('%s CHECKER PATTERN ENABLED'%name)
            caput(pv.me14e_pmac_str, 'P1439=1')
```

Also add the value of the PV (`ME14E-MO-IOC-01:GP111`) it sets to the parameter file.

noemifrisina:
Closes #65 

Adds new setting to to the pump probe options, which sends a command to the pmac to only have the pump_repeat applied every other city block.

### DIFF
--- a/src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/pumpprobe-py3v1.edl
+++ b/src/mx_bluesky/I24/serial/fixed_target/FT-gui-edm/pumpprobe-py3v1.edl
@@ -3,8 +3,8 @@ beginScreenProperties
 major 4
 minor 0
 release 1
-x 1570
-y 705
+x 3174
+y 381
 w 900
 h 700
 font "arial-medium-r-18.0"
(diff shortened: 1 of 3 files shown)


DominicOram:
Yes, it's small, which means that it's a really easy, small ophyd device. It seems like we might as well add this in ophyd now. Otherwise we'll end up adding more functionality in the old way of doing it whilst also trying to move to the new way of doing it, seems like wasted effort?

dperl-dls (approved):
Seems fine

Issue #54, opened 2023-10-27
I24: Move to click should use ophyd chip stages


DominicOram:
Currently move to click sends commands directly to the `pmac` to do a jog. There are EPICS motor records that we can use instead, we should use this with an ophyd device.

## Acceptance Criteria
* The move to click code uses an ophyd device to move the chip stages

PR #67, opened 2023-12-04
Use ophyd device for pmac in moveonclick

noemifrisina:
Set the pmac_string using an ophyd device in moveonclick.
Part of #54 

Needs Dodal [#254](https://github.com/DiamondLightSource/dodal/pull/254)

### DIFF
--- a/tests/I24/serial/fixed_target/test_moveonclick.py
+++ b/tests/I24/serial/fixed_target/test_moveonclick.py
@@ -36,13 +36,6 @@ def test_onMouse_gets_beam_position_and_sends_correct_str(
             call.set(expected_2J),
         ]
     )
-    # assert fake_caput.call_count == 2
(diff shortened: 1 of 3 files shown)


tests/I24/serial/fixed_target/test_moveonclick.py
```diff
@@ -18,24 +18,31 @@
...
+            call.set(expected_2J),
         ]
     )
+    # assert fake_caput.call_count == 2
+    # fake_caput.assert_has_calls(
+    #     [
+    #         call(ANY, expected_1J),
+    #         call(ANY, expected_2J),
+    #     ]
+    # )
```
DominicOram, line 45: Should: Remove commented out code

DominicOram (dismissed):
This looks fine, other than the comment in code. Obviously just a stepping stone to actually using the motor record

PR #69, opened 2023-12-07
Use ophyd device to move chip stage in movetoclick

### LINKED ISSUE #54: I24: Move to click should use ophyd chip stages

noemifrisina:
Closes #54 

Needs [Dodal#259](https://github.com/DiamondLightSource/dodal/pull/259)


src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -609,47 +618,53 @@ def moveto(place: str = "origin"):
...
-            caput(pv.me14e_stage_y, 25.40)
+        if place == Fiducials.origin:
+            pmac.x.move(0.0, wait=True)
+            pmac.y.move(0.0, wait=True)
+        if place == Fiducials.fid1:
+            pmac.x.move(25.40, wait=True)
+            pmac.y.move(0.0, wait=True)
+        if place == Fiducials.fid2:
+            pmac.x.move(0.0, wait=True)
+            pmac.y.move(25.40, wait=True)
```
DominicOram, line 653: Should: Could we not simplify this all down to something like:
```python
if chip_type == ChipType.Oxford or chip_type == ChipType.Minichip or chip_type == ChipType.Custom:
    chip_size = 25.4
elif chip_type == ChipType.OxfordInner:
    chip_size = 24.6
else:
   logger.warning("Unknown chip_type move")
   return

def move_xy(x, y):
    pmac.x.move(x, wait=True)
    pmac.y.move(y, wait=True)

if place == Fiducials.origin:
   move_xy(0,0)
if place == Fiducials.fid1:
   move_xy(chip_size,0)
if place == Fiducials.fid2:
   move_xy(0, chip_size)
```
Or even better, put the chip size into the enum itself

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -609,47 +618,53 @@ def moveto(place: str = "origin"):
...
 
     else:
         logger.warning("Unknown chip_type move")
+    return
```
DominicOram, line 657: Should: This return doesn't do anything?

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -187,7 +186,7 @@ def start_viewer(oav1: str = OAV1_CAM):
         if k == 117:  # U
             pmac.pmac_string.set("#3J:-1000")
         if k == 112:  # P
-            pmac.pmac_string.set("#3J:1000")
+            pmac.pmac_string.set("#3J:-1000")
```
DominicOram, line 189: Must: this makes it the same as `U`, if we need to switch direction can we do both and in a new PR?

DominicOram (changes requested):
Thank you, some comments in code

src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
noemifrisina: Ah this was a typo.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: Went with the enum option, felt a bit tidier

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -604,57 +604,31 @@ def load_full_map(fullmap_path: Path | str = FULLMAP_PATH):
...
         logger.warning("Unknown chip_type move")
-    return
+        return
+
+    logger.info(f"{str(ChipType(chip_type))} Move")
+    chip_move = ChipType(chip_type).get_approx_chip_size()
+
+    def move_xy(x, y):
+        pmac.x.move(x)
+        pmac.y.move(y)
```
DominicOram, line 624: Should: The default args to `move` have `wait=True` so this is the same behaviour as before. We also probably do want to wait on them to complete at some point so we should do something like:
```python
move_status = pmac.x.move(x, wait=False)
move_status &= pmac.y.move(y, wait=False)
move_status.wait()
```

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -675,17 +649,17 @@ def moveto_preset(place: str, pmac: PMAC = None):
...
         caput(pv.bl_mp_select, "In")
 
     elif place == "microdrop_position":
         logger.info("microdrop align position")
-        pmac.x.move(6.0, wait=True)
-        pmac.y.move(-7.8, wait=True)
-        pmac.z.move(0.0, wait=True)
+        pmac.x.move(6.0)
+        pmac.y.move(-7.8)
+        pmac.z.move(0.0)
```
DominicOram, line 662: Should: As above, this is still waiting for all the axes after each move as wait defaults to True

tests/I24/serial/fixed_target/test_chip_manager.py
```diff
@@ -29,40 +31,49 @@
...
+def test_moveto_oxford_origin(fake_caget: MagicMock, fake_pmac: PMAC):
     fake_caget.return_value = 0
     fake_pmac.x = MagicMock()
     fake_pmac.y = MagicMock()
     moveto(Fiducials.origin, fake_pmac)
     assert fake_caget.call_count == 1
-    fake_pmac.x.assert_has_calls([call.move(0.0, wait=True)])
-    fake_pmac.y.assert_has_calls([call.move(0.0, wait=True)])
+    fake_pmac.x.assert_has_calls([call.move(0.0)])
+    fake_pmac.y.assert_has_calls([call.move(0.0)])
```
DominicOram, line 46: Should: The fact you need to set `x` and `y` here is suspicious. `autospec` will create a `MagicMock` that matches what the thing is you're mocking, which you're saying with the type hints is a PMAC but it isn't, it's actually the `i24.pmac` function. This doesn't matter because you're code isn't even using `i24.pmac`, it just uses whatever pmac is passed into `moveto`. Basically I think this all boils down to not needing the patch at all:
```python
@patch("mx_bluesky.I24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.caget")
def test_moveto_oxford_origin(fake_caget: MagicMock):
    fake_pmac: PMAC = MagicMock(spec=PMAC)
    fake_caget.return_value = 0
    moveto(Fiducials.origin, fake_pmac)
    assert fake_caget.call_count == 1
    fake_pmac.x.assert_has_calls([call.move(0.0)])
    fake_pmac.y.assert_has_calls([call.move(0.0)])
```
And similar for other tests in this file.

tests/I24/serial/fixed_target/test_moveonclick.py
```diff
@@ -18,14 +20,14 @@
         ((638, 392), "#1J:-3828", "#2J:-2352"),
     ],
 )
-@patch("mx_bluesky.I24.serial.fixed_target.i24ssx_moveonclick.i24.pmac")
+@patch("mx_bluesky.I24.serial.fixed_target.i24ssx_moveonclick.i24.pmac", autospec=True)
 @patch("mx_bluesky.I24.serial.fixed_target.i24ssx_moveonclick.get_beam_centre")
 def test_onMouse_gets_beam_position_and_sends_correct_str(
```
DominicOram, line 25: Should: Same as above `fake_pmac` isn't what you think it is and `i24.pmac` doesn't need patching

DominicOram (changes requested):
Thanks, some more comments, sorry

src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
```diff
@@ -20,6 +20,19 @@ class ChipType(IntEnum):
...
+        return self.name
+
+    def get_approx_chip_size(self) -> float:
+        """Returns an approximation of the chip size for the move during alignment \
+            of the fiducials
+        """
+        if self.name == "OxfordInner":
+            return 24.60
+        else:
+            return 25.40
```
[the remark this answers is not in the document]
noemifrisina: Spun into #77

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -599,57 +601,52 @@ def load_full_map(fullmap_path: Path | str = FULLMAP_PATH):
...
+        move_status = pmac.x.move(x, wait=False)
+        move_status &= pmac.y.move(y, wait=False)
+        move_status &= pmac.z.move(z, wait=False)
+        move_status.wait()
 
     # Non Chip Specific Move
     if place == "zero":
         logger.info("Moving to %s" % place)
-        caput(pv.me14e_pmac_str, "!x0y0z0")
+        pmac.pmac_string.put("!x0y0z0", wait=True)
```
DominicOram, line 649: Should: As in https://github.com/DiamondLightSource/mx_bluesky/pull/76 this doesn't actually do the wait. Same with all the other uses in this PR.

DominicOram (changes requested):
One comment, otherwise good, thanks

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: Ah, thanks. I thought merging #76 would fix all of these but there were still a few missing. I think I fixed them all now

DominicOram (approved):
Great, ta!

PR #76, opened 2024-01-09
Beamline test fixes

noemifrisina:
Fixes from testing on beamline Dec23/Jan24

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ dependencies = [
     "bluesky",
     "ophyd",
     "pyepics",
+    "caproto",
     "jupyterlab",
     "matplotlib",
     "requests",
--- a/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
+++ b/src/mx_bluesky/I24/serial/fixed_target/i24ssx_moveonclick.py
@@ -39,22 +39,22 @@ def get_beam_centre():
 
 # Register clicks and move chip stages
 def onMouse(event, x, y, flags, param):
-    pmac = param
-    beamX, beamY = get_beam_centre()
     if event == cv.EVENT_LBUTTONUP:
+        pmac = param
+        beamX, beamY = get_beam_centre()
         logger.info("Clicked X and Y %s %s" % (x, y))
         xmove = -1 * (beamX - x) * zoomcalibrator
         ymove = -1 * (beamY - y) * zoomcalibrator
         logger.info("Moving X and Y %s %s" % (xmove, ymove))
         xmovepmacstring = "#1J:" + str(xmove)
         ymovepmacstring = "#2J:" + str(ymove)
-        pmac.pmac_string.set(xmovepmacstring)
-        pmac.pmac_string.set(ymovepmacstring)
+        pmac.pmac_string.put(xmovepmacstring, wait=True)
+        pmac.pmac_string.put(ymovepmacstring, wait=True)
 
 
-def update_ui(frame):
+def update_ui(oav, frame):
     # Get beam x and y values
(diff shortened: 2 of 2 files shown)


noemifrisina:
Code changes now tested on the beamline

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -189,10 +189,10 @@ def load_motion_program_data(
         # Pump setting chosen
         prefix = 14
         logger.info("Setting program prefix to %s" % prefix)
-        pmac.pmac_string.set("P1439=0")
+        pmac.pmac_string.put("P1439=0", wait=True)
```
DominicOram, line 192: Must: Sorry, I was wrong about the syntax here. You actually need to do `pmac.pmac_string.set("P1439=0").wait()` to wait on the value correctly. The `wait=True` only works if you're inside an RE. Same required elsewhere in this PR.

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -132,7 +132,7 @@ def write_parameter_file(param_path: Path | str = PARAM_FILE_PATH_FT):
 
     # If file name ends in a digit this causes processing/pilatus pain.
     # Append an underscore
-    if det_type == "pilatus":
+    if str(det_type) == "pilatus":
```
DominicOram, line 135: Should: Might be nice to do something like `if isinstance(det_type, Pilatus):` instead?

src/mx_bluesky/I24/serial/parameters/constants.py
```diff
@@ -14,26 +14,11 @@ class SSXType(Enum):
...
-CS_FILES_PATH = (
-    Path("src/mx_bluesky/I24/serial/parameters/fixed_target/cs").expanduser().resolve()
-)
+
+PARAM_FILE_PATH = Path(__file__).absolute().parent
+PARAM_FILE_PATH_FT = PARAM_FILE_PATH / "fixed_target"
+LITEMAP_PATH = PARAM_FILE_PATH / "fixed_target/litemaps"
+FULLMAP_PATH = PARAM_FILE_PATH / "fixed_target/fullmaps"
+PVAR_FILE_PATH = PARAM_FILE_PATH / "fixed_target/pvar_files"
+CS_FILES_PATH = PARAM_FILE_PATH / "fixed_target/cs"
```
DominicOram, line 24: Nit: Why not use the `PARAM_FILE_PATH_FT` here?

src/mx_bluesky/I24/serial/setup_beamline/setup_detector.py
```diff
@@ -67,14 +70,24 @@ def _move_detector_stage(detector_stage: DetectorMotion, target: float):
     )
 
 
+# Workaround in case the PV value has been set to the detector name
+def _get_requested_detector(det_type_pv: str):
```
DominicOram, line 74: Could: With detector being potentially being a `Detector` object, int or string of the name it would be good to have a type hint on what this returns and maybe a docstring to be explicit

DominicOram (changes requested):
Thanks, I like abstracting some of the detector choosing stuff to a function. Only main issue was my mistake, sorry. Also, could we add the `.txt` generated files to the `.gitignore`? Otherwise we're going to get them changing with every PR we pull from the beamline.

DominicOram (approved):
Great, thank you!

Issue #45, opened 2023-09-15
I24 ssx: Fix permission issues


noemifrisina:
When running as user, there are permission issues when writing and saving logfiles to `/dls_sw/i24/logs/`.
Also, as the eiger software can't make directories and users don't have permission to create them in the visit directory, at the moment the problem is solved by a workaround where the pilatus detector writes one image before data collection starts. 

Need to find a way to get around these permission issues. Maybe use a service, or run as gda ?

noemifrisina:
Use blueapi for this

noemifrisina:
Blueapi has been setup. We shook into either running on procserv if possible (see https://confluence.diamond.ac.uk/display/CNTRLS/configure-ioc) or having it run on the beamline control machine.

noemifrisina:
A working example of using procserv - while waiting for a cluster to be set up - is on the beamline and being tested.

noemifrisina:
Closing this as done. Procserv has been  running on the beamline for some time and mx-bluesky has been used during the last run for user beamtimes. 
Any further issues will be solved/improved once the cluster is running, see #906

PR #83, opened 2024-02-26
Workaround for write permissions

noemifrisina:
Temporary (hopefully) workaround for #45 , while waiting for a more permanent solution.
Write log and parameter files to `tmp` in the visit directory, to allow staff to test.


src/mx_bluesky/I24/serial/log.py
```diff
@@ -39,6 +41,12 @@
 logging.config.dictConfig(logging_config)
 
 
+def _read_visit_from_file() -> Path:
```
DominicOram, line 44: Could: The name is a bit misleading, this provides the visit_directory, not visit

src/mx_bluesky/I24/serial/log.py
```diff
@@ -51,18 +59,12 @@ def _get_logging_file_path() -> Path:
     logging_path: Path
 
     if beamline:
-        logging_path = Path("/dls_sw/" + beamline + "/logs/serial/")
+        # logging_path = Path("/dls_sw/" + beamline + "/logs/serial/")
```
DominicOram, line 62: Nit: Remove commented out code

DominicOram (approved):
Thanks, couple of minor points in the code. In general we need nicer ways to do this but happy for it to be in as a stopgap.

PR #84, opened 2024-02-26
Add dodal logger

### LINKED ISSUE #49: I24: use graylog
Set up graylog

noemifrisina:
A first try at using the logger from dodal.

For now, the configuration still uses its own StreamHandler (set to DEBUG level) in order to keep the terminal output the same as the scientists are used to, as well as a separate filehandler for debugging purposes - this will be deleted at a later time (issue to come). Because of this, the StreamHandler from dodal  has been removed to avoid duplication of all the messages above INFO level.

This should also close #49 , as the graylog handler comes with the dodal logger.


dperl-dls (changes requested):
~~I don't think this will work so well - with a `StreamHandler` at `DEBUG` and ophyd logs integrated, all the ophyd logs will be printed to console - every change of value on every PV, possibly hundreds of lines per second. The scientists won't be able to read the debug messages they are used to reading.~~

dperl-dls (approved):

PR #87, opened 2024-03-19
Update plans to use new Zebra device

noemifrisina:
See [Dodal#382](https://github.com/DiamondLightSource/dodal/pull/382).

Zebra device has been converted to ophyd_async. Update plans and tests accordingly.

**To test:**
- Run unit tests and confirm they pass

### DIFF
--- a/tests/I24/serial/conftest.py
+++ b/tests/I24/serial/conftest.py
@@ -1,5 +1,3 @@
-from unittest.mock import MagicMock
-
 import pytest
 from bluesky.run_engine import RunEngine
 from dodal.beamlines import i24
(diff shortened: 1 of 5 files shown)


tests/I24/serial/conftest.py
```diff
@@ -1,20 +1,21 @@
-from unittest.mock import MagicMock
-
 import pytest
 from bluesky.run_engine import RunEngine
 from dodal.beamlines import i24
 from dodal.devices.zebra import Zebra
 from ophyd.status import Status
 
 
+@pytest.fixture
```
dperl-dls, line 8: This actually isn't necessary, I realised recently, you can use `NullStatus` from `ophyd.sim` to get the same thing

noemifrisina: It wasn't used, so I just removed it

dperl-dls (approved):
LGTM

PR #96, opened 2024-05-16
Add new zebra plan for pump probe fixed target collection with long delays

### LINKED ISSUE #94: Add new zebra plan for fixed target pump probe
Add the new zebra set up to run pump probe collections on a fastchip with long intervals between the laser and the actual collection - which means having to open and close the shutter when taking images to avoid frying the crystals over the long exposure times.
(For reference in old scripts: `fastchip-zebratrigger-{pilatus,eiger}-medium1`)

A description of the zebra setting for this can be found [here](https://confluence.diamond.ac.uk/display/MXTech/Zebra+settings+I24)

On the edm side, a new choice has been added to the pump probe mus on the main screen (option 8 = "Medium1") and the new pump_probe screen will have a new graph (see picture below)

image

noemifrisina:
Closes #94 

- Plan to set up zebra for fixed target pump probe with long delays i.e. opening and closing the fast shutter between exposures
- New setting choice in pump probe MUX on edm screen for "Medium1"

To test:
- Check tests pass


src/mx_bluesky/I24/serial/setup_beamline/setup_zebra_plans.py
```diff
@@ -336,6 +337,106 @@ def setup_zebra_for_fastchip_plan(
     logger.info("Finished setting up zebra.")
 
 
+def setup_zebra_for_fastchip_pump_probe_with_long_delays_plan(
```
DominicOram, line 340: Should: Generally, I think we can be much more clever with sharing code between here and `setup_zebra_for_fastchip_plan` whilst making the code more readable. It seems that the only difference between them is:
* Control of the shutter on `PULSE_2`
* Delay on the start of the position compare

Additionally the order of a lot of these things doesn't really matter so could we do something like pull all common stuff out into something like `trigger_detector_at_every_chip_position(start_offset...` or maybe even more clean, add start offset to `setup_zebra_for_fastchip_plan` and just call that plan in this one?

We could then add a new plan stub for `setup_zebra_to_open_shutter_at_each_position`

src/mx_bluesky/I24/serial/setup_beamline/setup_zebra_plans.py
```diff
@@ -336,6 +337,106 @@ def setup_zebra_for_fastchip_plan(
...
+    yield from bps.abs_set(zebra.pc.gate_input, IN3_TTL, group=group)
+
+    # Output panel pulse_2 settings
+    yield from bps.abs_set(zebra.output.pulse_2.input, PC_GATE, group=group)
+    yield from bps.abs_set(zebra.output.pulse_2.delay, 0.0, group=group)
+    pulse2_width = num_exposures * exposure_time + SHUTTER_OPEN_TIME
+    yield from bps.abs_set(zebra.output.pulse_2.width, pulse2_width, group=group)
+
+    # Fast shutter
+    yield from bps.abs_set(zebra.output.out_pvs[4], PULSE2, group=group)
```
DominicOram, line 405: Should: Can we pull this TTL into a constant like we have for PILATUS/EIGER?

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -380,14 +387,24 @@ def start_i24(zebra: Zebra, parameters: FixedTargetParameters):
...
+            )
+        else:
+            yield from setup_zebra_for_fastchip_plan(
+                zebra,
+                parameters.detector_name,
+                num_gates,
+                parameters.num_exposures,
+                parameters.exposure_time_s,
+                wait=True,
+            )
```
DominicOram, line 407: Could: See below for discussion on how to split up the plan, we could even then call the separate parts here e.g.
```suggestion
         start_time_offset = SHUTTER_OPENING_TIME if PumpProbeSetting.Medium1 else 0
          yield from setup_zebra_for_fastchip_plan(
              zebra,
              parameters.detector_name,
              num_gates,
              parameters.num_exposures,
              parameters.exposure_time_s,
              start_time_offset
              wait=True,
          )
        if parameters.pump_repeat == PumpProbeSetting.Medium1:
            yield from open_shutter_at_each_location(...)
```

I'm not sure if that's much cleaner or not though

src/mx_bluesky/I24/serial/setup_beamline/setup_zebra_plans.py
```diff
@@ -336,6 +337,106 @@ def setup_zebra_for_fastchip_plan(
...
+        det_type (str): Detector in use, current choices are Eiger or Pilatus.
+        num_gates (int): Number of apertures to visit in a chip.
+        num_exposures (int): Number of times data is collected in each aperture.
+        exposure_time (float): Exposure time for each shot.
+    """
+    logger.info(
+        "Setup ZEBRA for a pump probe fixed target collection with long delays."
+    )
+
+    yield from set_shutter_mode(zebra, "manual")
```
DominicOram, line 392: Should: Why are we setting shutter mode to manual here then automatically controlling the shutter later? That seems like a contradiction

DominicOram (changes requested):
Thank you, some comments in code

src/mx_bluesky/I24/serial/setup_beamline/setup_zebra_plans.py
noemifrisina: Good point. Asked Robin and he's going to test changing this tomorrow...

src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Well, having just one call to `setup_zebra_for_fastchip_plan` instead of the `if.. else..` does make it a bit more readable imo... 
I wonder if it would be a bit tidier to call `open_shutter_at_each_location` from the main `setup_zebra` plan instead of the collection one?

src/mx_bluesky/I24/serial/setup_beamline/setup_zebra_plans.py
```diff
@@ -337,104 +341,44 @@ def setup_zebra_for_fastchip_plan(
     logger.info("Finished setting up zebra.")
 
 
-def setup_zebra_for_fastchip_pump_probe_with_long_delays_plan(
+def open_fast_shutter_at_each_position_plan(
```
DominicOram, line 344: Should: I think a bit of logging in this function would be good

tests/I24/serial/setup_beamline/test_zebra_plans.py
```diff
@@ -119,11 +122,29 @@ async def test_setup_zebra_for_fastchip(zebra: Zebra, RE):
...
+
+    RE(open_fast_shutter_at_each_position_plan(zebra, num_exposures, exposure_time))
+
+    # Check output Pulse2 is set
+    assert await zebra.output.pulse_2.input.get_value() == PC_GATE
+    assert await zebra.output.pulse_2.delay.get_value() == 0.0
+    # num_exp * exposure_time + 0.05 (SHUTTER_OPEN_TIME)
+    assert await zebra.output.pulse_2.width.get_value() == pytest.approx(
+        0.052, abs=1e-3
+    )
```
DominicOram, line 143: Could: I think it's cleaner to do the maths here properly i.e.
```suggestion
    expected_pulse_width = num_exp * exposure_time + 0.05 
    assert await zebra.output.pulse_2.width.get_value() == pytest.approx(
        expected_pulse_width, abs=1e-3
    )
```

DominicOram (approved):
Thanks, I think this is a lot cleaner. Couple of comments, happy for you to merge after addressing them

PR #103, opened 2024-06-04
i04 thawing plan

### LINKED ISSUE #102: Write i04 thawing plan
i04 would like a bluesky plan that:
* Takes a time for how long to spend thawing
* Turns on the thawer
* Sets the smargon velocity such that a full rotation takes that long
* Turns off thawing

DominicOram:
Fixes #102

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,21 +1,20 @@
 {
-    "python.linting.pylintEnabled": false,
-    "python.linting.flake8Enabled": false,
-    "python.linting.mypyEnabled": true,
-    "python.linting.enabled": true,
     "python.testing.pytestArgs": [],
     "python.testing.unittestEnabled": false,
     "python.testing.pytestEnabled": true,
-    "python.formatting.provider": "black",
-    "python.languageServer": "Pylance",
+    "python.analysis.autoImportCompletions": true,
     "editor.formatOnSave": true,
+    "python.defaultInterpreterPath": ".venv/bin/python",
     "[python]": {
         "editor.rulers": [
             88
         ],
         "editor.codeActionsOnSave": {
-            "source.fixAll.ruff": "never",
+            "source.fixAll.ruff": "explicit",
             "source.organizeImports.ruff": "explicit"
-        }
+        },
+        "editor.defaultFormatter": "charliermarsh.ruff"
     },
+    "terminal.integrated.gpuAcceleration": "off",
+    "python.analysis.typeCheckingMode": "basic",
 }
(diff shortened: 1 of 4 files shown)


src/mx_bluesky/i04/thawing_plan.py
```diff
@@ -0,0 +1,39 @@
...
+from dodal.common import inject
+from dodal.devices.smargon import Smargon
+from dodal.devices.thawer import Thawer, ThawerStates
+
+
+def thaw(
+    time_to_thaw: float,
+    rotation: float = 360,
+    thawer: Thawer = inject("thawer"),  # type: ignore
+    smargon: Smargon = inject("smargon"),  # type: ignore
```
callumforrester, line 13: Did the type checker not like this? That's worrying...
Any ideas @DiamondJoseph

.vscode/settings.json
callumforrester, line 1: Nit: I think this is better in a separate PR

callumforrester (approved):

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
     "opencv-python",
     "pydantic",
     "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
+    "blueapi>=0.4.3-a3",
```
DiamondJoseph, line 25: Shouldn't be required after get MsgGenerator from dodal

src/mx_bluesky/i04/thawing_plan.py
```diff
@@ -0,0 +1,39 @@
+import bluesky.plan_stubs as bps
+import bluesky.preprocessors as bpp
+from blueapi.core import MsgGenerator
+from dodal.common import inject
```
DiamondJoseph, line 4:
```suggestion
from dodal.common import MsgGenerator, inject
```

src/mx_bluesky/i04/thawing_plan.py
DominicOram: `inject` returns a string. I thought about making an issue but I'm not sure how you could fix it cleanly.

.vscode/settings.json
DominicOram: Yh, probably, this was me being lazy.

src/mx_bluesky/i04/thawing_plan.py
DiamondJoseph: Looks like mypy is still happy with it, but Pylance in the IDE isn't. If inject gains `-> Any` it behaves again. I'll make a PR to dodal.

DiamondJoseph (changes requested):

.vscode/settings.json
DominicOram: On second thoughts I'm sticking by my laziness

DiamondJoseph (approved):
I think it needs __init__.py in i04?

PR #111, opened 2024-06-13
Do not have paths for parameter files as input argument in plans

### LINKED ISSUE #107: I24ssx: Use pydantic FilePath for parameter/map file paths in plans
In many chip manager plans there is an need to pass the path to save a parameter or a map file. As of #95 , these paths are passed as an `Optional[str]` and the plan with default to an hard-coded value if this is `None`. This is not very tidy but the change was made because of an errror raised by `blueAPI` when setting a type to `Path`.
 
Using the `FilePath` from pydantic should help solve the issue, see [https://docs.pydantic.dev/latest/api/types/#pydantic.types.FilePath](https://docs.pydantic.dev/latest/api/types/#pydantic.types.FilePath)

noemifrisina:
Closes #107


dperl-dls (approved):
lgtm

PR #114, opened 2024-06-26
Tidy up stages initialisation

### LINKED ISSUE #51: I24: tidy up initialise in chip manager
The `initialise` method found in `Chip_Manager` for a fixed target collection is supposed to only reset the general purpose PVs and the PVs associated with the geobrick which sometimes need to be reset - usually when going to sacla and switching it off.

Right now  however it also determines the detector in use from the PV and needs to be run every time the detector is changed. This operation is unrelated and only there for convenience's sake. This could now be done somewhere else.

### LINKED ISSUE #109: I24ssx: remove setting .VMAX for stages
When the pmac stages are turned off because they are moved (either from lab or from other facility), the ME14E stages PVs need to re-initialised.
Currently the `.VMAX` field of the motor stages is also manually set to 20. This PV however is read-only in ophyd_async.

Setting of `.VMAX` should be removed after checking with controls that:
- like most IOCs the last value set to PVs is saved and will still be there after turning it on and off
- The current value for `.VMAX` is 20 (otherwise resetting `.VELO` to 20 will cause problems)

noemifrisina:
Closes #109 
Closes #51 

A bit of tidying up in `initialise_stages` i.e. the plan run once in awhile when the stages have been off and rebooted.

- [x] Remove setting .VMAX as it's now 20 by default in controls
- [x] Remove setting the detector, as the `ME14E-MO-IOC-01:GP101` PV can now be re-initialised using the MUX choice in the `Detector` edm. Add a message at the end reminding the user to do that.

### DIFF
--- a/src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
+++ b/src/mx_bluesky/I24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
@@ -49,7 +49,6 @@ def setup_logging():
 @log.log_on_entry
 def initialise_stages(
     pmac: PMAC = inject("pmac"),
-    detector_stage: DetectorMotion = inject("detector_motion"),
 ) -> MsgGenerator:
     """Initialise the portable stages PVs, usually used only once right after setting \
         up the stages either after use at different facility.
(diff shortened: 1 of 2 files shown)


dperl-dls (approved):
lgtm

PR #119, opened 2024-07-01
Fix linting 

noemifrisina:
Fixes error in CI

```error: `ruff <path>` has been removed. Use `ruff check <path>` instead.```

`ruff path` has been deprecated, need to update pre-commits


rtuck99 (approved):
approved

PR #121, opened 2024-07-05
Add General Purpose PVs to parameter model

### LINKED ISSUE #118: I24 SSX: Address general purpose PVs
We have a lot of general purpose PVs that are used throughout the script, many of these are actually just parameters from the EDM and so we would like to ultimately migrate them to be parameters.

## Acceptance Criteria
* For every general purpose PV in the i24 SSX scripts:
   * If they are actually parameters put them in the parameter models and only read them once
   * If not make a new issue to address them

noemifrisina:
Closes #118 

This spawned #122

### DIFF
--- a/src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
+++ b/src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
@@ -7,10 +7,6 @@ New ones may be added if needed in the future.
 """
 
 from enum import Enum, IntEnum
-from typing import Dict, Mapping
-
-import mx_bluesky.I24.serial.setup_beamline.pv as pv
-from mx_bluesky.I24.serial.setup_beamline import caget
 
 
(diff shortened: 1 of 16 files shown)


src/mx_bluesky/I24/serial/fixed_target/ft_utils.py
```diff
@@ -51,9 +46,42 @@ class PumpProbeSetting(IntEnum):
...
+
 
 class Fiducials(str, Enum):
     origin = "origin"
     zero = "zero"
     fid1 = "f1"
     fid2 = "f2"
+
+
+def get_chip_format(chip_type: ChipType) -> Mapping:
```
DominicOram, line 61: Could: Why not just have this return a `ChipDescription` object?

DominicOram (approved):
Couple of suggestions in code, otherwise looks good.

PR #124, opened 2024-07-12
Use BlueAPI abort

### LINKED ISSUE #117: I24 SSX: Use BlueAPI abort rather than general PV
The SSX scripts currently use `me14e_gp9` as a PV for whether we want to abort or not. The bluesky way of doing this is calling abort on the RE, this can be done through blueAPI using `blueapi abort`. We need to make sure that when the plan is aborted it correctly tidies up, we should do this with the [contingency wrapper](https://blueskyproject.io/bluesky/main/generated/bluesky.preprocessors.contingency_wrapper.html#bluesky.preprocessors.contingency_wrapper). We need to make sure the wrapper covers as much of the plan as possible so that if we get an abort when shutter is opening we properly close it. i.e. the chip collect will look something like:

```python

def run_fixed_target_plan(...):
  yield from bpp.contingency_wrapper(
      _run_fixed_target_plan(...), except_plan=run_aborted_plan(...), final_plan=tidy_up_plan(...)
  )

def tidy_up_plan(..):
    # Everything after the wait for the collection to finish

def _run_fixed_target_plan(...):
    # Same as current plan up to when the collection is running then we just have the wait for it to finish
```

noemifrisina:
Closes #117 

Needs [Dodal661](https://github.com/DiamondLightSource/dodal/pull/661)

### DIFF
--- a/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
+++ b/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
@@ -166,7 +166,7 @@ def write_parameter_file(detector_stage: DetectorMotion):
     pump_delay = float(caget(pv.ioc12_gp10)) if pump_status else None
 
     params_dict = {
-        "visit": log._read_visit_directory_from_file(),
+        "visit": log._read_visit_directory_from_file().as_posix(),
         "directory": caget(pv.ioc12_gp2),
         "filename": filename,
         "exposure_time_s": float(caget(pv.ioc12_gp5)),
--- a/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
+++ b/src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
@@ -16,6 +16,7 @@ from pprint import pformat
 from time import sleep
 
 import bluesky.plan_stubs as bps
+import bluesky.preprocessors as bpp
 from blueapi.core import MsgGenerator
 from dodal.common import inject
 from dodal.devices.hutch_shutter import HutchShutter, ShutterDemand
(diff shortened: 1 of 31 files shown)


src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -364,51 +365,64 @@ def run_extruder_plan(
...
+def collection_aborted_plan(zebra: Zebra, detector_name: str) -> MsgGenerator:
+    """A plan to run in case the collection is aborted before the end."""
+    global ABORTED
+    ABORTED = True
+    logger.warning("Data Collection Aborted")
+    yield from disarm_zebra(zebra)  # If aborted/timed out zebra still armed
+    if detector_name == "pilatus":
+        caput(pv.pilat_acquire, 0)
+    elif detector_name == "eiger":
+        caput(pv.eiger_acquire, 0)
```
DominicOram, line 408: Should: The `finally` plan will do the detector disarm, right?

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -364,51 +365,64 @@ def run_extruder_plan(
...
+    sleep(1.0)
```
DominicOram, line 409: Should: Why are we sleeping?

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -364,51 +365,64 @@ def run_extruder_plan(
...
+    ABORTED = False
+
+
+@log.log_on_entry
+def collection_aborted_plan(zebra: Zebra, detector_name: str) -> MsgGenerator:
+    """A plan to run in case the collection is aborted before the end."""
+    global ABORTED
+    ABORTED = True
+    logger.warning("Data Collection Aborted")
+    yield from disarm_zebra(zebra)  # If aborted/timed out zebra still armed
```
DominicOram, line 404: Should: Doesn't `reset_zebra_when_collection_done_plan` in the `finally` plan deal with this?

DominicOram (changes requested):
Looks like it should work. I'm not a fan of the `ABORTED` global variable. I think that you should just call `dcid.collection_complete(end_time, aborted=False)` in an `else_plan` and `dcid.collection_complete(end_time, aborted=True)` in the except plan. Ultimately we should be using a callback for ispyb and it will be handled in there, could you make an issue for this if there isn't one already?

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: I believe it's to allow the PVs to actually be set - since there is no wait option on this `caput`. 1s is probably quite long though, in other parts of the code when setting these PVs it's `0.5`...

src/mx_bluesky/I24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Yes, although after going through these changes I'm not entirely sure we should be disarming the zebra in that plan. Or to be more precise, it's fine as it is for the fixed target, but in the extruder the zebra is disarmed at collection end so we shouldn't need to do it unless the collection failed/was aborted

DominicOram: I don't think it really costs anything to do it in both places

dperl-dls (approved):
Looks good, if we see any issues we'll fix them separately

PR #127, opened 2024-07-17
Beamline test hotfixes

noemifrisina:
On 17072024


olliesilvester (approved):
I think this might fix your logging issue but I would suggest setting up the logging by calling your `default_logging_setup` instead of using the `logging.config`

noemifrisina:
Thanks. I do call `default_logging_setup` when setting up actually, used elsewhere in the code. For now I'm trying to keep the stream somewhat close to what the beamline has had for years, which means it's a bit of a work in progress still... Right now it's enough that the logs are readable I guess

PR #131, opened 2024-07-22
Double speed of thaw

### LINKED ISSUE #130: I04: Thaw in both directions
The thawing plan for I04 can be optimised to instead of running 360 in the time requested running 360 in half the time then -360 in half the time, meaning that we end up back at the same place we started.

## Acceptance Criteria
* The thawing plan ends up in the same place as we started
* The time taken to do this is the same

DominicOram:
Fixes #130 

To test:
* Confirm tests pass


noemifrisina (approved):
Thanks

PR #134, opened 2024-07-29
Estimate a timeout for fixed-target collections

### LINKED ISSUE #133: I24ssx: calculate timeouts for different fixed target pp settings
`pmac.run_program` needs a timeout to be passed to wait for the status PVAR value to be updated at end of collection.
This timeout value might be different depending on which pump probe setting has been requested, so it needs to be calculated. For this the `pumpprobe_calc` plan may be reused - after a bit of refactoring.

Note. Passing `None` will wait indefinitely.

noemifrisina:
Closes #133 

Needs #124 to be merged first.

Adds a small function to calculate the time the PMAC should wait for the collection to be completed, depending on the pump-probe settings.

NOTE. These timeouts are only rough estimates and will need to be tested on the beamline.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ dependencies = [
     "requests",
     "opencv-python",
     "pydantic",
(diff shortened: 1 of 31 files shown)


dperl-dls (approved):
LGTM, thanks

PR #138, opened 2024-08-02
move wiki docs to sphinx

dperl-dls:
- made a subdirectory for i24-ssx docs
- copied everything from the wiki to the docs


dperl-dls:
the huge i24 ssx document should still be split up and recategorised

PR #140, opened 2024-08-05
Add do fgs_plan

### LINKED ISSUE #80: Hyperion refactor: Add lowest level FGS plan
This issue begins moving Hyperion's plans into a more generic form inside this repo, see https://github.com/DiamondLightSource/mx-bluesky/issues/296. The idea is to decompose Hyperion bluesky plans into common building blocks for any MX beamline to use. A beamline can then construct its specific implementation by building up these smaller generic plans. We should also look into using bluesky's 'prepare' command for device setup.

To kick off , I think it makes sense for the first plan added to be the lowest level plan for the fast grid scan. Since it is the lowest level, we need to make sure that this plan is as generic as possible, and that any beamline wanting to do a fast grid scan can use this base plan. It should have similar functionality to `do_fgs()` in hyperion's `flyscan_xray_centre_plan.py`. Its parameters should be:

- A list of devices. We should work out which devices are essential for this plan ~(is it just the FGS motor?)~, and leave room for optional devices  (like zocalo).
- ~A generic 'device setup plan' (which could look like a two level plan, one to specify each individual device setup and one to specify the order/waiting requirements)~
- ~A generic 'device tidy plan', to be used if there is an exception during the plan, or, optionally, at the end of the plan~

The plan should:
- ~Set up all devices as specified in parameters and wait appropriately~
- kick off the FGS
- Optionally wait for any optional devices (like zocalo)
- do a `bps.complete()`
- ~Do any tidy-up~
- ~Be in a try block so tidy-up happens on an exception~

EDIT: This is like a plan stub, so is not designed to be ran by itself. Therefore, this plan shouldn't consider setup or tidy-up

olliesilvester:
Fixes #80 as part of https://github.com/orgs/DiamondLightSource/projects/6/views/1?pane=issue&itemId=73047222. This adds a slightly modified version of the `do_fgs` plan in Hyperion. Here, you can optionally device whether or not we need to do extra things - eg zocalo. This plan should be usable for VMXm, i03, i04, and hopefully other MX beamlines.

~The idea of the `level 1` folder is for it to be filled with plan stubs, i.e plans that cannot be ran on their own. I guess it could just be renamed to `plan_stubs`...~ I renamed it `plan_stubs`

To test:

- Check functionality is sensible, and compare to hyperion's do_fgs (https://github.com/DiamondLightSource/hyperion/blob/14421be796985f710129de1f105e2f1e65bf6a71/src/hyperion/experiment_plans/flyscan_xray_centre_plan.py#L329)
- Confirm tests are sufficient


dperl-dls (commented):
I think we will probably want another level of categorisation than this, maybe put `do_fgs` into a gridscan module?

DominicOram (changes requested):
Thanks, I think we can maybe keep them even more similar

src/mx_bluesky/plan_stubs/do_fgs.py
```diff
@@ -0,0 +1,53 @@
...
+    )
+    if pre_plans:
+        yield from pre_plans()
+    LOGGER.info("Wait for all moves with no assigned group")
+    yield from bps.wait()
+    LOGGER.info("kicking off FGS")
+    yield from bps.kickoff(grid_scan_device, wait=True)
+    gridscan_start_time = time()
+    if post_plans:
+        yield from post_plans()
```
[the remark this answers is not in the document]
DominicOram: So the wait for zocalo stage has the potential to take 2 seconds. I think maybe for the moment keep the `post_plans` bit and call it with a wait for stage on hyperion. I think it should be renamed to `during_collection_plans` though, as that's what it is

src/mx_bluesky/plan_stubs/do_fgs.py
```diff
@@ -0,0 +1,53 @@
...
+    expected_images = yield from bps.rd(grid_scan_device.expected_images)
+    exposure_sec_per_image = yield from bps.rd(detector.cam.acquire_time)
+    LOGGER.info("waiting for topup if necessary...")
+    yield from check_topup_and_wait_if_necessary(
+        synchrotron,
+        expected_images * exposure_sec_per_image,
+        30.0,
+    )
+    if pre_plans:
+        yield from pre_plans()
```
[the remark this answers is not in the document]
olliesilvester: Agreed, but I also think `pre_plans` gives us a bit more flexibility in case we want to also read other hardware at this point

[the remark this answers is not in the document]
DominicOram: I would suggest we don't build in complexity we don't yet need

src/mx_bluesky/device_setup_plans/read_hardware_for_setup.py
```diff
@@ -0,0 +1,10 @@
+import bluesky.plan_stubs as bps
+from dodal.devices.eiger import EigerDetector
+
+from mx_bluesky.i03.parameters.constants import CONST
+
+
+def read_hardware_for_zocalo(detector: EigerDetector):
```
olliesilvester, line 7: Arguably I could just move all these plans in the same PR. I thought i'd try and separate it out a bit though, otherwise it will quickly spiral

src/mx_bluesky/common/device_setup_plans/read_hardware_for_setup.py
```diff
@@ -0,0 +1,11 @@
+import bluesky.plan_stubs as bps
+from dodal.devices.eiger import EigerDetector
+
+from mx_bluesky.common.parameters.constants import MxConstants
+
+
+def read_hardware_for_zocalo(detector: EigerDetector):
```
DominicOram, line 7: Should: This read is also when zocalo is triggered, I think it's worth adding that into the docstring here

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -322,38 +329,15 @@ def kickoff_and_complete_gridscan(
         except_plan=lambda e: (yield from bps.stop(eiger)),  # type: ignore # See: https://github.com/bluesky/bluesky/issues/1809
         else_plan=lambda: (yield from bps.unstage(eiger)),  # type: ignore # See: https://github.com/bluesky/bluesky/issues/1809
     )
-    def do_fgs():
-        # Check topup gate
-        expected_images = yield from bps.rd(gridscan.expected_images)
-        exposure_sec_per_image = yield from bps.rd(eiger.cam.acquire_time)  # type: ignore # See: https://github.com/bluesky/bluesky/issues/1809
-        LOGGER.info("waiting for topup if necessary...")
-        yield from check_topup_and_wait_if_necessary(
+    def do_fgs_and_wait_for_zocalo_and_read_hardware():
```
DominicOram, line 332: Should: This doesn't wait for the zocalo results to be finished, that's done on the `trigger_and_read` of zocalo.

src/mx_bluesky/common/plan_stubs/do_fgs.py
```diff
@@ -0,0 +1,55 @@
...
+    expected_images = yield from bps.rd(grid_scan_device.expected_images)
+    exposure_sec_per_image = yield from bps.rd(detector.cam.acquire_time)  # type: ignore # See: https://github.com/bluesky/bluesky/issues/1809
+    LOGGER.info("waiting for topup if necessary...")
+    yield from check_topup_and_wait_if_necessary(
+        synchrotron,
+        expected_images * exposure_sec_per_image,
+        30.0,
+    )
+    # RE must be subscribed to an appropriate callback for this read information to be saved
+    yield from read_hardware_for_zocalo(detector)
```
DominicOram, line 42: Should: In re-reading the code I actually think we need to wait for the zocalo stage to be finished before we do this read, which will trigger the results to zocalo. We should do both during the collection

DominicOram (changes requested):
Sorry to put comments on this again. Is there a reason we didn't just put all of `kickoff_and_complete_gridscan` in common? I think it is common to all beamlines?

olliesilvester:
> Sorry to put comments on this again. Is there a reason we didn't just put all of `kickoff_and_complete_gridscan` in common? I think it is common to all beamlines?

No real reason, I was just trying to move bits incrementally, and wasn't sure about the Zocalo bits on other beamlines

src/mx_bluesky/common/device_setup_plans/read_hardware_for_setup.py
olliesilvester: Good catch, I didn't realise it did that!

DominicOram:
> > Sorry to put comments on this again. Is there a reason we didn't just put all of `kickoff_and_complete_gridscan` in common? I think it is common to all beamlines?
> 
> No real reason, I was just trying to move bits incrementally, and wasn't sure about the Zocalo bits on other beamlines

I agree about the zocalo stuff not being needed everywhere but I think the other decorators are common, and would be needed for nexus/ispyb writing

olliesilvester:
OK, I'll put the whole function in `common`

src/mx_bluesky/common/plan_stubs/do_fgs.py
olliesilvester: Hmm, the original code does the read before waiting for zocalo to stage
```
def do_fgs():
...
  yield from read_hardware_for_zocalo(eiger)
  LOGGER.info("Wait for all moves with no assigned group")
  yield from bps.wait()
  LOGGER.info("kicking off FGS")
  yield from bps.kickoff(gridscan, wait=True)
  gridscan_start_time = time()
  LOGGER.info("Waiting for Zocalo device queue to have been cleared...")
  yield from bps.wait(
  ZOCALO_STAGE_GROUP
   )  # Make sure ZocaloResults queue is clear and ready to accept our new data
```
But are you saying that Hyperion should also be doing the wait first?

olliesilvester: As discussed, we should stage zocalo immediately after the detector is armed, and wait for it to finish before doing this read.

I've made https://github.com/DiamondLightSource/mx-bluesky/issues/590 but I may do it in this PR

src/mx_bluesky/common/plan_stubs/do_fgs.py
```diff
@@ -0,0 +1,113 @@
...
+
+    Args:
+        gridscan (FastGridScanCommon):          Device which can trigger a fast grid scan and wait for completion
+        detector (EigerDetector)                Detector device
+        synchrotron (Synchrotron):              Synchrotron device
+        scan_points (list[AxesPoints[Axis]]):   Each element in the list contains all the grid points for that grid scan.
+                                                Two elements in this list indicates that two grid scans will be done, eg for Hyperion's 3D grid scans.
+        scan_start_indices (list[int]):         Contains the first index of each grid scan
+        plan_during_collection (Optional, MsgGenerator): Generic plan called in between kickoff but and completion, eg waiting on zocalo.
+        plan_name (str):                        Used to check if Zocalo callback should be started in this plan,
```
olliesilvester, line 84: This is pretty confusing to have in the plan stub. We are assuming Zocalo will always be triggered from this plan, so I think we should also make sure the ZocaloCallback is correctly configured just from this plan

olliesilvester:
I started to reshuffle some of the hyperion tests in `test_flyscan_xray_centre` to the common `test_do_fgs`, but this quickly led into me having to move the whole parameter model + all callbacks into common too, so that is a job for a separate ticket

tests/unit_tests/common/plan_stubs/test_do_fgs.py
```diff
@@ -0,0 +1,201 @@
...
+
+        msgs = assert_message_and_return_remaining(
+            msgs, lambda msg: msg.command == "wait"
+        )
+
+
+def test_kickoff_and_complete_gridscan_optionally_calls_during_collection_plan(
+    sim_run_engine: RunEngineSimulator, fgs_devices
+):
+    with patch("mx_bluesky.common.plans.do_fgs.check_topup_and_wait_if_necessary"):
```
DominicOram, line 121: Nit: I have a mild preference for patch decorators as it reduces the indent a bit but not that bothered

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -1035,6 +1027,11 @@ class CompleteException(Exception):
             return_value=Status(None, None, 0, True, True)
         )
 
+        finished_status = Status()
+        finished_status.set_finished()
+
+        fake_fgs_composite.eiger.filewriters_finished = finished_status
```
DominicOram, line 1033: Could: You can use `ophyd.sim.NullStatus` here instead of making you're own finished status

DominicOram (changes requested):
A couple of final questions around the tests but generally there.

tests/unit_tests/common/plan_stubs/test_do_fgs.py
```diff
@@ -0,0 +1,201 @@
...
+            kickoff_and_complete_gridscan(
+                fgs_device,
+                detector,
+                synchrotron,
+                scan_points=[],
+                scan_start_indices=[],
+            )
+        )
+        null_messages = [msg for msg in msgs if msg.command == "null"]
+        assert len(null_messages) == 0
```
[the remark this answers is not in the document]
olliesilvester: Whoops, nothing anymore! I think I meant to remove this

tests/unit_tests/common/plan_stubs/test_do_fgs.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+def test_kickoff_and_complete_gridscan_with_run_engine_correct_documents(
```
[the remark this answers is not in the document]
olliesilvester: Yep, agreed. I did start to do this, but then it lead to having to move the callbacks to common, which requires also moving the parameters + constants to common, so I reverted. Anyway, I'll add a comment saying to change this test once the zocalo callbacks have been moved

PR #141, opened 2024-08-06
Tidy I24 SSX docs a bit

dperl-dls:


dperl-dls:
Merging this without review because it's just shuffling docs around and nobody else is here right now

PR #150, opened 2024-08-07
135 stream to murko

### LINKED ISSUE #135: I04: Stream data into murko whilst thawing
To get the pin centering working in murko we need to reliably stream data to the system. Particularly it needs image data, image metadata such as zoom level, and omega data. See https://github.com/DiamondLightSource/mx_auto_mjpeg_capture for more details.

This should then be enabled during the thawing plan for i04 so that murko is receiving the data as we thaw

## Acceptance Criteria
* Murko is sent the data it needs to do image recognition during thawing
* There is documentation on how this works

DominicOram:
Fixes #135

Requires https://github.com/DiamondLightSource/dodal/pull/733

To test:
* Confirm unit tests pass
* Confirm new docs make sense


docs/developer/murko-integration/index.rst
```diff
@@ -0,0 +1,22 @@
+Murko Integration
+=================
+
+The `Murko system <https://github.com/MartinSavko/murko>`_ uses ML to analyse images of sample from optical cameras and determine where the sample is.
+
+It's integration at DLS, using Bluesky, is still a work in progress but this documentation aims to give an overview of how it works.
```
dperl-dls, line 6:
```suggestion
Its integration at DLS, using Bluesky, is still a work in progress but this documentation aims to give an overview of how it works.
```

dperl-dls (approved):
This seems fine for now. It could easily be a lot more generic if e.g. the stream url was configurable, and I wonder if the performance will always be good enough, but I think those are questions for later

src/mx_bluesky/i04/thawing_plan.py
```diff
@@ -1,16 +1,58 @@
...
+    oav: OAV = inject("oav"),
+    oav_to_redis_forwarder: OAVToRedisForwarder = inject("oav_to_redis_forwarder"),
+) -> MsgGenerator:
+    zoom_percentage = yield from bps.rd(oav.zoom_controller.percentage)
+    sample_id = yield from bps.rd(robot.sample_id)
+
+    yield from bps.abs_set(oav.zoom_controller.level, "1.0x", wait=True)
+
+    @subs_decorator(MurkoCallback(REDIS_HOST, REDIS_PASSWORD, MURKO_REDIS_DB))
+    @run_decorator(
```
dperl-dls, line 31: would it be nicer to give the `oav_to_redis_forwarder` a reference to an OAV and have it fetch these params itself?

dperl-dls (approved):
looks good, one general comment which I'm really not sure about

src/mx_bluesky/i04/thawing_plan.py
DominicOram: I thought about this when I originally did the ticket but I think keeping them separate is good:
* The OAV is currently `ophyd`, not `ophyd-async` so not 100% sure how the mixing will go (but probably fine)
* I think it could be quite valid to want to stream the OAV images out but not need the whole `OAV` device

But I'm also not 100% convinced though. I would suggest leave it for now and see how it evolves?

dperl-dls: yeah, fine by me

PR #165, opened 2024-08-20
Fix parameter file for serial fixed target

noemifrisina:
Cast the values for map type and pump probe when writing the `parameters.json` file.


dperl-dls (approved):
lgtm

PR #186, opened 2024-08-28
Remove input argument from `upload_parameters`

### LINKED ISSUE #160: Remove unused arguments in `upload_parameters()`
Could: It looks like `upload_parameters` is only ever called with `oxford`. If we don't use any other chips a better fix here would be to not even take the chip type as an argument and just hardcode oxford and width=8? Pull into a separate issue if we're not sure

_Originally posted by @DominicOram in https://github.com/DiamondLightSource/mx_bluesky/pull/159#discussion_r1719677375_

noemifrisina:
Closes #160 

- Remove input chip from `upload_parameters` as used only by oxford-type chips
- Update edm


DominicOram (approved):
LGTM, thanks

PR #189, opened 2024-08-29
Remove save parameters button

### LINKED ISSUE #89: I24 ssx: Remove button to save parameter file
For chip experiments the parameter file writer and the start for the collection are two different buttons, leading to the possibility of forgetting to set the parameter files before the collection.

Once a parameter model has been set up, we may be able to remove the first step.

noemifrisina:
Closes #89 

Remove button to save parameters before collection on the edm.
For the moment, still write a parameter file - can be changed later.


dperl-dls (approved):
Seems ok for now, can you make an issue (or link in a comment if there is already one) about not doing file IO in plans please? We should use callbacks for this.

PR #191, opened 2024-08-30
Get moveonclick to start

noemifrisina:
This is a workaround to get `moveonclick` to start. There seems to be an issue connecting both ophyd and ophyd_asynce devices while the plan is running - in this case the OAV device hasn't yet been moved to ophyd_async. 
The workaround instantiates the devices before running the plan.
See [here](https://github.com/bluesky/ophyd-async/issues/548) for details.


DominicOram (approved):
LGTM, thanks

PR #194, opened 2024-08-30
Get `moveonclick` to work

### LINKED ISSUE #44: I24 SSX: Take into account zoom in click to move

noemifrisina:
Multiple `moveonclick` fixes/changes:

- Work out  a value for `zoomcalibrator` from fitting curve: so far this value has been fixed to an empirically determined constant that didn't work well for all zoom levels. This adds a way to work out a correct value depending on the zoom level currently in use. Closes #44 
- Fix mouse callback: problem was that using `yield from bps.abs_set(pmac.pmac_string, ymovepmacstring, wait=True)` in `onMouse` made it a generator, which couldn't work in `setMouseCallback`
- Run the viewer outside of the RunEngine instead of inside.


src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -34,20 +30,34 @@ def _get_beam_centre(oav: OAV):
...
         logger.info(f"Clicked X and Y {x} {y}")
+        zoomcalibrator = RE(_calculate_zoom_calibrator(oav)).plan_result  # type: ignore
         xmove = -1 * (beamX - x) * zoomcalibrator
         ymove = -1 * (beamY - y) * zoomcalibrator
         logger.info(f"Moving X and Y {xmove} {ymove}")
         xmovepmacstring = "#1J:" + str(xmove)
         ymovepmacstring = "#2J:" + str(ymove)
-        yield from bps.abs_set(pmac.pmac_string, xmovepmacstring, wait=True)
-        yield from bps.abs_set(pmac.pmac_string, ymovepmacstring, wait=True)
+        RE(_move_to_position(pmac, xmovepmacstring, ymovepmacstring))
```
DominicOram, line 60: Should: Instead of calling the run engine twice we could just have a plan that calculates the zoom calibration then moves

DominicOram (changes requested):
Thanks, just one comment

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -34,20 +31,46 @@ def _get_beam_centre(oav: OAV):
...
+    zoomcalibrator = yield from _calculate_zoom_calibrator(oav)
+    beamX, beamY = beam_centre
+    x, y = clicked_position
+    xmove = -1 * (beamX - x) * zoomcalibrator
+    ymove = -1 * (beamY - y) * zoomcalibrator
+    logger.info(f"Moving X and Y {xmove} {ymove}")
+    xmovepmacstring = "#1J:" + str(xmove)
+    ymovepmacstring = "#2J:" + str(ymove)
+    yield from bps.abs_set(pmac.pmac_string, xmovepmacstring, wait=True)
+    yield from bps.abs_set(pmac.pmac_string, ymovepmacstring, wait=True)
```
DominicOram, line 62: Should: Can you either use `_move_to_position` here or remove the unused function?

DominicOram (changes requested):
Just one comment, sorry.

DominicOram (approved):
Great, thank you!

PR #429, opened 2024-09-02
Rotation plan sets aperture to large by default

### LINKED ISSUE #203: Aperture not moving in for rotation
During testing today we found that an aperture was not being selected for the rotation scan and the aperture/scatterguard was staying in `ROBOT_LOAD`. 

This may have been because we had no beam and so the XRC was not returning a sensible aperture to go to. Even so we should probably pick some default aperture (which we may have to do in the GDA side). 

## Acceptance Criteria
* We move to a aperture/scatterguard position before rotation

olliesilvester:
This makes the aperture position default to LARGE for rotation scans only, using pydantic validators.

Fixes #203 

### To test:
1. Check changes and new tests


src/mx_bluesky/hyperion/parameters/rotation.py
```diff
@@ -33,6 +35,8 @@
 )
 from mx_bluesky.hyperion.parameters.constants import CONST, I03Constants
 
+DEFAULT_APERTURE_POSITION = AperturePositionGDANames.LARGE_APERTURE
```
DominicOram, line 38: Nit: Can we put this somewhere in `constants`? Maybe under `ExperimentParamConstants.ROTATION`?

DominicOram (approved):
Great, thank you, minor nit, take it or leave it.

DominicOram (commented):
Thanks, can you update for pydantic 2 please?

PR #445, opened 2024-09-02
Zocalo results multiple sources

olliesilvester:
Fixes https://github.com/DiamondLightSource/dodal/issues/559

See related dodal PR: https://github.com/DiamondLightSource/dodal/pull/763


DominicOram (changes requested):
I think you might have accidentally pulled in some changes from https://github.com/DiamondLightSource/mx-bluesky/pull/429

DominicOram (approved):
Thanks!

PR #446, opened 2024-09-02
Configure shutter to always use the manual/auto option

### LINKED ISSUE #232: FGS: Correctly configure zebra to allow manual shutter operation
Whilst running the FGS we need to be sure that if there is an issue we can switch to manual operation of the shutter. The logic to do this controlled by a logic gate in the zebra. However, we have wired it up incorrectly and have bypassed this check.

## Acceptance Criteria
* During the fast grid scan a user is still able to set the shutter to manual operation through the EDM screen

DominicOram:
Fixes #232

Requires https://github.com/DiamondLightSource/dodal/pull/770

To test:
* Confirm tests still pass and logic is understandable


olliesilvester (changes requested):
Looks good, I just think setting the zebra to manual/automatic should be done within the set-up plans rather the top level plans, I may be wrong though

olliesilvester (approved):

PR #453, opened 2024-09-03
Correct default exposure time

olliesilvester:
Hyperion's zebra gridscans work at 250hz, which corresponds to a period of 1/250 (or 0.004) seconds. Our default value here was incorrect


DominicOram (approved):
Thanks!

PR #465, opened 2024-09-11
Fix the docker image CI build workflow

rtuck99:
This should fix the `Publish Docker Image` github workflow.

@DominicOram I don't have permission to push an image to the registry, but I've run the action manually in this branch as far as it will go up to the point where it pushes to `ghcr.io` where it fails presumably because I don't have permissions to push to the `diamondlightsource` ghcr.io repo - so you may want to try this before merging.


DominicOram (approved):
I think this looks fine and we'll soon find out if there are issues on the next release

PR #466, opened 2024-09-11
Use pmac abort program

noemifrisina:
Uses change in [Dodal#771](https://github.com/DiamondLightSource/dodal/pull/773)


DominicOram (changes requested):
Great thanks. One comment.

Should: We should change the `test_run_aborted_plan` to assert the abort signal is triggered as although this will still work it's testing implementation details of the device

PR #468, opened 2024-09-11
Fixes to kubernetes deployment helmcharts

rtuck99:
Add data dir to helmchart, make log dir writable, fix project dir version prefix


DominicOram (approved):
Great, thank you!

PR #480, opened 2024-09-12
Do not reload sample if its already been loaded

### LINKED ISSUE #240: Robot load: Running on the same sample twice results in the sample being unloaded and reloaded
Likely a GDA issue. When asking to do a hyperion robot load + data collection on a sample which is already loaded, the sample will be unloaded by the robot, and then reloaded. In this case, the robot shouldn't do anything at all

## Acceptance Criteria

- Implement fix and test on beamline

DominicOram:
Fixes #240

Requires https://github.com/DiamondLightSource/dodal/pull/785

To test:
* Confirm new test passes

### DIFF
--- a/src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
+++ b/src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
@@ -187,7 +187,7 @@ def raise_exception_if_moved_out_of_cryojet(exception):
         )
 
 
-def pin_already_loaded(
+def _pin_already_loaded(
     robot: BartRobot, pin_to_load: int, puck_to_load: int
 ) -> Generator[Msg, None, bool]:
     current_puck = yield from bps.rd(robot.current_puck)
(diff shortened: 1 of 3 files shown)


src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -185,11 +187,21 @@ def raise_exception_if_moved_out_of_cryojet(exception):
...
+    current_pin = yield from bps.rd(robot.current_pin)
+    return int(current_puck) == puck_to_load and int(current_pin) == pin_to_load
+
+
 def robot_load_then_centre_plan(
     composite: RobotLoadThenCentreComposite,
     params: RobotLoadThenCentre,
 ):
-    yield from prepare_for_robot_load(composite)
+    # TODO: get these from one source of truth #1347
```
rtuck99, line 202: This is now mx-bluesky issue 254

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -245,8 +257,17 @@ def robot_load_and_snapshots():
 
         yield from bps.wait("reset-lower_gonio")
 
-    yield from robot_load_and_snapshots()
-
+    if not (
```
rtuck99, line 260: Perhaps we should consider moving `robot_load_and_snapshots()` out of the function similar to `prepare_for_robot_load()` - it's quite long I think it's affecting readbility.

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -185,11 +187,21 @@ def raise_exception_if_moved_out_of_cryojet(exception):
         )
 
 
+def pin_already_loaded(
```
rtuck99, line 190: should be private

rtuck99 (approved):
Approved

PR #496, opened 2024-09-16
Only use config server if GDA didn't supply params

### LINKED ISSUE #494: Decide how turn on experimental features
In `config_server`, we have a few parameters which default to `True `or `False`. These parameters can also be set in GDA and passed into Hyperion's parameter model. We should be clear about how they're set

I think that for now, if the GDA parameter for, eg, `use_panda` is set, then this value should be used. If it's not supplied, then the config server should be used

## Acceptance Criteria
- Above logic is implemented

olliesilvester:
Fixes #494 

Constant values for feature flags now only exist in one place, and `FeatureFlags` now stores which parameters it was instantiated with so it can prioritise GDA-supplied values

This now requires a GDA change too: instead of setting `hyperion_params["use_cpu_and_gpu_zocalo"]`  and `hyperion_params["use_panda"]`, these features need to be specified in a dictionary. Eg, `hyperion_params["features"] = {"use_cpu_and_gpu_zocalo":True, "use_panda":True}`

# Acceptance criteria
If GDA parameters are supplied for gpu, panda and stub_offsets, these are used. If the parameters are missing, the config server is used instead


src/mx_bluesky/hyperion/parameters/gridscan.py
```diff
@@ -31,20 +33,30 @@
...
     selected_aperture: ApertureValue | None = Field(default=ApertureValue.SMALL)
 
+    @model_validator(mode="before")
+    @classmethod
+    def set_default_feature_flags(cls, values) -> Any:
+        cls.features = FeatureFlags()
+        if "use_panda" not in values:
+            values["use_panda"] = cls.features.best_effort().use_panda_for_gridscan
+        if "use_gpu" not in values:
+            values["use_gpu"] = cls.features.best_effort().use_gpu_for_gridscan
```
dperl-dls, line 57: Should: Only fetch the flags from the server once and only define them in one place.

With something like:

```suggestion
    def set_default_feature_flags(cls, values) -> Any:
        cls.features = FeatureFlags.best_effort()
        if "use_panda" in values:
            cls.features.use_panda_for_gridscan = values["use_panda"]
        ...
```

then the new definitions above could be removed?

dperl-dls (changes requested):

src/mx_bluesky/hyperion/parameters/gridscan.py
olliesilvester: I agree to only get the flags from the server once, but I think the definitions for, eg `use_panda` still need to be in `GridCommon`, just having them in the validator isn't enough for pydantic to know that `GridCommon` has these attributes

dperl-dls: oh, of course, sorry. but `use_panda` still shouldn't be in `GridCommon`, the flags should be passed in in `features`, which is a `BaseModel`
in the validator for `features` you can then start with FeatureFlags.best_effort() and update anything which is passed in

olliesilvester: Okay, that works, does that mean we should change the logic of `best_effort`? Right now it checks the URL and uses the constants if it cant find anything. But best efforts should
1. Check if GDA supplied values for each parameter, if yes then use that
2. Check if parameter exists in URL, if yes then use that
3. Use Hyperion constants as a last resort

Is that reasonable?

dperl-dls: I don't think you need to change the logic in `best_effort` to achieve that priority order, just set the defaults in teh `FeatureFlags` class to the constants. maybe add a new layer to `I03Constants` to hold defaults for feature flags.

dperl-dls: but if you want to add the ability to override stuff in best_effort with kwargs that sounds fine to me

olliesilvester: I think we still want GDA parameters to have priority. We use `best_effort` in our plans so these should also give us the GDA parameters when they're supplied

olliesilvester: I think I've changed my mind actually, I think the config server should take priority once it's actually being used

dperl-dls (approved):
This is great, thanks for reworking it! Minor comments about the comments, up to you

PR #504, opened 2024-09-18
Move transmission change in rotation to after detector unstaged

### LINKED ISSUE #500: Hyperion testing 17/09
* GDA was turning XBPM feedback off before passing to hyperion, hyperion then timed out waiting on feedback being stable - fixed by stopping GDA from touching xbpm feedback and transmission when Hyperion rotations are enabled - https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/42611
* Hyperion is switching transmission back to 100% before rotation has completely finished - fixed by moving the changing transmission to earlier - fixed in https://github.com/DiamondLightSource/mx-bluesky/pull/504
* Hyperion does not reset robot errors and so will not work if there was an error last time - we should be able to happily reset error 40 as this is just someone was near the robot previously - https://github.com/DiamondLightSource/mx-bluesky/issues/505

DominicOram:
Fixes #500

To test:
* Confirm new test works
* Confirm all points in the issue were fixed or new issues raised

### DIFF
--- a/src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
+++ b/src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
@@ -205,11 +205,6 @@ def rotation_scan_plan(
             "scan_points": [params.scan_points],
         }
     )
-    @transmission_and_xbpm_feedback_for_collection_decorator(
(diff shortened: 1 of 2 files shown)


noemifrisina (approved):
lgtm

PR #506, opened 2024-09-18
Read aperture properly rather than private method

DominicOram:
See https://github.com/DiamondLightSource/dodal/issues/782

Requires dodal version https://github.com/DiamondLightSource/dodal/pull/789

To test:
* Confirm test passes


tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
```diff
@@ -239,8 +239,7 @@ async def test_rotation_plan_moves_aperture_correctly(
         run_full_rotation_plan.aperture_scatterguard
     )
     assert (
-        await aperture_scatterguard.get_current_aperture_position()
-        == ApertureValue.SMALL
+        await aperture_scatterguard.selected_aperture.get_value() == ApertureValue.SMALL
```
DiamondJoseph, line 242: `selected_aperture` is the sole HintedSignal of `aperture_scatterguard`, you should be able to just `aperture_scatterguard.read()`?

DominicOram: I'm not sure I see the advantage of this as it's more verbose and now makes the test depend on the fact the signal is hinted but don't care so much so have changed it. Are you suggesting it because `read()` is the more externally facing API over `get_value()`?

DiamondJoseph: Looking at the change to the test I think I was just wrong?  I thought that
`selected_aperture.read() == ApertureValue.SMALL` 

And then we could ensure that the selected_aperture was always readable and always gave the ApertureValue as its output.

DominicOram: Yh, I think even reading the one signal it will still return a dict with the `name` and `value`.

But the test shouldn't care if it's readable or hinted. The test is "does doing this plan change the aperture in the expected way". We have other tests that are "can I read the selected aperture in the way I expect" like https://github.com/DiamondLightSource/mx-bluesky/blob/1dc26a58dfe242cdb5895ca9029f6a11c60edf51/tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py#L301

DiamondJoseph (approved):
Sorry re: confusion with the scatterguard read: I think it was neater before my previously requested changes but either way the change is more correct than before and either is compatible with the patched version of dodal

PR #507, opened 2024-09-18
Miscellaneous kubernetes deployment fixes

rtuck99:
* Enable hostNetwork in helmchart in order to access instrumentation network
* Fix version extraction from folder path
* Change image name to mx-bluesky to match the release CI workflow
* Fix helmchart default log folder to not clash with non-containerised version
* Do final docker image pip install with no-deps to prevent editable dodal being clobbered


helmchart/templates/deployment.yaml
```diff
@@ -111,3 +150,4 @@ spec:
             name: logs
           - mountPath: "/dls/{{ .Values.hyperion.beamline }}/data"
             name: data
+      hostNetwork: true
```
DominicOram, line 153: @callumforrester or @DiamondJoseph why do we need this to talk to EPICS but `blueapi` seems fine without it?

helmchart/templates/deployment.yaml
```diff
@@ -57,7 +57,46 @@ spec:
...
+              - SYS_MODULE
+              - SYS_NICE
+              - SYS_PACCT
+              - SYS_MODULE
+              - SYS_PTRACE
+              - SYS_RESOURCE
+              - SYS_TIME
+              - SYS_TTY_CONFIG
+              - SYSLOG
+              - WAKE_ALARM
```
DominicOram, line 99: Should: As discussed, this may not be necessary or it may be better to drop ALL. Can you make an issue to chase up, feel free to add other k8s tidying to that issue if you think there's some needed?

DominicOram (approved):
Great, thanks. Happy to merge as is with the spun off issue. Would be good to understand `hostNetwork` better but not urgent.

helmchart/templates/deployment.yaml
rtuck99: Raised #534

PR #517, opened 2024-09-23
Run code coverage with lower target on I24 SSX

### LINKED ISSUE #508: Have lower codecov targets for i24 SSX
Currently we have a 85% codecov target across the board. We are failing to hit this because the i24 SSX code has ~50% coverage. Because this code came from elsewhere originally we know that the coverage on it is low but improving. We should separate the repo up so that in general we should have 85% coverage but give an exception for SSX. We should be able to do this with codecov components https://docs.codecov.com/docs/components

## Acceptance Criteria
- Codecov targets are lower on SSX code than the rest of the codebase

DominicOram:
Fixes #508 

To test:
* Expand codecov report below, confirm that SSX has a 55% coverage target and the rest of the repo is unchanged at 85%


noemifrisina (approved):
Thank you

PR #519, opened 2024-09-23
Do not centre if chi has not changed

### LINKED ISSUE #483: Make sure not to run centring between screening and full collection
Currently if screening comes back saying to do a full collection on the current pin then centring is run again. This is unnecessary and should not be done.

Note: This may now be easier solved by making a specific screening plan rather than changing GDA/hyperion interactions

## Acceptance Criteria
- Centring is not done again if a screening comes back as should be doing a full collection

DominicOram:
Fixes #483

To test:
* Confirm new tests will catch the case of `robot_load_then_centre` being called before screening and will do nothing


src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -238,53 +238,73 @@ def robot_load_and_snapshots(
...
-                    "RobotLoadISPyBCallback",
-                ],
-            },
-        )
-    else:
-        LOGGER.info(
-            f"Pin/puck {params.sample_pin}/{params.sample_puck} already loaded, will not reload."
-        )
-    yield from pin_centre_then_xray_centre_plan(
+    return pin_centre_then_xray_centre_plan(
```
rtuck99, line 245: I think this should be `yield from` not return - although it doesn't make any difference for now, if you ever did anything else that yielded inside this method you would have to change the return to a `yield from` as well

rtuck99 (approved):
Approved apart from minor comment

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
DominicOram: Thanks, my bad

PR #524, opened 2024-09-24
Remove broken fixture cleanup code

rtuck99:
This just removes some code that cleaned up temporary files created by unit tests.
However, it doesn't work for the following reasons:

* glob("{foo}*") only returns /foo unless it ends with a /
* if any of the files returned is a directory, `os.remove()` fails
* The tests never created any files in the first place - *but they do now*
* The directory supplied in the params might be somewhere we don't want to delete and we don't check for this.


DominicOram (approved):
Great, thank you!

PR #531, opened 2024-09-26
Fix deploy scripts

noemifrisina:
- wrong/broken paths
- repo name
- updates docs

In the future there should be only one, see #532


docs/developer/general/how-to/deploy-a-release.rst
```diff
@@ -1,25 +1,63 @@
...
+
+
+.. code:: console
+
+    python ./utility_scripts/deploy/deploy_hyperion.py dev
+
+
+and a released version will be put in ``/scratch/30day_tmp/hyperion_release_test``.
+
+For building and deploying a Docker image please see :doc:`../../hyperion/deploying-hyperion`.
```
rtuck99, line 63: Can you please add a note for deploying on i03:

post-installation step: 
On i03 installation will succeed with error messages due to RedHat 7 versions of a dependency being unavailable. However the installation will be incomplete.

To fix this, on a RH 8 workstation
```
. ./.venv/bin/activate
pip install confluent-kafka
```

Then on the control machine
```
. ./.venv/bin/activate
pip install -e.
```

rtuck99: then 
```
pip install -e ../dodal
```

utility_scripts/deploy/deploy_hyperion.py
```diff
@@ -141,7 +141,7 @@ def _create_environment_from_control_machine(
 
 def main(options: Options):
     release_area = options.release_dir
-    this_repo_top = os.path.abspath(os.path.join(os.path.dirname(__file__), "../.."))
+    this_repo_top = os.path.abspath(os.path.join(os.path.dirname(__file__), "../../.."))
```
rtuck99, line 144: I don't think we actually need this change - `../..` will be the project directory which contains the .git folder, mx-bluesky needs it because it was `../.git` rather than `../../.git`
I think I only ran into this issue bc I initially ran the wrong script

rtuck99 (approved):
Approved subject to suggested changes

Issue #231, opened 2024-06-21
XRC: Ensure easy way to change box sizes and exposure time


DominicOram:
It may be useful to oversample the XRC and take data with smaller box sizes (even though the aperture is bigger than this). This should give us a more precise location for the centre of mass.

## Acceptance Criteria
* There is an easy way for the user to change box size and exposure time
* We communicate how to do this to the user
* This does snot affect aperture selection
* We confirm on the beamline that we still go to the correct place even with smaller box sizes

rtuck99:
Additional issue is that `flyscan_xray_centre` currently determines aperture on the basis of zocalo results as measured in boxes rather than physical size, so the aperture changes dependent on box size.

DominicOram:
Needs testing on the beamline

rtuck99:
Closing this, as the only outstanding work is
* #618

PR #542, opened 2024-09-30
Adds a box width parameter that can be passed to grid detect

DominicOram:
Goes some way to #231 but should try again on the beamline

## To test
* Confirm tests pass


DominicOram:
I think given it will be a while before we can test this we review/merge with the idea of checking if we think it will fix the issue and retest next run

olliesilvester (changes requested):
Changes look good, but from @rtuck99's comment on https://github.com/DiamondLightSource/mx-bluesky/issues/231, I'm not 100% sure if this will work

From my understanding, after doing FGS, Hyperion sets the aperture size based off the bounding box size returned from Zocalo. The bounding box size is in units of box size, but where do we tell Zocalo about the box size?

Also, we should define box size and bbox size somewhere, I got confused about it just now

DominicOram:
> Changes look good, but from @rtuck99's comment on #231, I'm not 100% sure if this will work
> 
> From my understanding, after doing FGS, Hyperion sets the aperture size based off the bounding box size returned from Zocalo. The bounding box size is in units of box size, but where do we tell Zocalo about the box size?
> 
> Also, we should define box size and bbox size somewhere, I got confused about it just now

I agree with this but the bounding box code is also in flux with https://github.com/DiamondLightSource/mx-bluesky/pull/553 at the moment. To avoid merge conflicts on that (already big) PR I have made https://github.com/DiamondLightSource/mx-bluesky/issues/618

olliesilvester (approved):

PR #550, opened 2024-10-03
Configure both inputs on Zebra auto AND gate during setup

### LINKED ISSUE #546: XRC not opening shutter on first or second grid scan
With Hyperion running on mx-bluesky v1.1.0, XRC grid scans didn't open the fast shutter. see https://ispyb.diamond.ac.uk/dc/visit/cm37235-4/dcg/13008691. This was caused by the changes done with the ZebraShutter and auto/manual mode

## Acceptance Criteria
- Fix bug and add tests
- Create a diagram which explains how the Zebra + shutter should be configured

olliesilvester:
Fixes #546 

Requires dodal change: https://github.com/DiamondLightSource/dodal/pull/823

Also attempts to make the ZebraShutter logic easier to understand

### DIFF
--- a/src/mx_bluesky/hyperion/device_setup_plans/setup_zebra.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/setup_zebra.py
@@ -75,13 +75,43 @@ def tidy_up_zebra_after_rotation_scan(
 
 
 def set_shutter_auto_input(zebra: Zebra, input: int, group="set_shutter_trigger"):
-    """Set the input that the shutter uses when set to auto.
+    """Set the signal that controls the shutter when set to auto. We use the second input to the
+    Zebra's AND2 gate for this input. Shutter must be in auto mode for this input to take control
 
     For more details see the ZebraShutter device."""
     auto_shutter_control = zebra.logic_gates.and_gates[AUTO_SHUTTER_GATE]
     yield from bps.abs_set(
-        auto_shutter_control.sources[AUTO_SHUTTER_INPUT_1], input, group
+        auto_shutter_control.sources[AUTO_SHUTTER_INPUT_2], input, group
     )
+
+
+def configure_zebra_and_shutter_for_auto_shutter(
+    zebra: Zebra, zebra_shutter: ZebraShutter, input: int, group="use_automatic_shutter"
+):
+    """Set the shutter to auto mode, and configure the zebra to trigger the shutter on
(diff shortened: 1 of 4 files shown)


src/mx_bluesky/hyperion/device_setup_plans/setup_zebra.py
```diff
@@ -73,15 +75,45 @@ def tidy_up_zebra_after_rotation_scan(
...
-        auto_shutter_control.sources[AUTO_SHUTTER_INPUT], input, group
+        auto_shutter_control.sources[AUTO_SHUTTER_INPUT_2], input, group
     )
 
 
+def configure_zebra_and_shutter_for_auto_shutter(
+    zebra: Zebra, zebra_shutter: ZebraShutter, input: int, group="use_automatic_shutter"
+):
+    """Set the shutter to auto mode, and configure the zebra to trigger the shutter on
+    an input source. For the input, use one of the source constants in zebra.py
```
DominicOram, line 92: Nit: I'm not a big fan of a comment like `For the input, use one of the source constants in zebra.py` I think it will just easily go out of date if the file is renamed.

src/mx_bluesky/hyperion/device_setup_plans/setup_zebra.py
```diff
@@ -73,15 +75,45 @@ def tidy_up_zebra_after_rotation_scan(
...
+
+    When the shutter is in auto/manual, logic in EPICS sets the Zebra's
+    SOFT_IN1 to low/high respectively. The Zebra's AND2 gate should be used to control the shutter while in auto mode.
+    To do this, we need (AND2 = SOFT_IN1 AND input), where input is any zebra signal
```
DominicOram, line 96:
```suggestion
    To do this, we need (AND2 = SOFT_IN1 AND input), where input is the zebra signal we want to control the shutter when in auto mode.
```

DominicOram (approved):
Couple of nits but generally good. Thanks!

src/mx_bluesky/hyperion/device_setup_plans/setup_zebra.py
olliesilvester: Yeah, I agree. I added a comment further down which links to an issue which would make specific typing for the Zebra inputs, so then this comment to clarify wouldn't be needed

Issue #475, opened 2024-09-12
Make a plan that will do an XRC and data collection in multiple places


DominicOram:
Get the results from XRC and feed the top N into a multiple rotation plan

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

DominicOram:
Fixed in https://github.com/DiamondLightSource/mx-bluesky/pull/553

PR #553, opened 2024-10-04
(#475) Xray-centre and collect on multiple centres

rtuck99:
Implements #475 

Requires 
~DiamondLightSource/dodal#814~
DiamondLightSource/dodal#858 and
DiamondLightSource/mx-bluesky#593

This implements multiple rotations on multiple centres for the `load_centre_collect_full_plan`.
The behaviour of other plans should remain unchanged, with moving to a single centre.
By default load_centre_collect also only does a single centre.
To specify multiple centres, include the following in the request parameters:

```
  "select_centres": {
    "name": "top_n_by_max_count",
    "n": 5
  },
```

where n is however many centres to select


src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -157,25 +162,59 @@ def flyscan_xray_centre(
...
+        yield from run_gridscan_and_fetch_results(
+            fgs_composite, params, feature_controlled
+        )
+
+    yield from run_gridscan_and_fetch_and_tidy(
+        composite, parameters, feature_controlled
+    )
+
+
+def flyscan_xray_centre(
```
DominicOram, line 179: I wonder if we should have a new place for "top level" plans, which is what this is really. It's hard to differentiate between this and `flyscan` and when you would use one and not the other. Probably needs a wider discussion

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -242,6 +262,38 @@ def run_gridscan_and_move(
         yield from bps.wait()
 
 
+def _xrc_result_to_flyscan_result(
```
DominicOram, line 265: Should: I think it is cleaner, and more idiomatic of bluesky, to move this calculation to the `FlyscanEventHandler`. It should be able to get everything it needs for the calculation by the events produced throughout the plan. This would reduce the complexity of the plans themselves and change `_fire_flyscan_result_event` to just be a `bps.read(zocalo)` Probably needs a discussion.

src/mx_bluesky/hyperion/experiment_plans/load_centre_collect_full_plan.py
```diff
@@ -41,6 +54,30 @@ def load_centre_collect_full_plan(
...
+        yield from robot_load_then_flyscan(composite, params.robot_load_then_centre)
+
+    yield from fetch_results_from_robot_load_then_flyscan()
+    assert (
+        flyscan_event_handler.flyscan_results
+    ), "Flyscan result event not received or no crystal found and exception not raised"
+
+    selection_params = params.select_centres
+    selection_func = getattr(flyscan_result, selection_params.name)  # type: ignore
+    assert callable(selection_func)
```
DominicOram, line 71: Should: I don't think letting the user provide a function name like this is very good. It's quite brittle to name changes/argument changes and I don't think we need this much flexibility right now anyway.

src/mx_bluesky/hyperion/experiment_plans/load_centre_collect_full_plan.py
```diff
@@ -41,6 +54,30 @@ def load_centre_collect_full_plan(
...
+        flyscan_event_handler.flyscan_results, **selection_args
+    )  # type: ignore
+    LOGGER.info(f"Selected hits {hits} using {selection_func}, args={selection_args}")
+
+    for hit in hits:
+        LOGGER.info(f"Performing rotations for {hit}")
+        yield from change_aperture_then_centre(hit, composite)
+        yield from multi_rotation_scan(
+            composite, params.multi_rotation_scan, oav_params
+        )
```
DominicOram, line 83: Must: I think what we actually want to do is take all the hits and wrap them up in one call to `multi_rotation_scan` with different x/y/z starts. Otherwise we're going to get the detector re-arming between them all etc. For a first pass this probably means just stay at the same aperture for each of them, which is fine.

src/mx_bluesky/hyperion/experiment_plans/pin_centre_then_xray_centre_plan.py
```diff
@@ -105,15 +111,27 @@ def pin_tip_centre_then_xray_centre(
     oav_config_file: str = OAV_CONFIG_JSON,
 ) -> MsgGenerator:
     """Starts preparing for collection then performs the pin tip centre and xray centre"""
+    # TODO listen for flyscan results and then centre
```
DominicOram, line 114: Should: We're doing this, no need for the comment

src/mx_bluesky/hyperion/experiment_plans/pin_centre_then_xray_centre_plan.py
```diff
@@ -105,15 +111,27 @@ def pin_tip_centre_then_xray_centre(
...
     eiger.set_detector_parameters(parameters.detector_params)
 
-    return start_preparing_data_collection_then_do_plan(
-        eiger,
-        composite.detector_motion,
-        parameters.detector_params.detector_distance,
-        pin_centre_then_xray_centre_plan(composite, parameters, oav_config_file),
-        group=CONST.WAIT.GRID_READY_FOR_DC,
-    )
+    flyscan_event_handler = FlyscanEventHandler()
```
DominicOram, line 120: Should: This pattern is used quite a bit now, I think it's probably worth making a wrapper of something like:
```python
def do_plan_and_move_to_xrc_result(plan):
    """Do the specified plan then move to whatever result XRC gives back"""
    yield from bpp.subs_wrapper(plan, flyscan_event_handler := FlyscanEventHandler())
    flyscan_results = flyscan_event_handler.flyscan_results
    assert (
        flyscan_results
    ), "Flyscan result event not received or no crystal found and exception not raised"
    yield from change_aperture_then_centre(flyscan_results[0], composite)
```

DominicOram (changes requested):
Thank you! This is great, we will see how it goes on Monday. Some comments in code, think a few places where we should probably all talk it through together.

Additionally, for multipin tests we're also probably going to need something that gives us the top result in each segment e.g. for a 100 box long multipin with 5 samples in we would want the top result in 0-20, 20-40 etc. We can do this on the beamline for the tests next week and then maybe pull it into a better defined issue.

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -112,26 +116,27 @@ def sample_motors(self) -> Smargon:
         return self.smargon
 
 
+class FlyscanEventHandler:
```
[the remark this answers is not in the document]
rtuck99: I think it's overkill for what should just be a simple event handler. What does it give us?

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -112,26 +116,27 @@ def sample_motors(self) -> Smargon:
...
+
 def create_devices(context: BlueskyContext) -> FlyScanXRayCentreComposite:
     """Creates the devices required for the plan and connect to them"""
     return device_composite_from_context(context, FlyScanXRayCentreComposite)
 
 
-def flyscan_xray_centre(
-    composite: FlyScanXRayCentreComposite,
-    parameters: ThreeDGridScan,
+def flyscan(
```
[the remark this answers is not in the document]
rtuck99: Hmm I think we should have a conversation about terminology and the use of the word "centring"

src/mx_bluesky/hyperion/experiment_plans/pin_centre_then_xray_centre_plan.py
rtuck99: I didn't partly as we are ultimately removing most of the places where it's used anyway

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
rtuck99: I started doing this and have run into complications with unit tests as now `ThreeDGridscan` needs to be injected in lots of places. Also I think I have run into additional complications due to #220 which previously wasn't an issue because the tests either didn't need to account for this bug or did it both ways which cancelled each other out.

tests/unit_tests/hyperion/experiment_plans/test_change_aperture_then_move_plan.py
```diff
@@ -0,0 +1,75 @@
...
+    test_fgs_params.features.set_stub_offsets = set_stub_offsets
+    msgs = sim_run_engine.simulate_plan(
+        change_aperture_then_move_to_xtal(
+            simple_flyscan_hit, smargon, aperture_scatterguard, test_fgs_params
+        )
+    )
+
+    msgs = assert_message_and_return_remaining(
+        msgs, lambda msg: msg.command == "set" and msg.args[0] == ApertureValue.MEDIUM
+    )
```
DominicOram, line 45: Should: Probably want to add `msg.obj == aperture_scatterguard` here

tests/unit_tests/hyperion/experiment_plans/test_change_aperture_then_move_plan.py
```diff
@@ -0,0 +1,75 @@
...
+    msgs = assert_message_and_return_remaining(
+        msgs,
+        lambda msg: msg.command == "set"
+        and msg.obj.name == "smargon-x"
```
DominicOram, line 49: Nit: I think it's nicer to use `msg.obj== smargon.x` here as it makes it easier to pick up with the IDE where the smargon is being used

tests/unit_tests/hyperion/experiment_plans/test_experiment_registry.py
```diff
@@ -19,6 +21,7 @@ def test_exposed_plans_in_reg():
         assert plan in PLAN_REGISTRY.keys()
 
 
+@pytest.mark.skip(reason="Conflicting MsgGenerator import strangeness")
```
DominicOram, line 24: Should: Could we link to an issue that will fix, maybe it's covered by https://github.com/DiamondLightSource/mx-bluesky/pull/564?

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -315,6 +315,61 @@ def test_read_hardware_for_ispyb_updates_from_ophyd_devices(
...
+            fgs_composite_with_panda_pcap,
+            test_fgs_params_panda_zebra,
+        )
+        RE, _ = RE_with_subs
+        event_doc = get_event_loop().create_future()
+
+        class FlyscanEventHandler(DocumentRouter):
+            def start(self, doc):
+                if "xray_centre_results" in doc:
+                    event_doc.set_result(doc["xray_centre_results"])
```
DominicOram, line 344: Could: Could we just use the real handler here and have this as more of an integration test?

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -521,10 +528,10 @@ def wrapped_gridscan_and_move():
...
             )
         )
-        assert (
-            await fgs_composite_with_panda_pcap.smargon.stub_offsets.center_at_current_position.proc.get_value()
-            == 1
-        )
+        # assert (
+        #     await fgs_composite_with_panda_pcap.smargon.stub_offsets.center_at_current_position.proc.get_value()
+        #     == 1
+        # )
```
DominicOram, line 534: Should: Why is this commented out? If it's broken can we at least point to an issue fixing it or ideally fix in this PR

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -708,88 +715,6 @@ def wrapped_gridscan_and_move():
...
 
-    @patch(
-        "mx_bluesky.hyperion.experiment_plans.flyscan_xray_centre_plan.run_gridscan",
-        autospec=True,
-    )
-    @patch(
-        "mx_bluesky.hyperion.experiment_plans.flyscan_xray_centre_plan.move_x_y_z",
-        autospec=True,
-    )
-    async def test_given_gridscan_fails_to_centre_then_stub_offsets_not_set(
```
DominicOram, line 719: Should: I think we still need a test that the stub offsets aren't set when we fail to diffract

tests/unit_tests/hyperion/experiment_plans/test_load_centre_collect_full_plan.py
```diff
@@ -168,6 +230,20 @@ def test_load_centre_collect_full_plan_happy_path_invokes_all_steps(
     rotation_scan_params = mock_rotation_scan.mock_calls[0].args[1]
     assert isinstance(rotation_scan_composite, RotationScanComposite)
     assert isinstance(rotation_scan_params, MultiRotationScan)
+    # XXX sample test file xyz conflicts with detected xyz
```
DominicOram, line 233: Should: I'm not sure what this comment is trying to tell me?

tests/unit_tests/hyperion/experiment_plans/test_pin_centre_then_xray_centre_plan.py
```diff
@@ -34,6 +40,46 @@ def test_pin_centre_then_xray_centre_params():
...
+)
+def test_pin_tip_centre_then_xray_centre_moves_to_centre_of_first_flyscan_result(
+    mock_change_aperture_then_move_to_xtal: MagicMock,
+    mock_detect_and_do_gridscan: MagicMock,
+    mock_pin_tip_centre: MagicMock,
+    test_pin_centre_then_xray_centre_params: PinTipCentreThenXrayCentre,
+    test_config_files,
+):
+    mock_detect_and_do_gridscan.side_effect = (
+        lambda _, __, ___: _fire_xray_centre_result_event(
```
DominicOram, line 63: Nit: 
```suggestion
        lambda *_: _fire_xray_centre_result_event(
```

DominicOram (changes requested):
Sorry this has taken me so long to get around to, some comments in code

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
DominicOram: I think the real way to fix this is actually https://github.com/DiamondLightSource/python-dlstbx/issues/255

src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py
```diff
@@ -0,0 +1,82 @@
...
+            # See: https://github.com/bluesky/bluesky/issues/1809
+            smargon.stub_offsets,  # type: ignore
+            StubPosition.CURRENT_AS_CENTER,  # type: ignore
+        )
+
+
+def _set_aperture_for_bbox_mm(
+    aperture_device: ApertureScatterguard, bbox_size_mm: list[float] | numpy.ndarray
+):
+    # TODO confirm correction factor see mx-bluesky 231
```
DominicOram, line 54: Should: I've actually spun this into https://github.com/DiamondLightSource/mx-bluesky/issues/618 to avoid merge conflicts

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -203,37 +256,17 @@ def run_gridscan_and_move(
...
-                        np.array(best_result["bounding_box"][1])
-                        - np.array(best_result["bounding_box"][0])
-                    )
-                    yield from set_aperture_for_bbox_size(
-                        fgs_composite.aperture_scatterguard, bbox_size
-                    )
+                flyscan_results = [
+                    _xrc_result_to_xray_centre_result(xr, parameters)
+                    for xr in xrc_results
+                ]
```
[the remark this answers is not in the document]
rtuck99: `GridscanISPyBCallback` does also get the XrcResults through an event document via a call to `get_processing_results_from_event()`, and it sends the grid coordinates to ispyb.

I think both types have value - the grid coordinates are of interest if you are interested in the gridscan analysis pipeline, and are quite useful if correlating to the annotated snapshot.

The absolute mm coordinates are needed for moving to the centre but are kind of useless on their own to an end user without a frame of reference.

src/mx_bluesky/hyperion/experiment_plans/common/xrc_result.py
```diff
@@ -0,0 +1,39 @@
...
+    unfiltered: Sequence[XRayCentreResult], n: int
+) -> Sequence[XRayCentreResult]:
+    sorted_hits = sorted(unfiltered, key=lambda result: result.max_count, reverse=True)
+    return sorted_hits[:n]
+
+
+def resolve_selection_fn(name: str):
+    if name == "TopNByMaxCount":
+        return top_n_by_max_count
+    raise ValueError(f"Invalid selection function {name}")
```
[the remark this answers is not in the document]
rtuck99: I can sort of do this but slightly uglier as Pydantic doesn't allow polymorphism with base classes - the fields have to be a type union instead.

tests/system_tests/hyperion/external_interaction/conftest.py
```diff
@@ -74,6 +81,23 @@
...
 
+def simulate_xrc_result(
+    sim_run_engine: RunEngineSimulator,
+    zocalo: ZocaloResults,
+    test_results: Sequence[dict],
+):
+    for k in test_results[0].keys():
+        sim_run_engine.add_read_handler_for(
+            getattr(zocalo, k), numpy.array([r[k] for r in test_results])
+        )
```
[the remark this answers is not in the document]
rtuck99: I think one of them comes from the ophyd-async changes merge and the other one was from this PR, as they both depended on the dodal changes  - I'll rationalise them.

tests/system_tests/hyperion/external_interaction/conftest.py
```diff
@@ -74,6 +81,23 @@
...
+
+
+def generate_xrc_result_event(device_name: str, test_results: Sequence[dict]) -> dict:
```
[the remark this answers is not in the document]
rtuck99: ditto

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
rtuck99: We used to set the stub offsets as part of the centring, I seem to remember we discussed this and decided to remove it as they are not used on i03.

I will remove the commented out code

tests/unit_tests/hyperion/experiment_plans/test_load_centre_collect_full_plan.py
rtuck99: The sample params file specifies x,y,z coordinates (because the model allows this). Whereas the plan gets the coordinates from the detected centres. I've added this to https://github.com/DiamondLightSource/mx-bluesky/issues/563 and will link in a comment

tests/system_tests/hyperion/external_interaction/conftest.py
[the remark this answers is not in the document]
DominicOram: I think this is the one that's not used so I will remove it

tests/unit_tests/hyperion/experiment_plans/test_load_centre_collect_full_plan.py
```diff
@@ -249,9 +326,197 @@ def test_load_centre_collect_full_plan_skips_collect_if_no_diffraction(
...
+    )
+    assert sum(1 for msg in msgs if msg.command == "robot_load_and_change_energy") == 1
+    msgs = assert_message_and_return_remaining(
+        msgs,
+        lambda msg: msg.command == "open_run" and "xray_centre_results" in msg.kwargs,
+    )
+    msgs = assert_message_and_return_remaining(
+        msgs, lambda msg: msg.command == "multi_rotation_scan"
+    )
+    expected_rotation_scans = [
```
DominicOram, line 427: Nit: This is a bit verbose, I took the liberty of cleaning it up a little

DominicOram (approved):
Great, thank you!

PR #554, opened 2024-10-07
I24 serial: Use async OAV for moveonclick

noemifrisina:
Use the ophyd async oav from [Dodal#815](https://github.com/DiamondLightSource/dodal/pull/815) in moveonclick.


DominicOram (approved):
Great, thanks!

PR #555, opened 2024-10-07
Make sure we enable dev/shm when required

### LINKED ISSUE #545: Hyperion testing 01/10
* Saw https://github.com/DiamondLightSource/dodal/issues/791 again
* Having the two `use_cpu_and_gpu_zocalo` and `use_gpu` is confusing and `enable_dev_shm` was only being set on `use_gpu`. Fixed by setting on the `or` of them both
* Issue where if the there was no diffraction and no results from zocalo the comparison of the results would fail

DominicOram:
Fixes #545 

See also https://github.com/DiamondLightSource/dodal/pull/819

To test:
* Confirm test fails on main and passes here


olliesilvester (changes requested):
Given that, right now, we only ever use the gpu results for comparison (rather than actually use them for analysis), I think we should just remove `use_cpu_and_gpu_zocalo` - we can use `use_gpu` for this too.

Also would be good to add a comment saying that `use_gpu=True` means dev/shm is enabled and that a comparison is made between CPU and GPU results

DominicOram:
As discussed, naming is confusing, we should remove`use_gpu` for now and rename `use_cpu_and_gpu_zocalo` to `compare_cpu_and_gpu_results` then only turn `dev/shm` on when `compare_cpu_and_gpu_results` is `True`

olliesilvester (approved):
Looks good. I'm just going to add a comment to the constant before merging

PR #560, opened 2024-10-09
Test rotation callbacks better

### LINKED ISSUE #479: Test the rotation callbacks against the real rotation plan
A number of the issues found in https://github.com/DiamondLightSource/mx-bluesky/issues/463 were not caught because the tests in `test_rotation_callbacks` are run against a `fake_rotation_scan` rather than the real one, which means that we have no tests covering the communication between the real rotation plan and the real callbacks. https://github.com/DiamondLightSource/mx-bluesky/pull/477 adds a couple but it would be better if `test_rotation_callbacks` used the real plan.

## Acceptance Criteria
- `test_rotation_callbacks` tests against the real plan (using the simulator should still be a good test)
- New tests added in https://github.com/DiamondLightSource/mx-bluesky/pull/477 are removed if they are now covered

DominicOram:
Fixes #479 

Needs https://github.com/DiamondLightSource/dodal/pull/831

To test:
* Confirm tests still pass
* Confirm tests test teh same functionality but this time against production code


dperl-dls (approved):
Great, simpler and more comprehensive.

PR #562, opened 2024-10-10
Fix bluesky upgrade typing

dperl-dls:
Add a bunch of `type: ignore`s linked to https://github.com/bluesky/bluesky/issues/1809
When upstream issues are resolved there may still be things to fix here, e.g. the `EigerDetector.stop()`


DominicOram (approved):
Well this is horrible. But thank you.

dperl-dls:
Yeah, at least they're all identical so they'll be easy to remove for testing the fixes in bluesky and when we're done

PR #564, opened 2024-10-10
Defer to bluesky.utils MsgGenerator

DiamondJoseph:
Companion change to https://github.com/DiamondLightSource/dodal/pull/832

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,7 +47,6 @@ dependencies = [
     "ophyd-async >= 0.3a5",
     "bluesky >= 1.13.0a4",
     "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
-    "dls-bluesky-core",
 ]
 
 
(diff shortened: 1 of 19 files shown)


DominicOram (changes requested):
I think the test failure is real and implies that `blueAPI` is not picking up plans annotated with the `MsgGenerator` from `bluesky`. Do we need to update `blueAPI` as part of this?

DominicOram:
@DiamondJoseph, what is the state of this? I think it may be the more long term fix for https://github.com/DiamondLightSource/mx-bluesky/pull/656. Would you like one of MX to pick it up?

DiamondJoseph:
@DominicOram now that blueapi 0.6.1 has been released with support for bluesky>=1.13 this should be good to go after merge errors resolved. I'll quickly knock those on the head now.

DominicOram (approved):
Great, thank you!

PR #565, opened 2024-10-10
Set voltages on all mirrors on energy change

### LINKED ISSUE #521: Energy Changes: Set HFM voltages as well as VFM ones
When changing stripe for an energy change we must change both the HFM and VFM voltages even though we only change the stripe on the VFM. These should be set using the values found in the lookup table that we're already using for the VFM.

## Acceptance Criteria
- When the VFM stripe is changed the HFM voltages are also changed

DominicOram:
Fixes #521

Requires https://github.com/DiamondLightSource/dodal/pull/834

To test:
* Confirm tests pass


rtuck99 (approved):
Approved. 
Although curious to know what is the thinking was behind merging horizontal and vertical voltages into a single device? The mirrors themselves are still separate devices. Is it always going to be the case that horizontal and vertical mirrors will be controlled using the same power supply mechanism on all different beamlines?

DominicOram:
> Although curious to know what is the thinking was behind merging horizontal and vertical voltages into a single device? The mirrors themselves are still separate devices. Is it always going to be the case that horizontal and vertical mirrors will be controlled using the same power supply mechanism on all different beamlines?

It wasn't that deliberate a decision. They are powered by the same physical hardware and so share an IOC/EDM. They're also mostly changed at the same time i.e. this ticket. So I think having them together is fine but I don't mind breaking them up if you feel strongly about it?

olliesilvester (approved):
Looks good. Probably need further discussion on `rebuild_oa_device_as_mocked_if_necessary`, but better to having something working for now

PR #566, opened 2024-10-14
Put wavelength into ispyb earlier

### LINKED ISSUE #495: Put wavelength in ispyb earlier
For analysis zocalo reads the wavelength out of ispyb. We have seen a race condition where, because the wavelength is supplied during collection rather than before hand, zocalo sometimes falls over without finding one.

We should instead put what we think the wavelength will be in to ispyb pre collection then (optionally) update it later.

## Acceptance Criteria
- A wavelength is put in ispyb before collection

DominicOram:
Fixes #495.

Reads the dcm energy before collection and updates ispyb accordingly

## To Test
* Confirm tests pass


noemifrisina (approved):
lgtm

rtuck99 (approved):
Approved

PR #567, opened 2024-10-14
Remove some warning ignores that are no longer needed

DominicOram:
To test:
* Run `./utility_scripts/dls_dev_env.sh` and confirm that the environment is still set up correctly


rtuck99 (approved):
Approved

PR #571, opened 2024-10-15
Update from beamline test 14102024

noemifrisina:
- Update the cs_maker.json scale factors for the coordinate system after recalibration
- Fix y direction in moveonclick (again) and test


src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_moveonclick.py
```diff
@@ -52,7 +52,7 @@ def _move_on_mouse_click_plan(
     beamX, beamY = yield from _get_beam_centre(oav)
     x, y = clicked_position
     xmove = -1 * (beamX - x) * zoomcalibrator
-    ymove = -1 * (beamY - y) * zoomcalibrator
+    ymove = 1 * (beamY - y) * zoomcalibrator
```
DominicOram, line 55: As discussed, the theory is that this may be due to the rotation of the grid. Could you make an issue to investigate the reasoning behind this and find a way to fix it permanently please?

DominicOram (approved):
Great, thanks

PR #572, opened 2024-10-15
Tidy up i24 serial logs

### LINKED ISSUE #557: I24 ssx: tidy up log configuration

noemifrisina:
- Use common log configuration
- Fix issue with log messages being written to file multiple times
- Set level of stream handler to INFO
- Only integrate bluesky and ophyd_async loggers, as no ophyd devices are used for serial anymore

Closes #557


noemifrisina:
Deferred to after #573 is done and running on the beamline because of permission problems if run by staff/users

olliesilvester (changes requested):
I know this wasn't actually finished yet, but I've added some comments anyway

src/mx_bluesky/beamlines/i24/serial/log.py
```diff
@@ -103,7 +106,7 @@ def config(
         integrate_all_logs=False,
     )
     # Remove dodal StreamHandler to avoid duplication of messages above debug
-    dodal_logger.removeHandler(dodal_logger.handlers[0])
+    # dodal_logger.removeHandler(dodal_logger.handlers[0])
```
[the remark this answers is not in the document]
noemifrisina: That was accidentally commented out, sorry, fixed it now

src/mx_bluesky/beamlines/i24/serial/log.py
```diff
@@ -126,3 +129,23 @@ def decorator(*args, **kwargs):
         return func(*args, **kwargs)
 
     return decorator
+
+
+def setup_collection_logs(expt: str, dev_mode: bool = False) -> MsgGenerator:
```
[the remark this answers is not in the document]
noemifrisina: As this is not the only start up bit that will need changing in the future, between web GUIs and once we get a cluster, I'm going to write a bigger ticket to include all the changes needed, this one included.

src/mx_bluesky/beamlines/i24/serial/log.py
```diff
@@ -1,39 +1,30 @@
...
-        return "ophyd" not in record.getMessage().lower()
+VISIT_PATH = Path("/dls_sw/i24/etc/ssx_current_visit.txt")
 
 
 # Logging set up
-logger = logging.getLogger("I24ssx")
-logger.addHandler(logging.NullHandler())
-logger.parent = dodal_logger
+SSX_LOGGER = logging.getLogger("I24serial")
+SSX_LOGGER.addHandler(logging.NullHandler())
```
olliesilvester, line 21: What's the `NullHandler` for?

src/mx_bluesky/beamlines/i24/serial/log.py
```diff
@@ -68,10 +58,8 @@ def _read_visit_directory_from_file() -> Path:
 
 
 def _get_logging_file_path() -> Path:
```
olliesilvester, line 60: I think we might have already discussed this, but I can't remember. How come this needs to be in a different location to that given by the common util function (`/dls_sw/{beamline}/logs/bluesky/)`?

src/mx_bluesky/beamlines/i24/serial/run_fixed_target.sh
```diff
@@ -25,14 +25,22 @@ if [[ $NO_PROCESERV_TEST == true ]]; then
...
 fi
 
+echo "Set up logging configuration"
+blueapi -c "${current}/blueapi_config.yaml" controller run setup_collection_logs '{"expt":"Serial Fixed"}'
+
 # Open the edm screen for a fixed target serial collection
 echo "Starting fixed target edm screen."
 edm -x "${edm_path}/FT-gui/DiamondChipI24-py3v1.edl"
 
-echo "Edm screen closed, bye!"
```
olliesilvester, line 32: no more `bye!` :(

olliesilvester (approved):
Thanks, added some small comments. There are quite a few lines with no test coverage, but I will approve anyway as I know you're keen to test a release today.

A lot of the the untested lines are logging statements, I'd suggest testing them using by setting up logging in `pytest_runtest_setup(item)` in an i24 conftest, similiar to what we have in `mx-bluesky/tests/conftest.py`. Then we can easily assert logging calls after a`@patch("mx_bluesky.beamlines.i24.serial.log.SSX_LOGGER")`

src/mx_bluesky/beamlines/i24/serial/log.py
noemifrisina: SO that users/staff can access it, this is the logger that looks like the old output they are used to. May not be needed for long, but for the moment it's useful for debugging

src/mx_bluesky/beamlines/i24/serial/run_fixed_target.sh
noemifrisina: Well... it's three lines or so later :)

Issue #62, opened 2023-11-24
I24 ssx: use detector devices from dodal


noemifrisina:
Use devices from dodal to trigger the detector for serial. Detector used: eiger 9M and pilatus 6M (second one needs to be created in dodal first)

noemifrisina:
Update:
- Eiger: waiting for work on fastcs eiger in ophyd_async to be completed
- Pilatus: detector going away soon-ish, so probably not needed
- Jungfrau (once functionality added to serial): incoming new detector, there's a device for the 1M in the commissioning branch that could be reused once updated to ophyd-async. Once 9M arrives, will also need updating

PR #577, opened 2024-10-16
Set up i24 serial to run on procserv

### LINKED ISSUE #573: I24ssx: set up scripts to start procserv
A small python ioc has been set up on a couple fo beamline workstation to start the blueapi server through procserv.
Include the startIOC.sh script used for this and modify the current run_{extruder,fixed_target}.sh files to not start the blueapi server unless requested.

noemifrisina:
Closes #573 

- Add a script to start blueapi in the IOC
- Add a test option for debugging to the bashscripts for each experiment type
- Remove pilatus hack for eiger collections and just make the directory (and fix tests)

~~Needs more testing on the beamline~~

- Now tested on the beamline, adds a fix for the collection time which was severely underestimated


src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -85,7 +85,8 @@ def calculate_collection_timeout(parameters: FixedTargetParameters) -> float:
     Returns:
         The estimated collection time, in s.
     """
-    buffer = 30
+    buffer = PMAC_MOVE_TIME * parameters.total_num_images + 2
+    # buffer = 30
```
olliesilvester, line 89: Did you mean to leave this comment in? It might be explain why we need a buffer for the timeout

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -56,10 +56,10 @@
 )
 from mx_bluesky.beamlines.i24.serial.write_nexus import call_nexgen
 
-ABORTED = False
-
 logger = logging.getLogger("I24ssx.fixed_target")
 
+PMAC_MOVE_TIME = 0.008  # Move time between positions on chip ~ 7-8 ms
```
olliesilvester, line 61: Will be good to have constants like this moved into more organised locations, but perhaps better to wait for me to finish my work on this

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -474,24 +475,8 @@ def start_i24(
     elif parameters.detector_name == "eiger":
```
olliesilvester, line 475: Nit: There is some duplicated code between this file and `i24ssx_Extruder_Coolect_py3v2.py`

src/mx_bluesky/beamlines/i24/serial/run_serial.py
```diff
@@ -6,6 +7,13 @@
 logger = logging.getLogger("I24ssx.run")
 
 
+def _parse_input(expt: str):
```
olliesilvester, line 10: If this file is the main entry point, it would be nice to be able to run this file directly, eg with a 

```
if __name__ == "__main__":
   ...
```

utility_scripts/procserv_ioc_start.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Sources the python environment and starts a blueapi server
+# on a python IOC to run serial on I24 with procserv
+
+# Get the directory of this script
+current_directory=$( realpath "$( dirname "$0" )" )
```
olliesilvester, line 7: Do we need protection against running this script when the ioc and/or when blueapi is already running?

olliesilvester (approved):
Looks good, I've pressed approve but some of my comments may need addressing before merging, have a read and see if any of them are urgent enough

src/mx_bluesky/beamlines/i24/serial/run_serial.py
noemifrisina: Ah no, the entry points are the two functions (see in pyproject), which is why we don't need to use if __name__ etc... it would actually make the code slightly more complicated

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Yeah, that probably won't be fixed until we get around to #62 and we can finish tidying up the code...

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Good point. In this case I was also half considering having this constant in `dodal`, but I don't have a strong opinion on this. Will make a note in the code to do this

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Ah, no, good spot

PR #580, opened 2024-10-16
Make common MX bluesky logging

### LINKED ISSUE #574: Make common MX logging
Move the logging that Hyperion uses into somewhere more general, so that other MX beamlines can use the same code with minimal configuration

## Acceptance Criteria
- Hyperion logs moved into `common` with sensible configuration options
- Hyperion and i24 use these logs instead of their own implementations

olliesilvester:
Fixes #574 

I will either put the below instructions in the docs or as a helper function:

A beamline should use these logs by:
1. Creating a new global logger, eg `LOGGER = logging.getLogger("Hyperion")`
2. Importing the dodal logger and setting this as the parent
3. Running `do_default_logging_setup`

eg:
```
from dodal.log import LOGGER as dodal_logger
from mx_bluesky.common.utils.log import do_default_logging_setup
LOGGER = logging.getLogger("Hyperion")
LOGGER.parent = dodal_logger
do_default_logging_setup(...)
```


noemifrisina:
> I will either put the below instructions in the docs or as a helper function

Something in the developer docs about how the logs work and how to set them up would definitely be helpful - thinking mainly in terms of having a reference for the future.

src/mx_bluesky/common/utils/log.py
```diff
@@ -0,0 +1,108 @@
...
+    tag_filter.dc_group_id = dcgid
+
+
+def set_uid_tag(uid):
+    """Set the unique id as a tag on all subsequent log messages.
+    Setting to None will remove the tag."""
+    tag_filter.run_uid = uid
+
+
+def do_default_logging_setup(file_name: str, graylog_port: int, dev_mode: bool = False):
```
olliesilvester, line 48: @noemifrisina has requested an option to toggle whether or not this function runs `integrate_bluesky_and_ophyd_logging`, so I'll add that

DominicOram:
Can we replace some of the code in `i24/serial/log.py` with calls to this common one to make it consistent?

src/mx_bluesky/common/utils/log.py
olliesilvester: From what @noemifrisina  was saying (but correct me if I'm wrong), it's useful to see the sys.sterr logs on i24, and it gets cluttered with ophyd+bluesky+ophyd-async logs. Also, she was seeing an issue where ophyd debug logs were appearing even when it wasn't set to debug level

olliesilvester:
> Can we replace some of the code in `i24/serial/log.py` with calls to this common one to make it consistent?

Yes, I was thinking this could be done in separate PR (maybe https://github.com/DiamondLightSource/mx-bluesky/pull/572). Happy to do it in here instead though

src/mx_bluesky/common/utils/log.py
noemifrisina: Yes, mainly to have the choice of getting everything or just setting up the ones we're interested in (in i24 case we don't need the debug logs from ophyd, just ophyd_async).

noemifrisina:
We could do this here, or I can pull the changes from this branch into #572 and do that bit in there. I'm fine either way, if we're going to have that in this PR I'll just close the other one.

olliesilvester:
@noemifrisina I've added the flag you requested, if we merge this are you happy to get `i24/serial/log.py` to use this?

noemifrisina:
> @noemifrisina I've added the flag you requested, if we merge this are you happy to get `i24/serial/log.py` to use this?

Yes, thank you. I'll do the integration into i24 in the other branch

noemifrisina (approved):
Please rebase on main before merging

PR #585, opened 2024-10-18
520 calculate bragg angle

### LINKED ISSUE #520: Energy Changes: Calculate the bragg angle ourselves
From https://github.com/DiamondLightSource/mx-bluesky/issues/277 we found that the bragg angle on the DCM `BL03I-MO-DCM-01:BRAGG` does not update quickly after we set `BL03I-MO-DCM-01:ENERGY`. We need the bragg angle for lookup tables on movements that we should be able to do at the same time as the energy change and so we will need to work out what the bragg angle is ourselves.

This can be calculated by substituting the d-spacing of Si into bragg's law. 

## Acceptance Criteria
- We calculate the bragg angle of the DCM early when changing energy
- We move the other motors we have to move at the same time as the DCM energy motor

rtuck99:
Fixes #520 

Calculates the Bragg angle directly from the target energy instead of reading from the EPICS PV which is liable to return transient values.

### DIFF
--- a/src/mx_bluesky/hyperion/utils/utils.py
+++ b/src/mx_bluesky/hyperion/utils/utils.py
@@ -4,9 +4,6 @@ from math import asin
 from scanspec.core import AxesPoints, Axis
 from scipy.constants import physical_constants
 
-# This value taken from BeamLineEnergy_DCM_Angstrom_To_Deg_converter.xml
-SI_111_SPACING_ANGSTROMS = 3.13475
-
 hc_in_eV_and_Angstrom: float = (
     physical_constants["speed of light in vacuum"][0]
     * physical_constants["Planck constant in eV/Hz"][0]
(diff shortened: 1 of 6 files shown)


src/mx_bluesky/hyperion/utils/utils.py
```diff
@@ -23,3 +26,17 @@ def convert_angstrom_to_eV(wavelength: float) -> float:
...
+    """Compute the bragg angle given the energy in kev.
+
+    Args:
+        energy_kev:  The energy in keV
+        d_a:         The lattice spacing in Angstroms
+    Returns:
+        The bragg angle in degrees
+    """
+    wavelength_a = convert_eV_to_angstrom(energy_kev * 1000)
+    d = d_a
```
DominicOram, line 41: Nit: This rename seems a bit redundant

DominicOram (approved):
Great, thank you!

PR #586, opened 2024-10-18
Pin scanspec to 0.7.2

rtuck99:
This pins scanspec to 0.7.2 due to bluesky/scanspec#149


rtuck99:
Closed because now fixed in scanspec 0.7.6

PR #587, opened 2024-10-21
Replace calls to deprecated inject() with direct factory calls

rtuck99:
This fixes the unit tests after the deprecation of `inject()`
see https://github.com/DiamondLightSource/dodal/issues/845
merged in PR
DiamondLightSource/dodal#854


DominicOram (commented):
I'm not sure if this is all we need to do. If we don't wait for connection then is there not a danger that if we immediately try and use the device now we will fail? On the other hand waiting for connection inside the default args I think means that we will be doing work on import?

dperl-dls (approved):
LGTM

We should revisit this later when we see what the issues are

dperl-dls:
@DominicOram I think the idea is that BlueApi has connected to all the devices before

PR #588, opened 2024-10-21
#579 Use the correct dockerfile for building the release image

### LINKED ISSUE #579: Release workflow results in wrong docker image getting pushed to ghcr.io
When a release of hyperion is performed, the following things happen:
`_container.yml` runs and pushes a docker image that is built using `Dockerfile` instead of `Dockerfile.release`

`publish_docker_image.yml` runs with the correct Dockerfile but fails because it doesn't have permissions to push to ghcr.io

rtuck99:
Fixes #579

Now uses the main CI `_container.yml` workflow to publish the release image


olliesilvester (changes requested):
I might have missed something - now that publish_docker_image.yml has been removed, which step actually pushes `Dockerfile.release` to ghcr?

There's also a GitHub Actions warning about checking file annotations, but I don't know if that's relevant

rtuck99:
> I might have missed something - now that publish_docker_image.yml has been removed, which step actually pushes `Dockerfile.release` to ghcr?
> 
> There's also a GitHub Actions warning about checking file annotations, but I don't know if that's relevant

It's using the standard `python-copier` template workflow, which is pretty convoluted:

-  The main workflow that gets invoked on various actions is `ci.yml`, which calls the other .yml workflows in turn
-  The `_release.yml` file is a red herring - that just does the Github release page
-  The actual job that does it calls `_container.yml`, which gets called for pretty much everything. However the last step which pushes to ghcr.io only gets invoked if the build is tagged `github.ref_type == 'tag'` - this is what stops every PR publishing a docker image

DominicOram (approved):
Great, thank you

PR #591, opened 2024-10-24
Update plans to call from dodal.plan_stubs

DiamondJoseph:
Bluesky distinguishes between [plans:](https://github.com/bluesky/bluesky/blob/main/src/bluesky/plans.py) complete experimental proceedures, which open and close data collection runs, which may be part of a larger plan that collect data multiple times, but that might also be run alone to collect data, and [plan_stubs:](https://github.com/bluesky/bluesky/blob/main/src/bluesky/plan_stubs.py) which do not create & complete data collection runs and are either isolated behaviours or building blocks for plans.

In order to make it clearer when a MsgGenerator can be safely used without considering the enclosing run (as opening a run whilst in a run without explicitly passing a RunID is likely to cause both runs to fail), when it is required to manage a run and when running a procedure will create data documents, we should adopt this standard.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd == 1.9.0",
     "ophyd-async >= 0.3a5",
     "bluesky >= 1.13.0a4",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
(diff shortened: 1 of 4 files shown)


DiamondJoseph:
This is a partner change for
https://github.com/DiamondLightSource/dodal/pull/793
Once that is merged I will remove the dodal pinning.

DominicOram (approved):
Great, thanks. Taking the pin off before merge is good for keeping things tidy but not 100% required if you forget

PR #602, opened 2024-10-29
Revert "Replace calls to deprecated inject() with direct factory calls"

noemifrisina:
Reverts DiamondLightSource/mx-bluesky#587


DominicOram (approved):
Ta

Issue #540, opened 2024-09-30
 Clear constants out of the parameter models


DominicOram:
A brief description of the issue, including specific stakeholders and the business case where appropriate

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

DominicOram:
@olliesilvester to update this issue

olliesilvester:
This is in progress - constants were moved around a bit in https://github.com/DiamondLightSource/mx-bluesky/pull/621/files#diff-dfa58ccbe746f8a3acce8c15df766466e71fd350abb79eea0647f43a5ae15be9 and they will be again in https://github.com/DiamondLightSource/mx-bluesky/pull/652. Once the latter is merged, I'll give a new list of the remaining constants to move out

DominicOram:
@olliesilvester can you update this now?

PR #603, opened 2024-10-29
Begin moving params and constants to common

olliesilvester:
Constants:
- Structured in frozen dataclasses as before, now in `common`. Specific beamlines/Hyperion can inherit any of these dataclasses and override default values. For now, default values are those used by Hyperion
- Need to chase up with analysis whether it's okay to use the same zocalo environment for all beamlines, and if we can change it to a better name (currently `artemis`)
- `zocalo_environment` and `oav_centring_file` are now constants rather than parameters

Parameters:
- Not much has changed, but moved most of this to common. The defaults, for now, are the values used by Hyperion.
- Specific beamlines/Hyperion can modify a common parameter model by inheriting it and overriding relevant values

Params/constants will likely be moved around lots as we get a better idea as to which things are beamline-specific, but after this PR, things should be easier to adjust


src/mx_bluesky/common/parameters/components.py
```diff
@@ -0,0 +1,222 @@
...
+    def take_snapshots(self) -> bool:
+        return bool(self.snapshot_omegas_deg)
+
+
+class WithOptionalEnergyChange(BaseModel):
+    demand_energy_ev: float | None = Field(default=None, gt=0)
+
+
+# TODO do this properly. These probably shouldn't default to anything anymore
+class WithVisit(BaseModel):
```
olliesilvester, line 123: It probably is better to not have defaults for a lot of these parameters, but I think it might be easier to remove the relevant defaults once the parameters are needed by another beamline?

DominicOram: I think https://github.com/DiamondLightSource/mx-bluesky/issues/540 will remove some of these anyway

src/mx_bluesky/common/parameters/components.py
```diff
@@ -0,0 +1,221 @@
...
+    TriggerMode,
+)
+from pydantic import BaseModel, ConfigDict, Field, field_validator, model_validator
+from pydantic_extra_types.semantic_version import SemanticVersion
+from scanspec.core import AxesPoints
+from semver import Version
+
+from mx_bluesky.common.parameters.constants import TEST_MODE, DetectorParamConstants
+
+PARAMETER_VERSION = Version.parse("5.1.0")
```
DominicOram, line 20: Should: Can you bump this? I think removing the `zocalo_environment` parameter is a version change

src/mx_bluesky/hyperion/external_interaction/config_server.py
```diff
@@ -45,3 +47,8 @@ def update_self_from_server(self):
                 else self.overriden_features[flag]
             )
             setattr(self, flag, updated_value)
+
+    @classmethod
+    def from_json(cls, input: str | None):
+        assert input is not None
+        return cls(**json.loads(input))
```
DominicOram, line 54: Could: Why can't we use `pydantic`s own JSON serialisation i.e. https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_validate_json?

DominicOram (approved):
Thanks. This looks broadly correct, couple of minor things. I think we probably do need a more holistic look at the param model and how we use it with constants but as an initial move this is fine.

olliesilvester:
> Thanks. This looks broadly correct, couple of minor things. I think we probably do need a more holistic look at the param model and how we use it with constants but as an initial move this is fine.

Yeah, I agree. It might be worth going over every parameter together at the next technical meeting and deciding its fate

DominicOram:
> Yeah, I agree. It might be worth going over every parameter together at the next technical meeting and deciding its fate

Good idea

src/mx_bluesky/hyperion/external_interaction/config_server.py
olliesilvester: Ah yeah, thanks, this isn't needed actually

PR #606, opened 2024-10-31
Implementation of simple lightweight event base class

rtuck99:
I was going to use this as part of implementing `blSampleStatus` updates however in the end have plumped for a slightly different solution, however I think this may be of use for other things.

Therefore I present this as a draft PR for general comment/feedback and if people want to use it.


DominicOram (commented):
I think this really needs discussion with the wider `bluesky` community about why this isn't a core feature of `bluesky` already. I'm concerned that by doing this we're diverging more from standard `bluesky` things.

PR #607, opened 2024-11-01
Fix MultiRotationScan so that pydantic model_dump_json() works 

rtuck99:
This fixes an issue found in system tests where pydantic was reporting failure to serialize `semver.SemanticVersion`.


rtuck99:
> Thanks! This fixes the issue and I think it's good but I feel like this is pointing to an underlying issue in https://docs.pydantic.dev/latest/api/pydantic_extra_types_semantic_version/ or we're doing something that `pydantic` doesn't expect us to be doing by manipulating the parameter model like this. I don't think it's urgent for us to go down that rabbithole though
Possibly we could enforce more rigour with https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra
I think in the default setting, it will allow additional fields but clearly whatever additional manipulation is being done by SemanticVersion only works if it's a properly annotated part of the model.

PR #623, opened 2024-11-07
Fix unit conversion in grid scan step size

### LINKED ISSUE #220: Grid parameters erroneously contain values in mm when variables are in um
`get_grid_detection_callback.py` contains the following code:
```
    def get_grid_parameters(self) -> GridParamUpdate:
        return {
            "x_start_um": self.start_positions[0][0],
            "y_start_um": self.start_positions[0][1],
            "y2_start_um": self.start_positions[0][1],
            "z_start_um": self.start_positions[1][2],
            "z2_start_um": self.start_positions[1][2],
            "x_steps": self.box_numbers[0][0],
            "y_steps": self.box_numbers[0][1],
            "z_steps": self.box_numbers[1][1],
            "x_step_size_um": self.x_step_size_mm,
            "y_step_size_um": self.y_step_size_mm,
            "z_step_size_um": self.z_step_size_mm,
        }
```

these mm values are then propagated elsewhere to other parameter values which are also named as being in microns.
`start_positions` also contains values in mm, because the values derive from `calculate_x_y_z_of_pixel()` which uses motor positions which are in mm

DominicOram:
Fixes #220

See dodal PR at https://github.com/DiamondLightSource/dodal/pull/892

## To test
* Ensure tests still pass
* Ensure units are now all sensible


src/mx_bluesky/hyperion/external_interaction/callbacks/grid_detection_callback.py
```diff
@@ -87,7 +87,7 @@ def get_grid_parameters(self) -> GridParamUpdate:
             "x_steps": self.box_numbers[0][0],
```
rtuck99, line 87: `x_start_um` etc are also all currently in mm but labelled as microns

rtuck99: If you change `x_start_um` to really be in microns, you will also need to update the calculations in flyscan_xray_centre as these all assume the positions to be in mm.
It might be easier to just rename everything from `um` to `mm` as that's what everything was really in and revert your current conversion rather than update all the calculations.

I leave the choice to you, as long as everything is consistent it doesn't matter but at the moment, this fix doesn't really improve matters without the above changes.

rtuck99 (changes requested):
Need to fix the start positions as well in order for everything to be consistent.

src/mx_bluesky/hyperion/external_interaction/callbacks/grid_detection_callback.py
DominicOram: Good spot, thanks! Hmm... It's tricky `um` is the natural units for scientists and `mm` the ones used in EPICS. So we will have to do the conversion somewhere. Given the parameters are potentially scientist facing I have kept them in `um` for now

dperl-dls (approved):
Great, thanks!

rtuck99 (approved):
Approved

PR #624, opened 2024-11-07
Fix pytest aborting after first encountered test failure

rtuck99:
This PR removes a redundant copy of some code which is a workaround for

https://github.com/pytest-dev/pytest/issues/7409

importantly it now enforces the PYTEST_RAISE environment variable must be set to 1 in order to enable the workaround.

The workaround when enabled causes pytest to bail early at the first error, so before this PR, any failing test would cause all subsequent tests to not be execute, which is undesirable.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@ import asyncio
 import gzip
 import json
 import logging
+import os
 import sys
 import threading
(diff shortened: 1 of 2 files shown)


dperl-dls (approved):
lgtm

PR #626, opened 2024-11-07
Fix slow `fixed_target` test and add some documentation about profiling

rtuck99:
This shaves off about 10s from the unit test execution time


dperl-dls (approved):
This is brilliant, thanks!

PR #628, opened 2024-11-10
Remove `from_json` in parameter model

### LINKED ISSUE #457: Remove `from_json`
Now that we're using pydantic 2 and not dynamically changing the extra fields at runtime (https://github.com/DiamondLightSource/mx-bluesky/pull/456) we should not need to have our custom `from_json` method and just be able to use `model_validate_json`. However, when I tried to do this I found a number of validation errors in the tests. 

## Acceptance Criteria
- We use the pydantic functions for converting json to models and remove `from_json`
- Tests all pass still

DominicOram:
Fixes #457

Contains changes from https://github.com/DiamondLightSource/mx-bluesky/pull/629, make sure that is merged first

To Test:
* Confirm tests still pass


pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd == 1.9.0",
     "ophyd-async >= 0.3a5",
     "bluesky >= 1.13.0a4",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
+    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@f8157dc325896748a743493482222772986c5969",
```
rtuck99, line 49: Can remove this

rtuck99 (approved):
Remove dodal pin, otherwise approved

PR #629, opened 2024-11-10
Update for latest ophyd-async version

DominicOram:
See https://github.com/DiamondLightSource/dodal/pull/880


rtuck99 (approved):
Approved

PR #630, opened 2024-11-10
Remove unused simple_beamline fixture

### LINKED ISSUE #610: Remove `simple_beamline`
The `simple_beamline` fixture is only used in one tests and there are other fixtures that we can use instead. We should reuse the more widely used ones.

## Acceptance Criteria
- `simple_beamline` is removed in favor of more widely used fixtures

DominicOram:
Fixes #610 

To test:
* Confirm tests still pass


rtuck99 (approved):
Approved

PR #637, opened 2024-11-11
Update to new exp-eye endpoints

### LINKED ISSUE #636: Rename exp-eye end points
The exp-eye end points are being moved so that `core` and `sample-handling` are under the same root path e.g. `/core/proposals/{proposalId}/sessions` would become `/proposals/{proposalId}/sessions`

There will be a brief period where both are available but we should move to the new endpoint ASAP.

## Acceptance Criteria
- We move to the new `exp-eye` endpoints

DominicOram:
Fixes #636 

To test:
* Confirm unit tests pass
* Confirm system tests pass (`pytest tests/system_tests/hyperion/external_interaction/test_exp_eye_dev.py`)


rtuck99 (approved):
Content looks ok. Exp-eye staging seems to be down so can't verify that the endpoints work in systest. I've pushed the equivalent change for the BLSampleStatus PR

DominicOram:
I have reconfirmed that the system tests pass now that staging is back up

Issue #578, opened 2024-10-16
 The mx-bluesky documentation is confusing, misleading and repetitive


rtuck99:
Some additional comments about the documentation - I think it's still confusing, partly the fault of python copier template. Perhaps in future we should just ignore most of their documentation templates because I think they are not really designed for our use case.

So, if you go to the documentation front page, to look for how to deploy mx-bluesky/hyperion you see this:

![image](https://github.com/user-attachments/assets/f1cfbaf2-9639-4f39-9da6-10a450e0efdb)

ok, so go to the "User Guide"
![image](https://github.com/user-attachments/assets/54858912-30a7-4ecd-b16e-7e779351354e)

under installation, this doesn't mention `dls_dev_env.sh`.

the real documentation is in mx-bluesky/build/html/developer/general/how-to/get-started.html

I think we should delete the installation documentation in the User Guide as generally users won't be installing mx-bluesky themselves either as a development install or a production deployment.

Similarly "Run in a container"

In fact, maybe we should just get rid of the User Guide section entirely as there is nothing in there at the moment for actual users.

Ok, so how about the Developer Guide:
![image](https://github.com/user-attachments/assets/36c6bca7-f70b-4e9e-ba85-12df99127b54)

Maybe we should change the title of General MX-Bluesky Developer Guide so we don't have two of them...

So lets try General MX-Bluesky Developer Guide (the first one):

![image](https://github.com/user-attachments/assets/9cd04f86-7e7c-4790-8e37-8524c2dafe61)

If I'm a developer, the install documentation must be in Developer Install? Wrong!

That's just the generic install documentation

Try Get started with mx-bluesky - at least this tells you to use `dls_dev_setup.sh`. But wait - it's really called  `dls_dev_env.sh`. 

...to be continued

olliesilvester:
@rtuck99 I agree that the docs are gross, are you happy for me to pick this up?

olliesilvester:
I agree with most of your points, but I think the hyperion developer guide should be a subcategory of the general mx-bluesky developer guide, rather than renaming one of them

olliesilvester:
Going to split this up into the following:

- Remove user guide, and in the main page explain that we have no users right now so we only need a developer guide
- Put Hyperion developer guide as a subcategory of genera mx-bluesky developer guide
[...]

PR #641, opened 2024-11-12
Improve docs: Remove user guide and change text on main page

olliesilvester:
Part of the documentation cleanup for https://github.com/DiamondLightSource/mx-bluesky/issues/578

To test:
- From the mx-bluesky virtual environment, run `sphinx-autobuild path_to_docs output-dir` 
- Once built, open new docs on `http://127.0.0.1:8000`
- Compare with https://diamondlightsource.github.io/mx-bluesky/main/index.html and approve if this PR is an improvement


DominicOram (approved):
Thanks, this is less confusing than before. I think it's now a bit odd we have a whole section there that is just one button but we can address in future PRs

PR #642, opened 2024-11-12
Fix system tests to work with new OAV snapshots

rtuck99:
This fixes the `test_ispyb_dev_connection` and `test_load_centre_collect_full_plan` system tests to work against the new OAV changes


DominicOram (approved):
Looks good, thanks. It was failing so added a bit more mocking

PR #643, opened 2024-11-12
Reduce filesystem access when validating parameters

### LINKED ISSUE #527: Parameters implicitly creating storage_directory on access
The `detector_params` property accessors in `GridCommon` and `RotationScan` create the `storage_directory` implicitly.

This violates the principle of least surprise. Also it means a filesystem access every time the property is accessed to check if the directory exists and create it if necessary.
The existence of the directory is asserted by `DetectorParams.directory` field validator (which is reasonable).
The directories should be created by the experiment plans that create the files. However many unit tests will also need to be updated as well.

See also #524 which was raised due to the code that cleans up after the unit tests having issues.

rtuck99:
This moves the check/creation of the `storage_directory` folder from property access to `DiffractionExperiment` model validation.

This doesn't completely remove duplicated calls to `os.makedirs` but it does reduce it from ~22 to a hopefully more sane number.
But there will still be some duplications as we construct various intermediate parameter instances, all of which have the `storage_directory` parameter.

Fixes #527

### DIFF
--- a/src/mx_bluesky/common/parameters/components.py
+++ b/src/mx_bluesky/common/parameters/components.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import os
 from abc import abstractmethod
 from collections.abc import Sequence
 from enum import StrEnum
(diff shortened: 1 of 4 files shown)


DominicOram (approved):
Looks good, thank you!

PR #627, opened 2024-11-07
I24 serial: Read values from ophyd devices in dcid

noemifrisina:
Closes #575 

Needs [dodal#894](https://github.com/DiamondLightSource/dodal/pull/894)

Read the values to pass to dcid from ophyd devices.

PR #645, opened 2024-11-13
I24 serial: commissioning fixes

noemifrisina:
Fixes from commissioning on 13/11/24

* Fix fixed-target run script
* Do no reset visit PV when initialising stages as staff should set that
* Update beam center values (see #647 )
* Add a huge buffer to expected collection time for PMAC program runner as it seems to build up the more blocks get scanned (see [Dodal#900](https://github.com/DiamondLightSource/dodal/issues/900))
* Increase the sleep between triggering the PMAC_STING in `cs_maker` in order to allow for the motors to finish moving (see #646 )
* Update missed loggers to new configuration
* Allow more than one collection to be written in a directory for eiger


src/mx_bluesky/beamlines/i24/serial/dcid.py
```diff
@@ -224,20 +222,20 @@ def generate_dcid(
             )
             resp.raise_for_status()
             self.dcid = resp.json()["dataCollectionId"]
-            logger.info("Generated DCID %s", self.dcid)
+            SSX_LOGGER.info("Generated DCID %s", self.dcid)
         except requests.HTTPError as e:
```
rtuck99, line 226: I think we should have some tests for this file

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -467,7 +471,7 @@ def start_i24(
         SSX_LOGGER.info("Using Eiger detector")
 
         SSX_LOGGER.debug(f"Creating the directory for the collection in {filepath}.")
-        Path(filepath).mkdir(parents=True)
+        Path(filepath).mkdir(parents=True, exist_ok=True)
```
rtuck99, line 475: why do we only create the directory for the eiger and not the pilatus?

src/mx_bluesky/beamlines/i24/serial/setup_beamline/setup_beamline.py
```diff
@@ -319,24 +317,24 @@ def pilatus(action, args_list):
...
+            SSX_LOGGER.debug(f"Argument: {arg}")
     # caput(pv.eiger_wavelength, caget(pv.dcm_lambda))
     caput(pv.eiger_detdist, str(float(caget(pv.det_z)) / 1000))
-    logger.warning("WARNING: Have you set header info?")
+    SSX_LOGGER.warning("WARNING: Have you set header info?")
     caput(pv.eiger_wavelength, caget(pv.dcm_lambda))
     caput(pv.eiger_omegaincr, 0.0)
-    caput(pv.eiger_beamx, 1605.7)
-    caput(pv.eiger_beamy, 1702.7)
+    caput(pv.eiger_beamx, 1600.0)
```
rtuck99, line 336: I think these should be defined as global constants at the top of the file rather than buried in the code here.

noemifrisina: This is being done in another PR, see https://github.com/DiamondLightSource/mx-bluesky/blob/3046a1d1df56f674b0fc9b87672f0a9653baf338/src/mx_bluesky/beamlines/i24/serial/setup_beamline/setup_beamline.py#L15

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Because the Pilatus creates its own directories, while the Eiger doesn't, so there's no need to explicitly create it. Until a few weeks ago - when we started running as i24detector on procserv - this one line was actually triggering a one image collection from the Pilatus just to get the directory....

src/mx_bluesky/beamlines/i24/serial/dcid.py
noemifrisina: Adding them in #627 , which is actually changing the code in this file so while I may add them here they would go away in a couple of hours?

rtuck99 (approved):
Approved

Issue #214, opened 2024-08-05
Move ispyb and nexus functions to mx_bluesky


olliesilvester:
For i03 and vmxm will use the same, if not very similar functions for ispyb and nexus interactions. Zocalo can stay in Hyperion for now, but will need to be moved too at some point.

## Acceptance Criteria:

- Everything in `hyperion.external_interaction.ispyb` and `hyperion.external_interaction.nexus` moved to `mx_bluesky`
- Hyperion now calls this from `mx_bluesky`

olliesilvester:
As discussed with @noemifrisina , this ticket is just for moving things into 'common'. Actually generalising these functions to not be i03-specific is a separate job

olliesilvester:
Prerequisite for this is moving Hyperion's parameter model to common, which I am halfway through

Issue #215, opened 2024-08-05
Move xray centre callbacks to mx_bluesky


olliesilvester:
VMXm and Hyperion will both use the same `ispyb` and `nexus` callbacks for their Xray centring. We should move this to `mx_bluesky`

## Acceptance Criteria

- Everything in `hyperion.external_interaction.callbacks.xray_centre` moved to `mx_bluesky`
- Hyperion now calls this part from `mx_bluesky`

DominicOram:
These have now been moved

PR #652, opened 2024-11-15
Move XRC callbacks to common

olliesilvester:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/214 and https://github.com/DiamondLightSource/mx-bluesky/issues/215

Moves lots of the `external_interaction` code into common. Some main bits:
- A common `ThreeDGridScan` which `HyperionThreeDGridScan` now inherits
- Functions for talking to Nexus and ISPyB moved to `common`
- All XRC and common callbacks moved to `mx_bluesky` common
- XRC nexus callbacks now needs to know what type of parameters are being used. It accepts `ThreeDGridScan` any any children of this type

Probably difficult to review, sorry


olliesilvester:
See https://github.com/DiamondLightSource/mx-bluesky/issues/655 for sorting out the tests. Annoyingly, we might want to do this before merging, so we can be sure that this PR didn't break anything, but I don't want to hold up @noemifrisina 's work

olliesilvester:
> > See #655 for sorting out the tests. Annoyingly, we might want to do this before merging, so we can be sure that this PR didn't break anything, but I don't want to hold up @noemifrisina 's work
> 
> If all the important bits have been moved and sorting out the tests is all that's left to this one, I think it would be worth it to sort out the tests before merging this. While that might take awhile, as you said at least we'd be reasonably sure things are still working after this PR. For my work, I can just make a new branch off this one in the meantime and make a start

Cool, as long as it's not holding you up

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -1,22 +1,24 @@
...
+from mx_bluesky.common.parameters.gridscan import ThreeDGridScan
+from mx_bluesky.common.utils.log import NEXUS_LOGGER
 
 if TYPE_CHECKING:
     from event_model.documents import Event, EventDescriptor, RunStart
 
+T = TypeVar("T", bound="ThreeDGridScan")
+
 
 class GridscanNexusFileCallback(PlanReactiveCallback):
```
noemifrisina, line 23: Is there a particular reason why we're moving this particular nexus callback and not the rotation one too? Would it be worth making a nore general one from which both this and the rotation one in hyperion - and any other ones we need in the future - can inherit? 
This would be a separate PR in any case, only trying to understand the what's going on...

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -21,6 +28,7 @@
     HardwareConstants,
 )
 from mx_bluesky.common.parameters.robot_load import RobotLoadAndEnergyChange
+from mx_bluesky.hyperion.parameters.constants import CONST
```
noemifrisina, line 31: Nit: maybe import CONST as hyperion_const or something? 

On a first read this threw me a bit because I was assuming `CONST` would be common and not hyperion specific...

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
olliesilvester: Yeah, there's no reason other than trying to not do everything at once. For now, I only need the bits in XRC, so I've only moved the parts needed for this

noemifrisina: Thanks. Sounds good to me, maybe add an issue if it doesn't exist already - or a comment about this on an existing one?

noemifrisina:
On a first go through, looks okay, thank you. Will wait to finish review and approve till  you're done with the tests. I'd say the fact that they pass right now is a good start though :)

src/mx_bluesky/common/parameters/gridscan.py
olliesilvester: Ah, good catch, I shouldn't be importing anything from `Hyperion` to `mx_bluesky/common`. We're meant to have linting rules which catches this, I'll check why it didn't work

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
olliesilvester: Yeah, as long as no one is waiting for this PR, I can include that here

src/mx_bluesky/hyperion/parameters/gridscan.py
```diff
@@ -9,27 +9,32 @@
...
+    ThreeDGridScan,
 )
 from mx_bluesky.hyperion.parameters.components import WithHyperionFeatures
 from mx_bluesky.hyperion.parameters.constants import CONST, I03Constants
 
 
-class HyperionGridCommon(GridCommon, WithHyperionFeatures):
-    # This class only exists so that we can properly select enable_dev_shm. Remove in
-    # https://github.com/DiamondLightSource/hyperion/issues/1395"""
+class HyperionThreeDGridScan(
```
DominicOram, line 26: Should: `ThreeDGridScan` already inherits from:
```python
    GridCommon,
    SpecifiedGrid,
    SplitScan,
    WithOptionalEnergyChange,
```
So we can drop the others here

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -35,22 +37,23 @@ class GridscanNexusFileCallback(PlanReactiveCallback):
...
-            assert isinstance(hyperion_params, str)
+        if doc.get("subplan_name") == PlanNameConstants.GRIDSCAN_OUTER:
+            mx_bluesky_parameters = doc.get("mx_bluesky_parameters")
+            assert isinstance(mx_bluesky_parameters, str)
             NEXUS_LOGGER.info(
-                f"Nexus writer received start document with experiment parameters {hyperion_params}"
+                f"Nexus writer received start document with experiment parameters {mx_bluesky_parameters}"
             )
-            parameters = HyperionThreeDGridScan.model_validate_json(hyperion_params)
+            parameters = self.param_type.model_validate_json(mx_bluesky_parameters)
```
DominicOram, line 56: Could: Can we just always use do `ThreeDGridScan.model_validate_json` here? We don't need all the additional stuff from Hyperion anyway? Maybe we can get away with an even broader base class e.g. in `ispyb_callback` we just convert to a `GridCommon`. Basically the more broader the baseclass the more reusable this callback should be

DominicOram (commented):
Couple of comments between meetings, not a full review

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
DominicOram: I think that's going to be quite a big refactor, I would do it in a new PR

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
olliesilvester: Yes, good idea

olliesilvester:
Regarding tests:
- Most of the external interaction unit tests have now been moved. The parts that remain are testing rotation scans - this PR doesn't involve that part, although it will be done at some point
- I haven't moved system tests as most use s03 or are hyperion specific. I think we need to sort out which system tests are worth keeping up to date and which ones are too high maintenance. There are some issues about this

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
```diff
@@ -35,22 +37,23 @@ class GridscanNexusFileCallback(PlanReactiveCallback):
     See: https://blueskyproject.io/bluesky/callbacks.html#ways-to-invoke-callbacks
     """
 
-    def __init__(self) -> None:
+    def __init__(self, param_type: type[T]) -> None:
         super().__init__(NEXUS_LOGGER)
+        self.param_type = param_type
```
rtuck99, line 42: What is this `param_type` actually used for? Do we need it, or the associated typevar?

tests/unit_tests/common/external_interaction/xray_centre/conftest.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.fixture
+def mock_ispyb_update_time_and_status():
+    with patch(
+        "mx_bluesky.common.external_interaction.callbacks.xray_centre.ispyb_callback.StoreInIspyb._update_scan_with_end_time_and_status"
+    ) as p:
+        yield p
+
+
+@pytest.fixture
```
rtuck99, line 42: Where are all of these fixtures used?

rtuck99 (approved):
Approved - only the ~one~ two minor comments

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/nexus_callback.py
olliesilvester: Whoops, forgot to delete this

tests/unit_tests/common/external_interaction/xray_centre/conftest.py
olliesilvester: Nowhere! :face_with_open_eyes_and_hand_over_mouth:

PR #656, opened 2024-11-15
Explicitly depend on dls-bluesky-core

DominicOram:
Our unit tests are failing as we still have a dependency on `dls-bluesky-core` (https://github.com/DiamondLightSource/mx-bluesky/actions/runs/11860322306/job/33055330489?pr=630). We used to get this dependency through implicitly `blueapi` but the latest `blueapi` version no longer has it so we must add it explicitly. 

https://github.com/DiamondLightSource/mx-bluesky/pull/564 will remove this properly


olliesilvester (approved):
I think it's been about a year since the test I changed actually passed!

PR #665, opened 2024-11-19
Fix flaky backlight test

### LINKED ISSUE #653: Fix unstable test for grid_detect_then_xray_centre_plan
Our unit test `test_detect_grid_and_do_gridscan` occasionally fails. This is likely due to something not being cleaned up properly. See https://github.com/DiamondLightSource/mx-bluesky/actions/runs/11859361401/job/33052132442 for an instance of it failing

## Acceptance Criteria
- Improved test

DominicOram:
Fixes #653

Rather than check that the backlight has finished moving now we just check that we told it to move, which should be a lot less flaky.

To test:
* Run the unit tests all through a few times and confirm they do not fail with the backlight timeout error

### DIFF
--- a/tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
@@ -10,7 +10,7 @@ from bluesky.utils import Msg
 from dodal.devices.aperturescatterguard import ApertureValue
 from dodal.devices.backlight import BacklightPosition
 from dodal.devices.oav.oav_parameters import OAVParameters
-from ophyd_async.core import set_mock_value
+from ophyd_async.core import get_mock_put, set_mock_value
 
 from mx_bluesky.common.parameters.gridscan import GridScanWithEdgeDetect
 from mx_bluesky.hyperion.experiment_plans.grid_detect_then_xray_centre_plan import (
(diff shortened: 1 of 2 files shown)


tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -114,7 +114,9 @@ async def test_detect_grid_and_do_gridscan(
     mock_grid_detection_plan.assert_called_once()
 
     # Check backlight was moved OUT
-    assert await composite.backlight.position.get_value() == BacklightPosition.OUT
+    get_mock_put(composite.backlight.position).assert_called_once_with(
```
DiamondJoseph, line 117: Does your plan not wait on the move for the backlight? Or is the Sim backend not respecting where it does?

DominicOram: The plan goes on to call another plan that waits on the backlight. However, in this particular test we have mocked out that other plan. I guess I could have fixed this by not mocking out the other plan or having our mock do the wait but this seemed the simplest solution

dperl-dls (approved):
LGTM

tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
rtuck99: Hmm, this seems to be just as unreliable as the previous code

PR #668, opened 2024-11-21
Fix for pydantic 2.10

dperl-dls:
- Changes our device composites to pydantic dataclasses with config which allows the devices to be used in the plan parameter model generated by BlueApi
- Updates where tests used plain MagicMocks which are now no longer accepted by the valiation. In most cases `spec=RealDeviceType` is sufficient, but when children are needed an ophyd-async `connect(mock=True)` is needed too as spec doesn't find those


noemifrisina (approved):
Thanks

PR #670, opened 2024-11-21
Add redis to murko forwarder

### LINKED ISSUE #570: Murko: Move socket_handler to mx_bluesky
The `socket_handler` in https://github.com/DiamondLightSource/mx_auto_mjpeg_capture/tree/main/socket_handler should live in `mx_bluesky` so that all the code for murko interactions is in one place.

The job of the `socket_handler` is to:
* Listen to when image data is published to redis
* Grab the data, decode to numpy arrays
* Batch the numpy arrays
* Send results to murko
* Put the results back on a redis key
* Publish to say there are new results

## Acceptance Criteria
- The socket handler code is moved into mx_bluesky
- There are tests on the above
- There is a system test where the OAVToRedis forwarder sends data to this via redis
- Docs are updated if necessary

DominicOram:
Fixes #570 

To test:
* Check it is understandable what the code is doing
* Confirm new unit tests work


src/mx_bluesky/scratch.py
rtuck99, line 1: It's empty!

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -0,0 +1,178 @@
+import io
+import json
+import pickle
+from typing import TypedDict
+
+import numpy as np
+import zmq
+from dodal.beamlines.i04 import MURKO_REDIS_DB, REDIS_HOST, REDIS_PASSWORD
```
rtuck99, line 8: I assume the password there isn't real password, and that on the beamline it's different and not checked in.
Even so, I think it's bad practice to have the password stored in a location that's inside the git workspace, as this makes it too easy to accidentally commit and push to the repo.
So we should find some other way of reading it e.g. from secure configuration file somewhere.

DominicOram: We're pulling it out of a k8s secret. See https://github.com/DiamondLightSource/dodal/pull/913/files. I think this is reasonably secure and the recommended way of doing it but happy to discuss

rtuck99 (approved):
Approved

PR #671, opened 2024-11-22
Fix failing backlight detection in unit test `test_detect_grid_and_do_gridscan()`

rtuck99:
Fix for #653 as previous fix #665 was not reliable


dperl-dls (approved):
sure, why not

PR #673, opened 2024-11-25
Remove initial_xyz and just read in the setup plan

### LINKED ISSUE #541: Remove `initial_xyz`
Following https://github.com/DiamondLightSource/mx-bluesky/pull/538 `initial_xyz` is no longer really used other than by the panda setup, which only really uses the `x`. Instead the panda setup should just read the x and otherwise we shouldn't be reading the initial at all.

## Acceptance Criteria
- The code reading `initial_xyz` in the xrc is removed

DominicOram:
Fixes #541 

To test:
* Confirm tests still pass


rtuck99 (commented):
I think we can now also remove `sample_motors` as there is nothing else that uses it.

olliesilvester:
@rtuck99 @shihab-dls @olliesilvester  we said we'd get this merged ourselves during sprint planning so this comment is to remind us to do that

rtuck99 (approved):
Re-approved

PR #675, opened 2024-11-25
Tidy call_nexgen in i24 SSX

### LINKED ISSUE #674: Tidy up `call_nexgen`
The `call_nexgen` function could do with some tidying up:
* It always returns the same thing, which is never used
* It takes both an `expt_type` and a parameters instance, this is redundant
* It checks if it's called with an eiger at the end, it should either not check (as it's only ever called with eiger) or check at the start

DominicOram:
Fixes #674

To test:
* Confirm tests pass and code is cleaner

### DIFF
--- a/src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
+++ b/src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
@@ -1,4 +1,5 @@
 import json
+from abc import abstractmethod
 from collections.abc import Sequence
 from pathlib import Path
 from typing import Literal
@@ -43,17 +44,28 @@ class LaserExperiment(BaseModel):
     pre_pump_exposure_s: float | None = None  # Pre illumination, just for chip
 
 
-class ExtruderParameters(SerialExperiment, LaserExperiment):
-    """Extruder parameter model."""
-
-    num_images: int
-    pump_status: bool
-
+class SerialAndLaserExperiment(SerialExperiment, LaserExperiment):
     @classmethod
     def from_file(cls, filename: str | Path):
         with open(filename) as fh:
             raw_params = json.load(fh)
         return cls(**raw_params)
+
+    @property
+    @abstractmethod
+    def experiment_type(self) -> str:
+        pass
+
+
+class ExtruderParameters(SerialAndLaserExperiment):
+    """Extruder parameter model."""
+
+    num_images: int
+    pump_status: bool
+
+    @property
+    def experiment_type(self) -> str:
+        return "extruder"
 
 
 class ChipDescription(BaseModel):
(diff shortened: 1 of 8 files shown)


src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
```diff
@@ -98,11 +110,9 @@ class FixedTargetParameters(SerialExperiment, LaserExperiment):
     checker_pattern: bool = False
     total_num_images: int = 0  # Calculated in the code for now
 
-    @classmethod
-    def from_file(cls, filename: str | Path):
-        with open(filename) as fh:
-            raw_params = json.load(fh)
-        return cls(**raw_params)
+    @property
+    def experiment_type(self) -> str:
```
noemifrisina, line 114: Nit: I think we may as well use SSXType here? Doesn't really break anything but it might be tidier in the long run to start removing all the "fixed-target"/"extruder" strings and use the type directly (I don't think there's that many left)

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
```diff
@@ -43,18 +44,29 @@ class LaserExperiment(BaseModel):
     pre_pump_exposure_s: float | None = None  # Pre illumination, just for chip
 
 
-class ExtruderParameters(SerialExperiment, LaserExperiment):
-    """Extruder parameter model."""
-
-    num_images: int
-    pump_status: bool
-
+class SerialAndLaserExperiment(SerialExperiment, LaserExperiment):
```
noemifrisina, line 47: Just a comment, I feel having `Laser` in the name is a bit misleading as it's not always a pump-probe experiment, but don't really mind if it stays like this

noemifrisina (changes requested):
Looks okay, thanks. Please rebase before merging, the linter is failing because of the new version of ruff but this is fixed on main already (and in the 575 branch)

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
DominicOram: I agree that it would be nicer to use an enum but I don't think I can use `SSXType` here as this string is what experiment type we want to send to `nexgen`. If we want to define it in an enum it should probably be defined in `nexgen` and imported here

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
DominicOram: I agree but the `FixedTargetParameters` and `ExtruderParameters` were already `LaserExperiment`s should they not have been?

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
noemifrisina: Ah, yes wasn't thinking about nexgen. You have a point. that fix should come first

noemifrisina (approved):

Issue #651, opened 2024-11-15
 Fix energy change issues uncovered from testing 2024-11-12


rtuck99:
During testing the following issues were found with the energy changes:

* Changing the energy causes the `dcm_offset` to be unstable and prevents it being set until the change is finished
* Bimorph mirror voltages time out when set simultaneously as the sets proceed incrementally instead of all at once, so actually need to set them one at a time
* lat/yaw need to be set on a stripe change, which should drive these changes as per GDA; the lookup-table for the lat (which is effectively a constant) then doesn't apply
* We should call the set energy again before collection/xrc so that the energy is correctly set even if a robot load hasn't been issued - if we are already at the right energy this should be fairly quick.

rtuck99:
Awaiting deployment + test

PR #676, opened 2024-11-26
651 energy change fixes

rtuck99:
This fixes energy change issues discovered during beamline testing #651 

See also DiamondLightSource/dodal#918

1. Energy changes should now result in the VFM yaw and lateral position (x-offset) being set depending on the mirror stripe.
Previously the yaw was unset and the lateral position adjusted according to a lookup-table that was effectively single-valued.
2. The energy change, and individual DCM changes are now serialized because they cannot be done in parallel
3. The timeouts for the yaw, lat operations are increased to 30s from 10s
4. Mirror voltage changes are now serialized instead of being attempted in parallel
5. Energy changes are now done in `robot_load_then_centre` regardless of whether a robot load and/or chi change is performed or not

1. Energy changes work on beamline
2. Unit tests pass

### Checks for reviewer
- [ ] Would the PR title make sense to a scientist on a set of release notes
- [ ] If a new device has been added does it follow the [standards](https://diamondlightsource.github.io/dodal/main/reference/device-standards.html)
- [ ] If changing the API for a pre-existing device, ensure that any beamlines using this device have updated their Bluesky plans accordingly
- [ ] Have the connection tests for the relevant beamline(s) been run via `dodal connect ${BEAMLINE}`

### DIFF
--- a/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
@@ -60,23 +60,29 @@ def adjust_mirror_stripe(
 ):
     """Feedback should be OFF prior to entry, in order to prevent
     feedback from making unnecessary corrections while beam is being adjusted."""
-    stripe, yaw, lat = mirror.energy_to_stripe(energy_kev)
+    mirror_config = mirror.energy_to_stripe(energy_kev)
 
     LOGGER.info(
-        f"Adjusting mirror stripe for {energy_kev}keV selecting {stripe} stripe"
+        f"Adjusting mirror stripe for {energy_kev}keV selecting {mirror_config['stripe']} stripe"
     )
-    yield from bps.abs_set(mirror.stripe, stripe, wait=True)
+    yield from bps.abs_set(mirror.stripe, mirror_config["stripe"], wait=True)
     yield from bps.trigger(mirror.apply_stripe)
 
     # yaw, lat cannot be done simultaneously
-    LOGGER.info(f"Adjusting {mirror.name} lat to {lat}")
-    yield from bps.abs_set(mirror.x_mm, lat, wait=True, timeout=YAW_LAT_TIMEOUT_S)
+    LOGGER.info(f"Adjusting {mirror.name} lat to {mirror_config['lat_mm']}")
+    yield from bps.abs_set(
+        mirror.x_mm, mirror_config["lat_mm"], wait=True, timeout=YAW_LAT_TIMEOUT_S
+    )
 
-    LOGGER.info(f"Adjusting {mirror.name} yaw to {yaw}")
-    yield from bps.abs_set(mirror.yaw_mrad, yaw, wait=True, timeout=YAW_LAT_TIMEOUT_S)
+    LOGGER.info(f"Adjusting {mirror.name} yaw to {mirror_config['yaw_mrad']}")
+    yield from bps.abs_set(
+        mirror.yaw_mrad, mirror_config["yaw_mrad"], wait=True, timeout=YAW_LAT_TIMEOUT_S
+    )
 
     LOGGER.info("Adjusting mirror voltages...")
-    yield from _apply_and_wait_for_voltages_to_settle(stripe, mirror_voltages)
+    yield from _apply_and_wait_for_voltages_to_settle(
+        mirror_config["stripe"], mirror_voltages
+    )
 
 
 def adjust_dcm_pitch_roll_vfm_from_lut(
--- a/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
@@ -121,13 +121,6 @@ def adjust_dcm_pitch_roll_vfm_from_lut(
     yield from dcm_roll_adjuster(DCM_GROUP)
     LOGGER.info("Waiting for DCM roll adjust to complete...")
 
-    # DCM offset
-    # wait for other DCM motions and energy change to complete before changing offset
-    yield from bps.wait(DCM_GROUP, timeout=30)
-    offset_mm = undulator_dcm.dcm_fixed_offset_mm
-    LOGGER.info(f"Adjusting DCM offset to {offset_mm} mm")
-    yield from bps.abs_set(dcm.offset_in_mm, offset_mm, wait=True)
-
     #
     # Adjust vfm mirror stripe and mirror voltages
     #
--- a/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
@@ -19,6 +19,7 @@ from mx_bluesky.hyperion.utils.utils import (
 
 MIRROR_VOLTAGE_GROUP = "MIRROR_VOLTAGE_GROUP"
 DCM_GROUP = "DCM_GROUP"
+YAW_LAT_TIMEOUT_S = 30
 
 
 def _apply_and_wait_for_voltages_to_settle(
@@ -46,14 +47,12 @@ def _apply_and_wait_for_voltages_to_settle(
         for voltage_channel, required_voltage in zip(
             channels.values(), required_voltages, strict=True
         ):
-            LOGGER.debug(
+            LOGGER.info(
                 f"Applying and waiting for voltage {voltage_channel.name} = {required_voltage}"
             )
             yield from bps.abs_set(
-                voltage_channel, required_voltage, group=MIRROR_VOLTAGE_GROUP
+                voltage_channel, required_voltage, group=MIRROR_VOLTAGE_GROUP, wait=True
             )
-
-    yield from bps.wait(group=MIRROR_VOLTAGE_GROUP)
 
 
 def adjust_mirror_stripe(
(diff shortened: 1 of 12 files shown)


src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -46,31 +47,42 @@ def _apply_and_wait_for_voltages_to_settle(
...
+    # yaw, lat cannot be done simultaneously
+    LOGGER.info(f"Adjusting {mirror.name} lat to {mirror_config['lat_mm']}")
+    yield from bps.abs_set(
+        mirror.x_mm, mirror_config["lat_mm"], wait=True, timeout=YAW_LAT_TIMEOUT_S
+    )
+
+    LOGGER.info(f"Adjusting {mirror.name} yaw to {mirror_config['yaw_mrad']}")
+    yield from bps.abs_set(
+        mirror.yaw_mrad, mirror_config["yaw_mrad"], wait=True, timeout=YAW_LAT_TIMEOUT_S
+    )
```
DominicOram, line 80: Could: Rather than ask the device for it's settings based on the energy and then apply them in the plan could we just put it all into the device and allow us to do `mirror.set(energy_kev)` with the device handling the rest? This seems like it could be cleaner?

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -109,31 +121,16 @@ def adjust_dcm_pitch_roll_vfm_from_lut(
     yield from dcm_roll_adjuster(DCM_GROUP)
     LOGGER.info("Waiting for DCM roll adjust to complete...")
 
-    # DCM Perp pitch
+    # DCM offset
+    # wait for other DCM motions and energy change to complete before changing offset
+    yield from bps.wait(DCM_GROUP, timeout=30)
```
DominicOram, line 126: Could: Similar to above, given the `offest_mm` is now being done in the `undulator_dcm` should we do all these moves in the device side to avoid the logic being split over so many places?

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -46,31 +47,42 @@ def _apply_and_wait_for_voltages_to_settle(
...
 
+    # yaw, lat cannot be done simultaneously
+    LOGGER.info(f"Adjusting {mirror.name} lat to {mirror_config['lat_mm']}")
+    yield from bps.abs_set(
+        mirror.x_mm, mirror_config["lat_mm"], wait=True, timeout=YAW_LAT_TIMEOUT_S
+    )
+
+    LOGGER.info(f"Adjusting {mirror.name} yaw to {mirror_config['yaw_mrad']}")
+    yield from bps.abs_set(
+        mirror.yaw_mrad, mirror_config["yaw_mrad"], wait=True, timeout=YAW_LAT_TIMEOUT_S
```
DominicOram, line 79: Should: It's misleading that the `LAT_TIMEOUT` is also being used on `yaw`

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -178,12 +182,23 @@ def robot_load_then_xray_centre(
             parameters,
         )
         LOGGER.info("Pin not loaded, loading and centring")
```
DominicOram, line 184: Nit: Unrelated to this PR, I know but it's odd this message is logged after the fact

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -178,12 +182,23 @@ def robot_load_then_xray_centre(
...
     else:
-        LOGGER.info("Pin already loaded and chi not changed so doing nothing")
-        return
+        # Robot load normally sets the energy so we should do this explicitly if no load is
+        # being done
+        demand_energy_ev = parameters.demand_energy_ev
+        LOGGER.info(f"Setting the energy to {demand_energy_ev}eV")
+        if demand_energy_ev:
+            yield from set_energy_plan(
+                demand_energy_ev / 1000,
```
DominicOram, line 192: Could: Given we're now doing this same logic in the robot load (the only other place this plan is called) maybe we could just change `set_energy_plan` to take an optional in `kev` and move the `if` and `/1000` into there?

src/mx_bluesky/hyperion/experiment_plans/load_centre_collect_full_plan.py
```diff
@@ -85,4 +85,8 @@ def load_centre_collect_full(
             multi_rotation.rotation_scans.append(combination)
     multi_rotation = MultiRotationScan.model_validate(multi_rotation)
 
+    assert (
+        multi_rotation.demand_energy_ev
+        == parameters.robot_load_then_centre.demand_energy_ev
+    ), "Setting a different energy for gridscan and rotation is not supported"
```
DominicOram, line 91: Should: We should be able to support this quite easily. Can you make an issue for it?

DominicOram (changes requested):
Great, thanks A few comments.

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -46,31 +47,42 @@ def _apply_and_wait_for_voltages_to_settle(
...
 
     LOGGER.info(
-        f"Adjusting mirror stripe for {energy_kev}keV selecting {stripe} stripe"
+        f"Adjusting mirror stripe for {energy_kev}keV selecting {mirror_config['stripe']} stripe"
     )
-    yield from bps.abs_set(mirror.stripe, stripe, wait=True)
+    yield from bps.abs_set(mirror.stripe, mirror_config["stripe"], wait=True)
     yield from bps.trigger(mirror.apply_stripe)
 
+    # yaw, lat cannot be done simultaneously
```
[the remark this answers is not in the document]
rtuck99: Yaw and lat are compound axes of the front and back motors, so moving one will cancel any move of the other

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
rtuck99: That's why is called YAW_LAT_TIMEOUT and not LAT_TIMEOUT

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
rtuck99: The yaw and lat are part of the dcm, they change because of the mirror stripe change which is the VFM and a separate device, unless we want to roll up the VFM into the `undulator_dcm` to  make it `vfm_undulator_dcm`?

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
rtuck99: I don't think it should be done in the undulator_dcm, I'm beginning to think the opposite, that the `undulator_dcm` should just be a dumb device and everything be orchestrated in this plan.

src/mx_bluesky/hyperion/experiment_plans/load_centre_collect_full_plan.py
rtuck99: #686

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
rtuck99: Yes, I should have done this - fewer tests and simpler code!

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
DominicOram: D'oh, sorry

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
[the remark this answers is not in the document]
DominicOram: Great, thank, I've added this to the comment

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
```diff
@@ -109,31 +121,16 @@ def adjust_dcm_pitch_roll_vfm_from_lut(
...
     LOGGER.info("Waiting for DCM roll adjust to complete...")
 
-    # DCM Perp pitch
+    # DCM offset
+    # wait for other DCM motions and energy change to complete before changing offset
+    yield from bps.wait(DCM_GROUP, timeout=30)
     offset_mm = undulator_dcm.dcm_fixed_offset_mm
     LOGGER.info(f"Adjusting DCM offset to {offset_mm} mm")
-    yield from bps.abs_set(dcm.offset_in_mm, offset_mm, group=DCM_GROUP)
+    yield from bps.abs_set(dcm.offset_in_mm, offset_mm, wait=True)
```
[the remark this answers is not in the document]
DominicOram: Must: But, from memory, we needed that on the beamline because we needed the DCM energy to be set before we could set the DCM offset? So if we do this here we will need a wait for `UNDULATOR_GROUP` (which is poorly named as it is moving the DCM too) to finish first. I think that then costs us some time as we'll be ending up waiting on the undulator here too.

Should: Additionally, if we're going to do it here the `wait=True` is frustrating because AFAIK we don't need to wait on this to have finished before we start changing things on the mirrors as they're entirely separate devices.

General point: This is why I think we should push this further into the device, if we push it all the way into the DCM itself it's more obvious about what is the minimum thing we're waiting on.

src/mx_bluesky/hyperion/device_setup_plans/dcm_pitch_roll_mirror_adjuster.py
DominicOram: I think it's cleaner to put all the logic for a particular device into the device so a plan just needs to do `yield from bps.set(dcm, energy)` and the dcm handles all of it's rolls/pitches etc. Especially as IMHO the asyncio syntax for ordering operations is cleaner than the `wait=GROUP` syntax. But I think either way that's outside of the scope here, sorry for bringing it up in the first place

DominicOram (changes requested):
Thanks. Still some comments, sorry.

noemifrisina (changes requested):
lgtm, changes were addressed, please resolve the conflicts

noemifrisina (approved):

PR #682, opened 2024-11-27
Fix load_centre_collect imports and annotations

rtuck99:
This fixes `load_centre_collect_full` to be in line with other plans as it was missed out of the previous PR updating for blueapi changes #564 
Also it uses the `from __future__ import annotations` required to register the plan correctly.


noemifrisina (approved):
lgtm, just rebase on main before merging please

Issue #558, opened 2024-10-08
Use a callback for ispyb deposition


noemifrisina:

DominicOram:
The collections look like:
```
dataCollectionId: 15521233
                 BLSAMPLEID: NULL
                  SESSIONID: 27532659
             experimenttype: NULL
       dataCollectionNumber: NULL
                  startTime: 2024-10-09 02:29:10
                    endTime: 2024-10-09 02:32:57
                  runStatus: DataCollection Successful
                  axisStart: NULL
                    axisEnd: NULL
                  axisRange: 0.001
                    overlap: NULL
             numberOfImages: 6084
           startImageNumber: 0
             numberOfPasses: NULL
               exposureTime: 0.01
             imageDirectory: /dls/i24/data/2024/mx38934-5/.../
                imagePrefix: NULL
                imageSuffix: NULL
      imageContainerSubPath: NULL
               fileTemplate: xavi00052_#####.cbf
                 wavelength: 0.99987
                 resolution: 2.47
           detectorDistance: 490
                      xBeam: 220.968
                      yBeam: 225.079
                   comments: NULL
         printableForReport: 1
               CRYSTALCLASS: NULL
            slitGapVertical: NULL
          slitGapHorizontal: NULL
               transmission: 100
            synchrotronMode: NULL
      xtalSnapshotFullPath1: /...pia.png
      xtalSnapshotFullPath2: /.../index.png
      xtalSnapshotFullPath3: NULL
      xtalSnapshotFullPath4: NULL
               rotationAxis: NULL
                   phiStart: NULL
                 kappaStart: NULL
                 omegaStart: NULL
                   chiStart: NULL
         resolutionAtCorner: NULL
             detector2Theta: NULL
               DETECTORMODE: NULL
              undulatorGap1: NULL
              undulatorGap2: NULL
              undulatorGap3: NULL
          beamSizeAtSampleX: 0.007
          beamSizeAtSampleY: 0.007
            centeringMethod: NULL
         averageTemperature: NULL
        ACTUALSAMPLEBARCODE: NULL
ACTUALSAMPLESLOTINCONTAINER: NULL
     ACTUALCONTAINERBARCODE: NULL
    ACTUALCONTAINERSLOTINSC: NULL
    actualCenteringPosition: NULL
                  beamShape: NULL
      dataCollectionGroupId: 13123468
                 POSITIONID: NULL
                 detectorId: 58
[...]

PR #688, opened 2024-11-28
I24 serial: Stop writing map and parameter files

### LINKED ISSUE #460: I24 ssx: don't do file IO inside plans
Right now we're writing parameters and map files for serial collections directly in plans (eg. #189 ).
We should be using callback for this.

noemifrisina:
First part of #460 

Needs [nexgen#314](https://github.com/DiamondLightSource/nexgen/pull/314) and a new deployment with parameters updated in `nexgen-server`

- Stop writing parameter file along with data in collection directory, information for debugging is in logs
- Stop writing map file and upload pvars to geobrig directly at the start of the collection plan

A callback for the userlog at collection end will be done in a separate PR


noemifrisina:
Small note. The nexgen PR is now merged, the nexgen-server deployment will be done once this change is on the beamline to avoid breaking things beforehand

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -221,59 +230,24 @@ def define_current_chip(
...
-
-    SSX_LOGGER.warning("Automatic Setting Mapping Type to Lite has been disabled")
+    SSX_LOGGER.info(f"Chipid {ChipType.Oxford}, width {OXFORD_CHIP_WIDTH}")
+    for block in range(1, 65):
+        value = 1 if block in chip_map else 0
+        pvar = PVAR_TEMPLATE % block
+        pvar_str = f"{pvar}={value}"
+        SSX_LOGGER.debug(f"Set {pvar_str} for block {block}")
+        yield from bps.abs_set(pmac.pmac_string, pvar_str, wait=True)
+        sleep(0.02)
```
rtuck99, line 249: What is this wait for? Is this masking a controls issue. I feel like we ought to be able to either blast strings as fast as it will accept, or write them in bulk using a different mechanism.

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -141,12 +135,9 @@ def enter_hutch(
 
 
 @log_on_entry
-def write_parameter_file(
-    detector_stage: DetectorMotion, attenuator: ReadOnlyAttenuator
-):
+def read_parameters(detector_stage: DetectorMotion, attenuator: ReadOnlyAttenuator):
     """Writes a json parameter file that can later be parsed by the model."""
```
rtuck99, line 139: This comment needs updating

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -40,10 +38,6 @@
     log_on_entry,
 )
 from mx_bluesky.beamlines.i24.serial.parameters import ExtruderParameters
-from mx_bluesky.beamlines.i24.serial.parameters.constants import (
-    PARAM_FILE_NAME,
```
rtuck99, line 44: Can these now be removed?

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
```diff
@@ -489,8 +474,9 @@ def run_extruder_plan(
     start_time = datetime.now()
     SSX_LOGGER.info(f"Collection start time: {start_time.ctime()}")
 
-    yield from write_parameter_file(detector_stage, attenuator)
-    parameters = ExtruderParameters.from_file(PARAM_FILE_PATH / PARAM_FILE_NAME)
+    parameters: ExtruderParameters = yield from read_parameters(
```
rtuck99, line 477: Maybe we should have some tests for this plan

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: Essentially this is not just sending a string but once it's received the pmac does some things in the background for which we don't have any readback, the sleep is a "we worked out over testing that we should wait for this time so we're sure the pmac is done". You're right, there should be a better way of doing it, and it may involve asking controls but a starting solution for this might be to add this to the ophyd devices (the details haven't been worked out yet).
I'll add some comment to the code

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Ah thanks, completely missed it

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
```diff
@@ -122,6 +132,24 @@ def nexgen_experiment_type(self) -> str:
     def ispyb_experiment_type(self) -> SSXType:
```
rtuck99, line 132: There is an enum `IspybExperimentType` which has all of the ISPyB experiment types (that we know of)

noemifrisina: Yes, but until some other changes are done that leads to an annoying circular import... Also as this enum is used in multiple places in the code - not only for ispyb - that's a much bigger change than the scope of this PR (Should be a part of #558 )

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: Wait... this looks like it was removed already? Was this the line you were looking at?

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
noemifrisina: There are some in tests/unit_test/beamlines/i24/serial/i24/extruder/test_extruder_collect.py but admittedly may be incomplete, can you take a look and see if anything is missing please?

tests/unit_tests/beamlines/i24/serial/extruder/test_extruder_collect.py
```diff
@@ -1,4 +1,4 @@
-from unittest.mock import ANY, MagicMock, call, mock_open, patch
+from unittest.mock import ANY, MagicMock, call, patch
```
rtuck99, line 1: I think a lot of the tests in this file don't provide very much value. There's an awful lot of mocking of itty-bitty calls to `caput` `caget` `bps.rd` etc, and the actual tests are lots of `assert`s on call count but not asserting what the parameters to the call were, so we know mostly that something happened but not that it was correct.

I think I would rather have more focused tests that just test small individual aspects/code paths of the plans that shared common fixtures to cut down on boilerplate rather than one big woolly happy-path test that doesn't really check anything.

tests/unit_tests/beamlines/i24/serial/fixed_target/test_ft_collect.py
```diff
@@ -78,17 +78,20 @@ def test_write_userlog(fake_mkdir, fake_log, dummy_params_without_pp):
...
-        mock_open(read_data=chipmap_str),
-    ):
-        tot_num_imgs = datasetsizei24(2, dummy_params_without_pp.chip, MappingType.Lite)
-    assert tot_num_imgs == 800
+    dummy_params_without_pp.num_exposures = 2
+    dummy_params_without_pp.chip_map = [20]
+    assert dummy_params_without_pp.total_num_images == 800
+    set_datasize(dummy_params_without_pp)
     fake_caput.assert_called_once_with("ME14E-MO-IOC-01:GP10", 800)
+    assert fake_log.info.call_count == 1
```
rtuck99, line 93: If logging certain information about the parameters is an important piece of functionality, I think it is worth putting that into a specifically named test like `test_set_datasize_logs_parameter_info()`, and asserting the message contents, so that is clear that these logging statements shouldn't be removed later.

src/mx_bluesky/beamlines/i24/serial/extruder/i24ssx_Extruder_Collect_py3v2.py
rtuck99: No i meant the actual constants in `constants.py`

src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
rtuck99: that seems wrong if importing from `mx_bluesky.common.parameters` introduces a circular import on beamline-specific code

rtuck99 (approved):
Approved after discussions with @noemifrisina about the various issues

Issue #563, opened 2024-10-10
Multi-rotation and multipin plans don't support certain combinations that are implied by the parameter model


rtuck99:
Currently the plans only support a fixed number of frames per rotation, whereas the `MultiRotationScan`  `RotationScanPerSweep` allows variable rotation width which implies variable frames if you have fixed exposure per frame. Also there's a fixed `omega_start` in the `DetectorParams` which we currently assume to be the same as the first.

We should limit the flexibility allowed by the model if we decide not to implement this functionality, or at least validate that these constraints are maintained.

DominicOram:
Go through the parameter model and lock it down so that it only lets you change the things we actually currently support. More issues for allowing these things.

rtuck99:
One thing that is currently available in the model for the multi rotation scan is that `rotation_scans`   supports `x/y/z_start_um` whereas if doing the load_centre_collect plan the position is determined by the detected centres

PR #689, opened 2024-11-28
Improve mx_bluesky parameter validation

rtuck99:
This addresses #563  and adds additional parameter validation to prevent invalid combinations of parameters.

It also adds PlantUML parameter class diagram to the docs generated by CI


tests/unit_tests/hyperion/experiment_plans/test_load_centre_collect_full_plan.py
```diff
@@ -159,6 +168,99 @@ def test_can_serialize_load_centre_collect_params(load_centre_collect_params):
...
+        match="Sweeps with different numbers of frames are not supported.",
+    ):
+        LoadCentreCollect(**params)
+
+
+def test_params_with_start_xyz_is_rejected():
+    params = raw_params_from_file(GOOD_TEST_LOAD_CENTRE_COLLECT_MULTI_ROTATION)
+    params["multi_rotation_scan"]["rotation_scans"][1]["x_start_um"] = 1.0
+    params["multi_rotation_scan"]["rotation_scans"][1]["x_start_um"] = 2.0
+    params["multi_rotation_scan"]["rotation_scans"][1]["x_start_um"] = 3.0
```
olliesilvester, line 193: I assume you meant to change x, y and z here

olliesilvester (changes requested):
This is nice!

Two things:
- The image displayed on the docs for the parameter hierarchy is very small, could we make it easier to read? 
image

- This introduces a (small) extra step when building the docs locally. We now need to do:
```
python ./utility_scripts/generate_plantuml.py > docs/developer/hyperion/reference/param_hierarchy.puml
tox -e docs autobuild
```
Might be worth having a small utility script `locally_autobuild_docs` that does this for us

rtuck99:
>     * This introduces a (small) extra step when building the docs locally. We now need to do:
> 
> 
> ```
> python ./utility_scripts/generate_plantuml.py > docs/developer/hyperion/reference/param_hierarchy.puml
> tox -e docs autobuild
> ```
> 

I've changed things so that instead of having an extra step in the github workflow, there is a pre-command step in the `tox` configuration so that `tox -e docs` now runs the generation automatically.

Unfortunately I can't do much about the page layout but you can open the diagram in a new tab and see it full size.

olliesilvester (approved):

Issue #692, opened 2024-11-29
Hyperion testing 28/11


DominicOram:
Did some brief testing of https://github.com/DiamondLightSource/mx-bluesky/pull/685 on the beamline. Found some minor issues:

Error of:
```
ERROR:root:device `array_data` raised unexpected exception TypeError
Traceback (most recent call last):
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.1a/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/core/_device.py", line 172, in connect
    await self._connect_task
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.1a/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/core/_signal.py", line 67, in connect_real
    await self.backend.connect(timeout)
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.1a/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/epics/core/_p4p.py", line 324, in connect
    self.converter = make_converter(self.datatype, self.initial_values)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.1a/mx-bluesky/.venv/lib/python3.11/site-packages/ophyd_async/epics/core/_p4p.py", line 255, in make_converter
    raise TypeError(
TypeError: BL03I-DI-OAV-01:PVA:ARRAY with inferred datatype ndarray from typeid='epics:nt/NTNDArray:1.0' specifier='U' cannot be coerced to Array1D[np.uint8]
```
Fixed by changing the type to `np.ndarray`.

The docs at https://diamondlightsource.github.io/mx-bluesky/main/developer/general/how-to/deploy-a-release.html say to `pip install dodal` after `mx_bluesky` this ends up with the `dodal` from `pypi`, not the local one as:
1. Pip installing mx_bluesky installs the pypi version
2. If you try and pip install the local version pip recognises that you have the one from pypi and just says there's no need to install anything new

rtuck99:
* The pip fix was rejected in https://github.com/DiamondLightSource/mx-bluesky/pull/693
* OAV PinTipDetection fix merged in https://github.com/DiamondLightSource/dodal/pull/937

PR #693, opened 2024-11-29
Clean up order of deploying

DominicOram:
Fixes #https://github.com/DiamondLightSource/mx-bluesky/issues/692

To test:
* Do a full deployment based on the instructions
* Or just trust my reasoning on the issue ;)

### DIFF
--- a/docs/developer/general/how-to/deploy-a-release.rst
+++ b/docs/developer/general/how-to/deploy-a-release.rst
@@ -53,5 +53,5 @@ For building and deploying a Docker image please see :doc:`../../hyperion/deploy
     .. code:: console
 
         . ./.venv/bin/activate
-        pip install -e .
         pip install -e ../dodal
+        pip install -e .


noemifrisina (approved):

docs/developer/general/how-to/deploy-a-release.rst
```diff
@@ -53,5 +53,5 @@ For building and deploying a Docker image please see :doc:`../../hyperion/deploy
     .. code:: console
 
         . ./.venv/bin/activate
-        pip install -e .
         pip install -e ../dodal
```
rtuck99, line 56: This is wrong and will result in the reverse of what you want.
If you `pip install -e .`, this will install an editable version of `mx-bluesky` from the local git folder, however it will also install a non-editable version of `dls-dodal` from pypi, _overriding the editable version previously installed_

Because our deployments are based on editable versions of the folder, we want to install `mx-bluesky` first and `dls-dodal` second.

PR #694, opened 2024-12-03
parse the RunStop reason field instead of using events, following discussion with bluesky

rtuck99:
Relates to #716 

Following a discussion with bluesky team, this PR removes the event handler introduced in 
https://github.com/DiamondLightSource/mx-bluesky/pull/606
in favour of parsing the `reason` field in the RunStop document

### DIFF
--- a/src/mx_bluesky/hyperion/external_interaction/callbacks/sample_handling/sample_handling_callback.py
+++ b/src/mx_bluesky/hyperion/external_interaction/callbacks/sample_handling/sample_handling_callback.py
@@ -1,16 +1,5 @@
-import dataclasses
-from collections.abc import Generator
-from functools import partial
-from typing import Any
+from event_model import RunStart, RunStop
 
-import bluesky.plan_stubs as bps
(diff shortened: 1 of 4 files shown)


olliesilvester (approved):
Nice, this is much better than what we had before. Unfortunately I took ages to review this so now there are merge conflicts :(

src/mx_bluesky/hyperion/external_interaction/callbacks/sample_handling/sample_handling_callback.py
```diff
@@ -20,26 +9,6 @@
...
-    yield from bps.read(_ExceptionEvent(type(exception).__name__))
-    yield from bps.save()
-
-
-sample_handling_callback_decorator = make_decorator(
-    partial(contingency_wrapper, except_plan=_exception_interceptor)
-)
 
 
 class SampleHandlingCallback(PlanReactiveCallback):
```
olliesilvester, line 14: Instead of saying that this callback intercepts exceptions, this now checks the RunStop document for exceptions

PR #594, opened 2024-10-24
Use ophyd_async OAV

noemifrisina:
Needs [Dodal#857](https://github.com/DiamondLightSource/dodal/pull/857)

- Check tests pass
- Check no functionality has been changed in the code

PR #695, opened 2024-12-03
Fix system tests for oav zoom controller and load_centre_collect params

rtuck99:
This adds missing zoom levels to the oav zoom controller following #594 
It also updates the `demand_energy_ev` test data for system test parameters to be valid following #676


noemifrisina (approved):
lgtm, thanks

Issue #697, opened 2024-12-04
 Issues from mx-bluesky testing 2024-12-04


rtuck99:
```
python ./utility_scripts/deploy/deploy_mx_bluesky.py i03
  File "/scratch/ws/mx-bluesky/.venv/lib/python3.11/site-packages/git/remote.py", line 895, in _get_fetch_info_from_stderr
    proc.wait(stderr=stderr_text)
  File "/scratch/ws/mx-bluesky/.venv/lib/python3.11/site-packages/git/cmd.py", line 834, in wait
    raise GitCommandError(remove_password_if_present(self.args), status, errstr)
git.exc.GitCommandError: Cmd('git') failed due to: exit code(1)
  cmdline: git fetch -v -- origin refs/tags/*:refs/tags/*
```

occurs when have a tag that differs from those in origin

* deploy script should print something it took ages with the slow fs, see #242 - we should fix the script as k8s is still some way off from production. stdout, stderr redirection fixed on beamline/local dev but also (possibly related) issue with borked `pip install` and script not exiting cleanly #700 
* `deploy_hyperion_tok8s.sh` `ensure_version_py()` doesn't work properly if a non-k8s installation has already been performed as .venv already exists but without setuptools_scm - we should check for this and install it #700
* The warning that we are installing the wrong version to k8s isn't very prominent #700 
* `deploy_hyperion_to_k8s.sh` references the wrong folders due to various hyperion/mx-bluesky renaming issues #700 
* Various confusions in k8s deployment script due to version with or without `v` prefix, should standardise on supplying without and then adding it where necessary #701 
* The deployed container image no longer starts, with log message:
```
usage: mx-bluesky [-h] [-v]
mx-bluesky: error: unrecognized arguments: --external-callbacks
```
#701
* ```'HyperionLoadCentreCollect' object has no attribute 'context'``` - fixed on beamline
* ~rotation nexus files are reported in ispyb from the wrong (i.e. previous) rotation~
* GPU processing not enabled (/dev/shm not set up) but is waiting for gpu processing: https://github.com/DiamondLightSource/mx-bluesky/issues/738
* Changing the omega_start value doesn't appear to generate the desired results as nexus files now look wrong with the change, needs more investigation with looking at actual smargon motor values from test data to figure out what changes are actually needed. #247

rtuck99:
Testing results from 2024-12-05:

* Ran into issue where gridscan fails with 
```
[...]

PR #701, opened 2024-12-06
700 fix deployment issues

### LINKED ISSUE #700: Beamline fixes from testing 2024-12-04
This is to address issues in #697 that were fixed on the beamline during testing:

* Fix for #242 to make the deployment scripts output to the console when they ssh into the control server
* Fix the `deloy_hyperion_to_k8s.sh` script to refer to the correct folder locations and app version strings which changed due to renames elsewhere
* Discard zocalo results with total_count below a threshold value so that background noise is ignored on non-diffracting samples.
* Apply chi change when the specified chi is 0

Also fixes the following:
* Always make sure `setuptools_scm` is installed when running `deploy_hyperion_to_k8s.sh`

## Acceptance Criteria
- `deploy_hyperion_to_k8s.sh` deploys the image without error and works for side-by-side deployment with non-containerised version
- Non-diffracting samples always return zero hits and rotations are therefore skipped
- Chi is always applied when specified as 0 in the request to hyperion and restores the chi before the pin-detection after a previous collection at non-zero chi.
- `deploy_mx_bluesky.py` outputs to the console the install scripts being executed on the control server

rtuck99:
Fixes #700 

* The deploy script `deploy_hyperion_to_k8s.sh` now deploys the image correctly (but see also #697 where the image is broken, to be fixed in another PR)
* `deploy_mx_bluesky.py` now outputs to the console when remoting to the control server
* Zocalo results below a threshold of 3 are now ignored
* The requested chi is now not ignored if it is 0

Also fixes the following described in #697 :

* Always make sure setuptools_scm is installed when running deploy_hyperion_to_k8s.sh
* Make the version check during k8s deployment more prominent and fix the pathname munging so that it compares correctly
* Fix where the an image was built with `Dockerfile.release` but an image built with `Dockerfile` was the one actually pushed

### DIFF
--- a/.github/workflows/_container.yml
+++ b/.github/workflows/_container.yml
@@ -51,6 +51,7 @@ jobs:
         # Docker cache and publish it
         with:
           context: .
+          file: Dockerfile.release
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/src/mx_bluesky/hyperion/device_setup_plans/manipulate_sample.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/manipulate_sample.py
@@ -100,11 +100,11 @@ def move_phi_chi_omega(
     axes are optional."""
 
     LOGGER.info(f"Moving smargon to phi, chi, omega: {(phi, chi, omega)}")
-    if phi:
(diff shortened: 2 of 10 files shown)


noemifrisina (approved):
lgtm, please rebase before merging

PR #704, opened 2024-12-09
Fixes for ophyd async 0.9.0a1

rtuck99:
This fixes mx-bluesky to work with the latest dodal pre-release 0.9.0a1

See also related dodal PR DiamondLightSource/dodal#952


noemifrisina (approved):
lgtm

PR #708, opened 2024-12-10
I24 serial: Add some detector specific params in parameter model and beam center calculation

### LINKED ISSUE #647: I24 serial: set the beam center using lookup tables
Set up a plan that the beamline staff can use from the command line to update the beam center values at the beginning of each run.

### LINKED ISSUE #684: I24 Serial: Parameter model should contain a DetectorParams
In Hyperion we call `detector_params` on our parameter models to give us information about the detector (see https://github.com/DiamondLightSource/mx-bluesky/blob/e710442771a6698003dedb2c3db4b021902eee85/src/mx_bluesky/common/parameters/gridscan.py#L39). We should have the same interface in the i24 SSX code.

## Acceptance Criteria
- There is a detector_params property on the I24 SSX parameter model
- Calling it returns a `DetectorParams` with most information filled in

noemifrisina:
This should also take care of the beamcenter calculation.
UPDATE: This will not close #684 as it won't use the DetectorParams in the form they are now, but only add the property currently needed to the parameter model. 
Closes #647 

Needs [Dodal#955](https://github.com/DiamondLightSource/dodal/pull/955)
Needs also [Dodal#981](https://github.com/DiamondLightSource/dodal/pull/981)


src/mx_bluesky/beamlines/i24/serial/parameters/experiment_parameters.py
```diff
@@ -34,7 +39,9 @@ def _parse_visit(cls, visit: str | Path):
 
     @property
     def collection_directory(self) -> Path:
-        return Path(self.visit) / self.directory
+        directory = Path(self.visit) / self.directory
```
rtuck99, line 42: This property is accessed in 9 different places in production code, I don't think it should be creating the directory.

rtuck99:
Ok, so I've had a bit of a think, and I think there are a number of problems with `DetectorParams` which we should address. I've written them up in
* DiamondLightSource/dodal#966

I'm reluctant to approve this PR because I think on balance on its own it makes things worse rather than better, in that it spreads usage of `DetectorParams` without addressing any of its shortcomings, and it also doesn't really use it enough to justify doing so.

I think a better approach would be to address part of DiamondLightSource/dodal#966 relating to the beam_xy_converter, make i24 and hyperion code both better in an incremental fashion.

rtuck99 (changes requested):
See previous comment

noemifrisina:
After going through this PR with @rtuck99 and a lot of discussions on https://github.com/DiamondLightSource/dodal/issues/966, I've decided to revert all the changes using the current DetectorParams as that is going to be a separate big refactoring job in dodal - involving also the hyperion side of the code. This will be done in a separate branch to avoid holding up other work.

As for this PR, it now only instantiates the detector constants  needed in the parameter model in this PR, as well as uses the linear_interpolation_lut function from dodal to calculate the beam center position, which are the bits of DetectorParams needed by serial at this time.

rtuck99 (approved):
Approved. I await the next exciting instalment in this continuing saga

Issue #272, opened 2024-03-08
Deploy should delete old deployments


DominicOram:
We're starting to get to a point where manually removing old deployments is becoming tedious. Each deployment is ~1.7GB so we should try and prune them somewhat regularly. It would be good if deploying a new version also did this pruning.

## Acceptance Criteria
* We decide how many deployments we want to keep (I suggest 4?)
* The deployment script asks if you want to delete the old ones
* If the old ones are symlinked to `latest` or `stable` they're not deleted
* Otherwise old deployments are deleted

PR #710, opened 2024-12-11
Deployment script deletes old deployments

shihab-dls:
These changes add a --prune-deployments flag, keeping 4 most recent deployments, and deleting the excess if not symlinked to hyperion_stable or hyperion_latest, following #272

### DIFF
--- a/utility_scripts/deploy/deploy_mx_bluesky.py
+++ b/utility_scripts/deploy/deploy_mx_bluesky.py
@@ -10,7 +10,7 @@ import subprocess
 from typing import NamedTuple
 from uuid import uuid1
 
-from create_venv import run_process_and_print_output
+from create_venv import run_process_and_print_output, setup_venv
 from git import Repo
 from packaging.version import VERSION_PATTERN, Version
 
@@ -231,8 +231,7 @@ def main(beamline: str, options: Options):
                     mx_repo, path_to_create_venv, path_to_dls_dev_env, beamline
                 )
             else:
-                # setup_venv(path_to_dls_dev_env, mx_repo.deploy_location)
-                pass
+                setup_venv(path_to_dls_dev_env, mx_repo.deploy_location)
 
     # If on beamline I24 also deploy the screens to run ssx collections
     if beamline == "i24":
--- a/utility_scripts/deploy/deploy_mx_bluesky.py
+++ b/utility_scripts/deploy/deploy_mx_bluesky.py
@@ -163,7 +163,7 @@ def _prune_old_deployments(release_area: str):
             for deployment in sorted_deployments[MAX_DEPLOYMENTS:]
         }
 
-        # Removes old deployments with symlinks
+        # Excludes deployments that are symlinked
         for link in symlinks:
             link_path = os.path.dirname(os.readlink(os.path.join(release_area, link)))
             if link_path in full_path_old_deployments:
--- a/utility_scripts/deploy/deploy_mx_bluesky.py
+++ b/utility_scripts/deploy/deploy_mx_bluesky.py
@@ -5,11 +5,12 @@ Deploy latest release of mx-bluesky either on a beamline or in dev mode.
 import argparse
 import os
 import re
+import shutil
 import subprocess
 from typing import NamedTuple
 from uuid import uuid1
 
-from create_venv import run_process_and_print_output, setup_venv
+from create_venv import run_process_and_print_output
 from git import Repo
 from packaging.version import VERSION_PATTERN, Version
 
@@ -40,6 +41,7 @@ class Options(NamedTuple):
     quiet: bool = False
     dev_mode: bool = False
     use_control_machine: bool = True
+    prune_deployments: bool = False
     # NOTE For i24 for now will need to set it to false from the command line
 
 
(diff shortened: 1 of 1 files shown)


utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
             process.kill()
 
 
+def _prune_old_deployments(release_area: str):
+    def get_creation_time(deployment):
+        return os.path.getctime(os.path.join(release_area, deployment))
```
olliesilvester, line 141: Apparently, on linux, `getctime` gives you the time of the last metadata change of that file - which isn't always the file creation time (eg if permissions change then the time will show this instead). 

I think it's fine though. Worst case, an old file gets changed metadata and then we end up not deleting it. I'd just add a comment briefly describing this behavior incase it confuses someone later on

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
...
+            if (
+                os.path.dirname(os.readlink(os.path.join(release_area, link)))
+                in full_path_old_deployments
+            ):
+                full_path_old_deployments.remove(
+                    os.path.dirname(os.readlink(os.path.join(release_area, link)))
+                )
+
+        for old_deployment in full_path_old_deployments:
+            print(f"Deteling old deployment {os.path.basename(old_deployment)}")
```
olliesilvester, line 170:
```suggestion
            print(f"Deleting old deployment {os.path.basename(old_deployment)}")
```

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
...
+            set_of_deployments.remove(item)
+
+    sorted_deployments = sorted(set_of_deployments, key=get_creation_time, reverse=True)
+    if len(sorted_deployments) > max_deployments:
+        full_path_old_deployments = {
+            os.path.join(release_area, deployment)
+            for deployment in sorted_deployments[max_deployments:]
+        }
+
+        for link in symlinks:
```
olliesilvester, line 160: nit: I think it's slightly more readable to put `os.path.dirname(os.readlink(os.path.join(release_area, link)))` in a variable rather than write the line twice

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
             process.kill()
 
 
+def _prune_old_deployments(release_area: str):
```
olliesilvester, line 139: Could: put some short comments in this function to describe what each chunk of logic is doing

olliesilvester (changes requested):
Thanks, looks mostly good, just some minor comments

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -41,6 +42,7 @@ class Options(NamedTuple):
     quiet: bool = False
     dev_mode: bool = False
     use_control_machine: bool = True
+    prune_deployments: bool = False
     # NOTE For i24 for now will need to set it to false from the command line
```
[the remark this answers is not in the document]
shihab-dls: Got it, doing this now:
`    quiet: bool = False
    dev_mode: bool = False
    prune_deployments: bool = True

# NOTE For i24 for now will need to set it to false from the command line
use_control_machine: bool = True
    use_control_machine: bool = True`

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
             process.kill()
 
 
+def _prune_old_deployments(release_area: str):
+    def get_creation_time(deployment):
+        return os.path.getctime(os.path.join(release_area, deployment))
+
+    max_deployments = 4
```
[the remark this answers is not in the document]
shihab-dls: Way better, doing this now:
`MAX_DEPLOYMENTS = 4`

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +136,41 @@ def _create_environment_from_control_machine(
...
+    deployments = os.listdir(release_area)
+    set_of_deployments = set(deployments)
+
+    symlinks = set()
+    for item in deployments:
+        if item in ["hyperion", "hyperion_stable", "hyperion_latest"]:
```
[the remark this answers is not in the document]
shihab-dls: This is fair; initially I thought symlinks not `hyperion_stable/hyperion_latest` **should** be deleted if targeting an old deployment. But it's probably best/safer to exclude any symlink target from the pruning.

utility_scripts/deploy/deploy_mx_bluesky.py
shihab-dls: Makes sense! Amended.

utility_scripts/deploy/deploy_mx_bluesky.py
shihab-dls: Got it; added brief descriptions to main blocks of logic

utility_scripts/deploy/deploy_mx_bluesky.py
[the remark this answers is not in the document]
olliesilvester: I'm probably being overcautious tbf

olliesilvester (approved):
Thanks, looks good!

utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -134,6 +139,42 @@ def _create_environment_from_control_machine(
...
+    sorted_deployments = sorted(set_of_deployments, key=get_creation_time, reverse=True)
+
+    # Excludes most recent deployments
+    if len(sorted_deployments) > MAX_DEPLOYMENTS:
+        full_path_old_deployments = {
+            os.path.join(release_area, deployment)
+            for deployment in sorted_deployments[MAX_DEPLOYMENTS:]
+        }
+
+        # Removes old deployments with symlinks
```
olliesilvester, line 166: I know i just approved but I think this comment is a bit misleading actually - sounds like we're deleting the deployments with symlinks

PR #714, opened 2024-12-12
Error when inputting nonexistant feature flags

### LINKED ISSUE #620: Warn if we supplied non-existant features to FeatureFlags
After https://github.com/DiamondLightSource/mx-bluesky/pull/496, we can overwrite feature toggles by supplying them in GDA. However, we can also supply non-existent feature toggles and, while nothing will fail, we also don't have an easy way of knowing if the features we have tried to use could be found

## Acceptance Criteria
- In its `model_validator`, `FeatureFlags` compares the input it received with the parameter it is expecting, and a warning is logged if there is a mismatch

shihab-dls:
This should fix #620 by checking if the overridden feature flags are defined. Currently, I have this set to raise a ValueError with context of the incorrect passed feature, but maybe we want to simply warn and move on.

### DIFF
--- a/src/mx_bluesky/common/external_interaction/config_server.py
+++ b/src/mx_bluesky/common/external_interaction/config_server.py
@@ -29,7 +29,18 @@ class FeatureFlags(BaseModel, ABC):
     def mark_overridden_features(cls, values):
         assert isinstance(values, dict)
         values["overriden_features"] = values.copy()
+        cls.validate_overridden_features(values)
         return values
+
+    @classmethod
+    def validate_overridden_features(cls, values: dict):
+        """Validates overridden features to ensure they are defined in the model fields."""
+        defined_fields = cls.model_fields.keys()
+        invalid_features = [key for key in values.keys() if key not in defined_fields]
+
+        if invalid_features:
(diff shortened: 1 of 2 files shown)


src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -29,8 +29,19 @@ def get_config_server() -> ConfigServer: ...
     def mark_overridden_features(cls, values):
         assert isinstance(values, dict)
         values["overriden_features"] = values.copy()
+        cls.validate_overridden_features(values)
         return values
 
+    @classmethod
+    def validate_overridden_features(cls, values: dict):
```
olliesilvester, line 36: Nit: Function name might be better with an appended underscore since it's only called internally

olliesilvester (approved):
Great, thanks! Just one optional comment

olliesilvester:
Feel free to merge whenever

src/mx_bluesky/common/external_interaction/config_server.py
shihab-dls: Makes sense! Have amended.

Issue #221, opened 2024-07-25
Document grid scan coordinate systems


rtuck99:
Rotations after xray centring sometimes are not aligned with the centre of the crystal, see:

Xray centred crystal: https://ispyb.diamond.ac.uk/dc/visit/cm37235-3/id/14661060
Manually centred same crystal: https://ispyb.diamond.ac.uk/dc/visit/cm37235-3/id/14661069

Observations after discussion with Mark, possibly this is multi-factorial:

* In `zocalo_results.py` we subtract a 0.5 box-width amount from the coordinates we get back from zocalo - it's not clear to me if this is correct or not, not knowing enough about how the analysis interprets its input data or what it outputs
* the number of data points for small crystals is very small (for some it may be a crystal only 2 x 1 x 1 boxes) and this may be affecting the quality of the returned centre-of-mass coordinate depending on the assumptions made in the analysis
* The heatmap superimposed on the grid in ispyb is somewhat misleading as it suggests we have more data than we really do - a discretised flat shading per box would be more representative than smooth shading interpolated from the limited data points

* Suggest that we reduce the grid size to collect more points per crystal, as per DiamondLightSource/mx-bluesky#231 
* Investigate if the 0.5 grid size correction is actually the correct thing to do.

DominicOram:
We need to go through the whole stack and work out the coordinate systems and write a nice document on it

rtuck99:
In testing on 1.4.1 #697 we are now seeing that ~about 10% of the time we are getting gridscans failing with 
```
File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.1/dodal/src/dodal/devices/fast_grid_scan.py", line 107, in grid_position_to_motor_position
    raise IndexError(f"{grid_position} is outside the bounds of the grid")
```
where the x-axis of the crystal centre is -0.5 due to the applied correction and this falls outside a santity check for the grid box

rtuck99:
PRs raised for dodal and mx-bluesky. There is also a PR TBC for `dlstbx`, which also should have https://github.com/DiamondLightSource/python-dlstbx/pull/253 reverted

PR #715, opened 2024-12-12
221 Document gridscan coordinate systems

rtuck99:
This addresses #221 and improves documentation about the reporting of the Zocalo Results 

See also DiamondLightSource/dodal#958

It also fixes an issue where the bounding box returned in the `XRayCentreResult` is off by 0.5 of a grid-box, however I don't think this impacts anything as this attribute isn't used yet.

### DIFF
--- a/src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
+++ b/src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
@@ -290,6 +290,8 @@ def _xrc_result_in_boxes_to_result_in_mm(
     # the corners of the box in motor-space; we do not apply this correction
     # to the xray-centre as it is already in continuous space and the conversion has
     # been performed already
+    # In other words, xrc_result["bounding_box"] contains the position of the box centre,
+    # so we subtract half a box to get the corner of the box
     return XRayCentreResult(
         centre_of_mass_mm=xray_centre,
         bounding_box_mm=(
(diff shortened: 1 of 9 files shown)


src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
```diff
@@ -286,14 +286,18 @@ def _xrc_result_in_boxes_to_result_in_mm(
     xray_centre = fgs_params.grid_position_to_motor_position(
         np.array(xrc_result["centre_of_mass"])
     )
+    # A correction is applied to the bounding box to map discrete grid coordinates to
+    # the corners of the box in motor-space; we do not apply this correction
+    # to the xray-centre as it is already in continuous space and the conversion has
+    # been performed already
     return XRayCentreResult(
         centre_of_mass_mm=xray_centre,
         bounding_box_mm=(
```
olliesilvester, line 295: As discussed the entire logic here is quite hard to explain via comments, but I think this helps to explain the 0.5
```suggestion
        # Bounding box coordinates give position of the centre of the box, subtract half a box size to get the start of the box
        bounding_box_mm=(
```

olliesilvester (approved):
Thanks! I think this part of the code is always going to be confusing, but it's much more understandable after your changes

src/mx_bluesky/hyperion/experiment_plans/flyscan_xray_centre_plan.py
olliesilvester: Whoops, didn't mean to delete line 294 in this suggestion

PR #717, opened 2024-12-13
Pinning pydantic extra types to 2.10.1

shihab-dls:
Encountered the pydantic error `TypeError: SemanticVersion() takes no arguments` when running tests; this was due to pydantic-extra-types being at 2.10.0 instead of 2.10.1.
Request to pin this to => 2.10.1


rtuck99 (approved):
Approved. Thanks!

PR #718, opened 2024-12-16
References to undulator and dcm in underlator_dcm device

shihab-dls:
PR [957](https://github.com/DiamondLightSource/dodal/pull/957) in dodal uses `Reference` to reference undulator and dcm devices in undulator_dcm composite. This PR amends mx-bluesky references to undulator_dcm.


olliesilvester (approved):
Thanks!

Issue #611, opened 2024-11-04
 Remove `_fake_grid_detection` and use real grid detection instead


DominicOram:
In a similar issue to https://github.com/DiamondLightSource/mx-bluesky/issues/479 we have some tests that are running against `_fake_grid_detection`. This means if we update the grid detection prod code we need to update this too and so they will inevitably get out of sync. Instead all our tests should be run against the real grid detection prod code.

## Acceptance Criteria
- `_fake_grid_detection` is removed
- Tests that used it instead use production grid detection code

PR #721, opened 2024-12-18
Use production grid detection plan in unit tests

shihab-dls:
This should fix issue #611, where references to `_fake_grid_detection` have been removed and replaced with logic that runs the real grid detection plan.


tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -151,15 +174,22 @@ def _do_detect_grid_and_gridscan_then_wait_for_backlight(
...
     test_config_files: dict,
 ):
     oav_params = OAVParameters("xrayCentring", test_config_files["oav_config_json"])
 
-    mock_grid_detection_plan.side_effect = _fake_grid_detection
+    _setup_grid_detection(
+        mock_rd, mock_pre_centring_setup_oav, mock_wait_for_tip_to_be_found
+    )
+
+    test_full_grid_scan_params.grid_width_um = 200
```
shihab-dls, line 191: I realize that I had to change this parameter to achieve the correct number of triggers in the assertions.
I believe the underlying processes are the same, but the combination of mocked `bps.rd` values results in the difference in triggers that requires this change in `grid_width_um`.

Under the assumption that the underlying processes are the same, changing the asserted number of triggers to match that expected by the mocked `bps.rd` value seems reasonable. But I should look into what values are required for the expected number of triggers first...

tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -35,34 +35,34 @@
...
+
+def _fake_generator(value):
+    yield from bps.null()
+    return value
+
+
+def _setup_grid_detection(
+    mock_rd, mock_pre_centring_setup_oav, mock_wait_for_tip_to_be_found
+):
+    mock_rd.side_effect = _fake_generator_array(
```
olliesilvester, line 54: Variable name is a little vague here - and it would be useful to have a comment explaining what each of these values refer to in the grid detection plan

olliesilvester (commented):
Only had a quick glance, looks reasonable but I will check the values more thoroughly another time. Just one comment for now

tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py
```diff
@@ -81,11 +81,21 @@ def grid_detect_devices_with_oav_config_params(
...
 ) -> GridDetectThenXRayCentreComposite:
     set_mock_value(grid_detect_devices.oav.zoom_controller.level, "7.5x")
+    grid_detect_devices.pin_tip_detection.triggered_bottom_edge = MagicMock()
+    grid_detect_devices.pin_tip_detection.triggered_top_edge = MagicMock()
     return grid_detect_devices
 
 
 @patch(
-    "mx_bluesky.hyperion.experiment_plans.grid_detect_then_xray_centre_plan.grid_detection_plan",
+    "mx_bluesky.hyperion.experiment_plans.oav_grid_detection_plan.bps.rd",
```
rtuck99, line 90: Intercepting all calls to bps.rd with a mock is pretty wild! I think there are better ways to do this

rtuck99:
I've just pushed a commit with an outline sketch of what I think is a better way setup the test fixtures, that should make it more robust and enable more reuse

rtuck99 (changes requested):
See above

rtuck99 (approved):
Approved

PR #723, opened 2024-12-19
Update attenuator import paths and names

olliesilvester:
Accompanies the change for https://github.com/DiamondLightSource/dodal/pull/969


rtuck99 (approved):
Approved, pending anything outstanding on the dodal PR. I don't really have anything to add to the dodal PR, other than I think @DiamondJoseph  has already said which I agree with

Issue #618, opened 2024-11-05
 Ensure aperture selection based on bounding box takes into account step size


DominicOram:
After https://github.com/DiamondLightSource/mx-bluesky/pull/542 the box size of an XRC should be able to be varied. However, the aperture selection, which is based on the bounding box is still calculated using the fixed box size. We should instead change the calculation to be if the bounding box is less than 50um select medium, otherwise large. Additionally we should document what the bounding box is as it's a bit confusing

## Acceptance Criteria
- Aperture selection is based on whether the bounding box is less than 50um or not
- Bounding box is documented better in docstrings

olliesilvester:
@shihab-dls Discussed with @rtuck99. The outcome is:
We only care about the bounding box size (in mm). We can remove the weird conversion of 'bounding box units mm to units boxes'

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/724

PR #724, opened 2024-12-20
Select aperture based on bounding box size

shihab-dls:
This addresses #618, where a fixed `grid_width` was assumed and used to calculate how many 'boxes' (i.e., a grid box) a bounding box would occupy, affecting the aperture size required.

After conversations with Ollie and Rob, the intended behavior was decided to be changing the aperture based on the size of a bounding box in millimeters, requiring a size of < 0.05mm to select medium aperture. Here, 'bounding box size' solely considers the x-component of a three-dimensional bounding box.

This removed the need for any conversions based on grid boxes.


src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py
```diff
@@ -49,22 +50,20 @@ def change_aperture_then_move_to_xtal(
...
-
-
-def set_aperture_for_bbox_size(
+def set_aperture_for_bbox_mm(
     aperture_device: ApertureScatterguard,
-    bbox_size: list[int] | numpy.ndarray,
+    bbox_size: list[float] | numpy.ndarray,
 ):
+    """bbox_size represents the [x,y,z] lengths, in mm, of a bounding box
+    containing a crystal. This discribes (in no particular order):
```
rtuck99, line 58:
```suggestion
    containing a crystal. This describes (in no particular order):
```

src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py
```diff
@@ -49,22 +50,20 @@ def change_aperture_then_move_to_xtal(
...
-    yield from set_aperture_for_bbox_size(aperture_device, bbox_size_boxes)
-
-
-def set_aperture_for_bbox_size(
+def set_aperture_for_bbox_mm(
     aperture_device: ApertureScatterguard,
-    bbox_size: list[int] | numpy.ndarray,
+    bbox_size: list[float] | numpy.ndarray,
 ):
+    """bbox_size represents the [x,y,z] lengths, in mm, of a bounding box
```
rtuck99, line 57: Docstring comments should be in Google style:
https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

This makes any API documentation we generate look nice, and also in your IDE if you configure it

src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py
```diff
@@ -49,22 +50,20 @@ def change_aperture_then_move_to_xtal(
...
+    containing a crystal. This discribes (in no particular order):
+    * The maximum width a crystal occupies
+    * The maximum height a crystal occupies
+    * The maximum depth a crystal occupies
+    Constructing a three dimensional cuboid, completely encapsulating the crystal."""
+
     # bbox_size is [x,y,z], for i03 we only care about x
     new_selected_aperture = (
-        ApertureValue.MEDIUM if bbox_size[0] < 2 else ApertureValue.LARGE
+        ApertureValue.MEDIUM if bbox_size[0] < 0.05 else ApertureValue.LARGE
```
rtuck99, line 66: bbox_size_mm please

tests/system_tests/hyperion/test_aperturescatterguard_system.py
```diff
@@ -39,5 +39,5 @@ def test_aperture_change_callback(ap_sg: ApertureScatterguard):
     cb = ApertureChangeCallback()
     RE = RunEngine({})
     RE.subscribe(cb)
-    RE(set_aperture_for_bbox_size(ap_sg, [2, 2, 2]))
+    RE(set_aperture_for_bbox_mm(ap_sg, [2, 2, 2]))
```
rtuck99, line 42: I think the input parameters for the test should be more realistic. 2mm is very large for a crystal!

rtuck99: Also, although we do incidentally test the aperture in quite a few other places, I think parametrizing this test so that we can check the threshold is obeyed would still be a good idea.

Overall though, good work

rtuck99 (changes requested):

src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py
shihab-dls: Ah, got it! I've amended the docstring now; do let me if this follows convention now, thank you!

tests/system_tests/hyperion/test_aperturescatterguard_system.py
shihab-dls: Sounds good! I've dropped this down to 0.02mm, which should switch the aperture to Medium, and parametrized the test with 0.05mm which should use Large, comparing against an `expected_aperture` for each case.

PR #725, opened 2024-12-20
698 check the beamstop is in position

### LINKED ISSUE #698: Hyperion should check the beamstop is in position before collection
Hyperion doesn't currently move the beamstop into the correct position or check that it is in position before running collections.
This was discovered during testing (see #697) 

Before doing a gridscan or a rotation, hyperion should check that the beamstop is in position and if it is not abort with an error.

This is something we should do as an interim measure until #484 is addressed - moving the beamstop is not always safe to do depending on the initial state of the beamline, and so implementing the move correctly is likely to be a complex change requiring analysis of existing GDA behaviour

## Acceptance Criteria
- Hyperion does not proceed and aborts with an error if the beamstop is not in the correct position

rtuck99:
* Fixes #698 
* See also DiamondLightSource/dodal#970

This change introduces a check for the beamstop position for hyperion, when `load_centre_collect_full` is invoked, a check will be performed against the beamstop position, if it does not match the position specified in the `beamlineParameters` file `in_beam_x/y/z_STANDARD` properties then the plan will abort with a `BeamstopException` and the BLSampleStatus will be `ERROR - beamline`.
GDA should not continue to process requests.


olliesilvester (approved):
Looks good, thanks. Some optional comments.

Also, I would've thought we also need to do a `check_beamstop` in `rotation_scan_plan`, since I thought we weren't doing the `load_centre_collect_full` in production just yet. Unless I'm outdated, could you add that too before merging

Issue #518, opened 2024-09-23
 Check undulator gap on each collection


DominicOram:
There have been issues in the past where, after a beam dump, undulator gap has not come back correctly. As such we should check that it is correct just before running each collection just after the XBPM feedback is stable.

## Acceptance Criteria
- The undulator gap is checked on each collection
- If it is wrong we correct it to what it should be given the current energy

PR #726, opened 2025-01-03
Set undulator gap after XBPM is stable

shihab-dls:
Should address #518, where Undulator gap is incorrect following beam dump.

the Undulator device checks if we are already at the correct gap when set; therefore, this PR sets the gap after XBPM is stable (in transmission_and_xbpm_feedback_for_collection_decorator), and before collection, without checking first.

### DIFF
--- a/src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
+++ b/src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
@@ -305,6 +305,10 @@ def _move_and_rotation(
     params: RotationScan,
     oav_params: OAVParameters,
 ):
+    if params.demand_energy_ev:
+        yield from bps.abs_set(
+            composite.undulator, params.demand_energy_ev / 1000, wait=True
+        )
     motor_time_to_speed = yield from bps.rd(composite.smargon.omega.acceleration_time)
     max_vel = yield from bps.rd(composite.smargon.omega.max_velocity)
     motion_values = calculate_motion_profile(params, motor_time_to_speed, max_vel)
--- a/tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
@@ -702,3 +702,17 @@ def test_rotation_scan_correctly_triggers_zocalo_callback(
             ),
         )
     mock_zocalo_interactor.return_value.run_start.assert_called_once()
+
+
+def test_rotation_scan_set_undulator_gap_after_xbpm_stable(
+    rotation_scan_simulated_messages,
+    test_rotation_params: RotationScan,
+):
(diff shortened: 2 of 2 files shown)


DominicOram (changes requested):
If we're concerned about it coming back after a beam dump then we should probably put the check in `transmission_and_xbpm_feedback_for_collection_wrapper` as this is what is always run straight after a beam dump

DominicOram (changes requested):
Sorry to be a pain and ask you to refactor this again. I think a cleaner way to do this would be to read the energy from the DCM and update the Undulator to match. @npat113, I assume this is ok? This has the advantage of:
* We don't need to pass the `energy_ev` around
* It is valid for plans to be called with `energy_ev = None`, this means "Just use the current energy". With the proposed solution if the undulator doesn't come back in this case we will not correct it

src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -307,6 +307,10 @@ def _move_and_rotation(
     params: RotationScan,
     oav_params: OAVParameters,
 ):
+    if params.demand_energy_ev:
```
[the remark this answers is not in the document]
shihab-dls: The comment should definitely be there for context; as for the logging message, seems that
`if difference > tolerance`
A subsequent logging message `"Gap is already in the correct place for the new energy value f"{energy_kev}, no need to ask it to move"` is produced, which I don't think is misleading.

DominicOram (changes requested):
Great, thank you! Final thing (promise) can we get a test in `test_xbpm_feedback` that confirms that the energy is read from the DCM and set to the undulator?

tests/unit_tests/hyperion/device_setup_plans/test_xbpm_feedback.py
```diff
@@ -19,6 +19,39 @@ def fake_undulator_set(undulator, done_status):
...
+        return_value={"value": {"value": energy_in_kev}}
+    )
+
+    @transmission_and_xbpm_feedback_for_collection_decorator(
+        fake_undulator_set, xbpm_feedback, attenuator, dcm, 0.1
+    )
+    def my_collection_plan():
+        yield from bps.null()
+
+    set_mock_value(xbpm_feedback.pos_stable, True)  # type: ignore
```
DominicOram, line 40: Nit: If you set this to 1 it's more like the actual behaviour and you don't need the type ignore

DominicOram (approved):
Minor nit but take it or leave it. Thank you for putting up with so many requested changes! Good work.

tests/unit_tests/hyperion/device_setup_plans/test_xbpm_feedback.py
shihab-dls: That's much nicer; I'll amend this in the other tests as well as they are minor changes

PR #735, opened 2025-01-07
Update MX beamlines to use new device factory - Part 1

rtuck99:
This addresses
* DiamondLightSource/dodal#846

See also dodal PR
* DiamondLightSource/dodal#975

Additionally, this PR also adds unit test duration summary to the end of pytest, this was done because it seems that tests are taking longer, see related issue that is not addressed here
* DiamondLightSource/dodal#973

### DIFF
--- a/tests/unit_tests/hyperion/conftest.py
+++ b/tests/unit_tests/hyperion/conftest.py
@@ -11,7 +11,11 @@ BANNED_PATHS = [Path("/dls"), Path("/dls_sw")]
 
 @pytest.fixture(scope="session")
 def i03_device_factories():
-    return [f for f in collect_factories(i03).values() if hasattr(f, "cache_clear")]
+    return [
+        f
+        for f in collect_factories(i03, include_skipped=True).values()
+        if hasattr(f, "cache_clear")
+    ]
 
 
 @pytest.fixture(scope="function", autouse=True)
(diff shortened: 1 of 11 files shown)


rtuck99:
Need to update beamline for
* https://github.com/DiamondLightSource/dodal/pull/970

tests/unit_tests/hyperion/conftest.py
```diff
@@ -2,10 +2,28 @@
...
+def i03_device_factories():
+    return [
+        f
+        for f in collect_factories(i03, include_skipped=True).values()
+        if hasattr(f, "cache_clear")
+    ]
+
+
+@pytest.fixture(scope="function", autouse=True)
+def clear_device_factory_caches_after_every_test(i03_device_factories):
```
olliesilvester, line 21: Do you think something similar would be useful in dodal?

olliesilvester (approved):
Looks good! Feel free to merge once the dodal change is ready

tests/unit_tests/hyperion/conftest.py
rtuck99: Seems like dodal follows a different pattern for unit test fixtures - rather than have global fixtures using `device_instantiation` seems to mostly be one-off constructions of devices, thus far it doesn't seem to have been necessary.

PR #740, opened 2025-01-10
Let GDA write to ISPyB sample status 

### LINKED ISSUE #604: Make plans that will allow GDA to write sample status
Following https://github.com/DiamondLightSource/mx-bluesky/issues/595 we would like the same behaviour in GDA (see https://jira.diamond.ac.uk/browse/MXGDA-4148). The easiest way to do this will be to leverage the code written in bluesky so that we can call it via BlueAPI.

As we do this using callbacks we will have to make some "dummy" plans for this that trigger the callbacks e.g. something like:

```python 

@run_decorator
def deposit_load_sample():
    @run_decorator()
    def _inner():
        yield from bps.null()
    yield from bps.subscribe(SampleIspybCallback())
    yield from _inner()

@run_decorator
def deposit_sample_error(error_type: Literal["Sample", "Beamline"]):
    @run_decorator()
    def _inner():
        if error_type == "Sample":
             raise WarningException()
        elif error_type == "Beamline":
             raise Exception()
        yield from bps.null()
    yield from bps.subscribe(SampleIspybCallback())
    yield from _inner()
```

## Acceptance Criteria
- There are endpoints in BlueAPI that allow us to set the sample status in ispyb as specified in https://github.com/DiamondLightSource/mx-bluesky/issues/595 without causing any other affects on the beamline

shihab-dls:
Closes #604  
Dummy plans are provided to allow GDA to trigger error or loaded status in `blsamplestatus`.


olliesilvester:
I think you meant to tag https://github.com/DiamondLightSource/mx-bluesky/issues/604 instead in the description?

olliesilvester (dismissed):
I think the `deposit_sample_error` plan you made does correctly do what we want, however, we've also slightly changed the behaviour of the  `sample_handling_callback` which is used in other plans (`load_centre_collect`). This means it will update the sample status to 'loaded' at the end of the full `load_centre_collect` plan.

Perhaps we could remove the `_record_loaded` feature of `SampleHandlingCallback`, and instead subclass `SampleHandlingCallback`, then overload `activity_gated_stop` to include `_record_loaded` + call the super `activity_gated_stop`. Then we'd have a callback which does the right stuff and isn't used by our UDC plans

olliesilvester:
Also pls can you link the gerrit change you made here?

shihab-dls:
Here's the Gerrit [PR](https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/43191)

olliesilvester:
We removed the code for internal callbacks since you opened this, so we need to deal with the callbacks in a more manual way now. 
We need a new way for general `mx-bluesky` plans to have their own callbacks - Hyperion should be able to use these too. I will write a new issue for this

olliesilvester:
This PR is now waiting on https://github.com/DiamondLightSource/mx-bluesky/issues/807

olliesilvester:
Ignore my comments about callbacks - for the plans in this, we can just use [this decorator](https://nsls-ii.github.io/bluesky/generated/bluesky.preprocessors.subs_decorator.html#bluesky.preprocessors.subs_decorator) to handle them

olliesilvester:
If BlueAPI calls this plan, the plan needs to have it  `@bpp.subs_decorator(callback)`

If Hyperion is calling this plan, Hyperion has already subscribed to the callback on instantiation, so we need want to subscribe again, or it'll try to write to sample status twice. I'll sort this out on Monday

olliesilvester:
@shihab-dls and I have both contributed to this PR, so probably needs @DominicOram or @rtuck99 to review now

src/mx_bluesky/common/external_interaction/callbacks/sample_handling/sample_handling_callback.py
```diff
@@ -37,17 +38,24 @@ def activity_gated_stop(self, doc: RunStop) -> RunStop:
...
         return doc
 
-    def _record_exception(self, exception_type: str):
-        expeye = ExpeyeInteraction()
+    def _record_exception(self, exception_type: str, expeye: ExpeyeInteraction):
         assert self._sample_id, "Unable to record exception due to no sample ID"
         sample_status = self._decode_sample_status(exception_type)
         expeye.update_sample_status(self._sample_id, sample_status)
 
+    def _record_loaded(self, expeye: ExpeyeInteraction):
```
rtuck99, line 55: I feel a bit uncomfortable about having the `SampleHandlingCallback` write the LOADED sample status on successful exit, I think it would be better if that were actually done by the `RobotLoadISPyBCallback`. My reasoning is that if we later decide we want to write other sample status values in the middle of a plan, this will overwrite them. In normal operation in hyperion it's the responsibility of the robot load callback to write the value early on when we do the load, so I think this ought to be something that it does as a special-case.

src/mx_bluesky/common/plans/write_sample_status.py
```diff
@@ -0,0 +1,103 @@
...
+    See https://github.com/DiamondLightSource/mx-bluesky/issues/880"""
+
+    @bpp.run_decorator(
+        md={
+            "metadata": {"sample_id": params.sample_id},
+            "activate_callbacks": ["SampleHandlingCallback"],
+        }
+    )
+    def _inner():
+        if params.exception_type == SampleStatusExceptionType.BEAMLINE:
```
rtuck99, line 53: I think it would be nicer to pass the desired BLSampleStatus value in directly through the run metadata, than to throw an exception in the plan, that way there is a much more direct correspondence between the parameter and the resulting database column value.

src/mx_bluesky/hyperion/experiment_plans/experiment_registry.py
```diff
@@ -5,6 +5,13 @@
 
 import mx_bluesky.hyperion.experiment_plans.flyscan_xray_centre_plan as flyscan_xray_centre_plan
```
rtuck99, line 6: I think you will need to add the hyperion plans to __init__.py, also test cases in `test_when_context_created_then_contains_expected_number_of_plans` in test_main_system

src/mx_bluesky/common/plans/write_sample_status.py
```diff
@@ -0,0 +1,103 @@
...
+            raise AssertionError()
+        elif params.exception_type == SampleStatusExceptionType.SAMPLE:
+            raise SampleException
+        yield from bps.null()
```
rtuck99, line 57: This line is never used

src/mx_bluesky/common/plans/write_sample_status.py
```diff
@@ -0,0 +1,103 @@
...
+
+@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
+class NullDeviceComposite:
+    pass
+
+
+def create_devices(context: BlueskyContext) -> NullDeviceComposite:
+    """Create the necessary devices for the plan."""
+    return device_composite_from_context(context, NullDeviceComposite)
+
```
rtuck99, line 38: We don't seem to have any tests that check whether all our `create_device` methods are invoked. But maybe we should?

src/mx_bluesky/common/plans/write_sample_status.py
olliesilvester: That's true, but we do need a `yield from bps.null()` somewhere otherwise the function doesn't count as a Bluesky plan

src/mx_bluesky/common/external_interaction/callbacks/sample_handling/sample_handling_callback.py
olliesilvester: Good catch, I need to rethink lots of how this works...

We want the two plans to be accessible through GDA via BlueAPI or via Hyperion, but Hyperion shouldn't be subscribed to a callback which says 'set sample loaded on ispyb' every time it sees a stop document.

src/mx_bluesky/common/plans/write_sample_status.py
olliesilvester: We decided to do it this way because then we could reuse the `SampleHandlingCallback`. We could write a new callback to do it this way instead though, I don't really have an opininon. @DominicOram @shihab-dls thoughts?

src/mx_bluesky/common/plans/write_sample_status.py
DominicOram: Nit: Order doesn't matter though, I would just put it so that we do the `bps.null()` first then raise the exceptions. Just to placate CI

src/mx_bluesky/common/external_interaction/callbacks/sample_handling/sample_handling_callback.py
DominicOram: I disagree and think that all writing to sample status should be done in the same Callback. Don't disagree enough to block the PR though. Pushed into https://github.com/DiamondLightSource/mx-bluesky/issues/913

DominicOram (approved):
Looks good to me, some minor comments but not gonna hold this up for them

Issue #318, opened 2023-11-08
Motor tests should assert the setpoint, not the readback


DominicOram:
When trying to convert the `smargon` to `ophyd-async` we struggled to simulate the behaviour of a motor moving it's readback to it's setpoint. This caused a lot of test failures as tests were asserting the motor readback was a given value. These tests should instead assert that the setpoint is that value (as this is really what we want to test anyway)

## Acceptance Criteria
* Motor tests assert the setpoint of the motor, not the readback
* Tests still pass
* We put a note somewhere that this is best practice
* We look at removing the mock motors in the smargon to see if tests still pass

olliesilvester:
Closing this after #742 , but I'm not clear on the part about mock motors. We still need `patch_async_motor`, for various other reasons (like setting the velocity to a non-0 value).

@DominicOram Can you double check you're happy that this issue is done?

DominicOram:
> Closing this after [#742](https://github.com/DiamondLightSource/mx-bluesky/pull/742) , but I'm not clear on the part about mock motors. We still need `patch_async_motor` for various other reasons (like setting the velocity to a non-0 value).
> 
> [@DominicOram](https://github.com/DominicOram) Can you double check you're happy that this issue is done?

Yh, this is fine. Thanks!

PR #742, opened 2025-01-10
Assert setpoint in motor tests

shihab-dls:
This addresses issue #318.

All motor tests that assert `user_readback == ` should be changed to assert `user_setpoint == `. I have some clarifications of changes as comments below, that I would appreciate thoughts on.

### DIFF
--- a/tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
@@ -47,7 +47,6 @@ def mock_pin_tip(pin_tip: PinTipDetection):
 async def test_given_the_pin_tip_is_already_in_view_when_get_tip_into_view_then_tip_returned_and_smargon_not_moved(
     smargon: Smargon, oav: OAV, RE: RunEngine, mock_pin_tip: PinTipDetection
 ):
-    set_mock_value(smargon.x.user_readback, 0)
     set_mock_value(mock_pin_tip.triggered_tip, np.array([100, 200]))
 
     mock_pin_tip.trigger = MagicMock(return_value=NullStatus())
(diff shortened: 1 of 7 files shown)


tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
```diff
@@ -137,8 +134,6 @@ async def test_given_tip_at_zero_but_will_be_found_when_get_tip_into_view_then_s
     )
     set_mock_value(mock_pin_tip.validity_timeout, 0.15)
 
-    set_mock_value(smargon.x.user_readback, 0)
```
shihab-dls, line 140: I think mocking `user_readback = 0` at the start of tests like this is redundant, as the mocked signal defaults to 0 anyways

tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
```diff
@@ -259,7 +259,6 @@ async def test_rotation_plan_smargon_doesnt_move_xyz_if_not_given_in_params(
     assert params.y_start_um is None
     assert params.z_start_um is None
     for motor in [smargon.phi, smargon.chi, smargon.x, smargon.y, smargon.z]:
-        assert await motor.user_readback.get_value() == 0
```
shihab-dls, line 262: Given we're asserting the setpoint now, the alternative `motor.user_setpoint.get_value() == 0` seems redundant, as we next assert `get_mock_put(motor.user_setpoint).assert_not_called()`, wherein `user_setpoint.get_value()` could not have changed from its default `0`

olliesilvester (approved):
Looks good. Not 100% on what some of the criteria in the issue is referring too, but I will chase up with Dom at some point

PR #746, opened 2025-01-14
Correctly pass parameters through all of Hyperion's UDC parameters

### LINKED ISSUE #738: GPU processing not working with gda.mx.hyperion.xrc.compare_cpu_and_gpu=true
During testing of mx-bluesky 1.4.1 (see #697) it was discovered that although `gda.mx.hyperion.xrc.compare_cpu_and_gpu=true` was enabled in `domain.properties`, GPU gridscan processing was not occurring because `dev_shm` wasn't being enabled; this was causing a long wait for the GPU zocalo results which never arrived until the wait times out. The remainder of testing had to proceed with the GPU setting disabled.

## Acceptance Criteria
- The above flag works correctly with GPU processing.

olliesilvester:
- Fixes #738 by giving our Hyperion parameters the `WithHyperionUDCFeatures` component and making sure Hyperion's different parameter models implement the correct version of `detector_parameters`

- Removes the definition of `detector_parameters` from the model components in `common`.  This:
  1. Forces top level-parameter models to implement `detector_parameters`, since the common components are abstract without their own definitions. This part was discussed with @rtuck99 
  2. Allows the Hyperion plans which run prior to `grid_detect` to contain their own definition of `detector_params`, rather than using the one which was defined in `GridCommon`

- Clearer distinction between Hyperion's plans which run before and after `grid_detect` (and so can have `SpecifiedGrid` in their parameters)


olliesilvester:
Also see https://github.com/DiamondLightSource/mx-bluesky/issues/747 which ideally should be fixed before next release

src/mx_bluesky/hyperion/parameters/robot_load.py
```diff
@@ -0,0 +1,23 @@
...
+from mx_bluesky.common.parameters.robot_load import (
+    RobotLoadAndEnergyChange,
+    RobotLoadThenCentre,
+)
+from mx_bluesky.hyperion.parameters.components import WithHyperionUDCFeatures
+from mx_bluesky.hyperion.parameters.gridscan import HyperionPinTipCentreThenXrayCentre
+
+
+class HyperionRobotLoadThenCentre(RobotLoadThenCentre, WithHyperionUDCFeatures):
+    thawing_time: float = Field(default=HardwareConstants.THAWING_TIME)
```
rtuck99, line 13: `thawing_time` is also in `RobotLoadThenCentre`
`tip_offset_um` is also in `GridCommon`

src/mx_bluesky/hyperion/parameters/gridscan.py
```diff
@@ -103,3 +98,13 @@ def panda_FGS_params(self) -> PandAGridScanParams:
...
 
 class OddYStepsException(Exception): ...
+
+
+class HyperionPinTipCentreThenXrayCentre(
+    PinTipCentreThenXrayCentre, WithHyperionUDCFeatures
+): ...
+
+
+class HyperionGridScanWithEdgeDetect(
```
rtuck99, line 108: This doesn't seem to be used anywhere?

src/mx_bluesky/hyperion/parameters/robot_load.py
```diff
@@ -0,0 +1,23 @@
...
+    tip_offset_um: float = Field(default=HardwareConstants.TIP_OFFSET_UM)
+
+    def robot_load_params(self):
```
rtuck99, line 16: Can these methods have annotations? That way tooling can help more.

rtuck99: Should these be properties?

rtuck99 (dismissed):
Approved with above changes

olliesilvester:
I will address these first thing tomorrow

Also, in `robot_load_then_xray_centre`, we create the detector params
 ```
detector_params = yield from fill_in_energy_if_not_supplied(
        composite.dcm, parameters.detector_params
    )
```
This uses the `GridCommon` definition of `DetectorParams`, which doesn't allow us to set `dev_shm`. I will update this PR to also fix that + a test to prove it

File: src/mx_bluesky/hyperion/parameters/robot_load.py (hunk 3 above)
Comment (olliesilvester) [reply src/mx_bluesky/hyperion/parameters/robot_load.py:16]: Yes to annotations!

I'll make them properties as that's probably more consistent with the other params

File: src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py [hunk 4]
```diff
@@ -101,7 +106,7 @@ def activity_gated_start(self, doc: RunStart):
             )
             mx_bluesky_parameters = doc.get("mx_bluesky_parameters")
             assert isinstance(mx_bluesky_parameters, str)
-            self.params = GridCommon.model_validate_json(mx_bluesky_parameters)
+            self.params = self.param_type.model_validate_json(mx_bluesky_parameters)
```
olliesilvester, line 109: Because `GridCommon` is now abstract, we need to use an implemented version of this model here. Since the type is bound to `GridCommon`, I don't think this will affect generalisability.

Similar change done in `nexus_callback.py`

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -23,86 +18,37 @@
...
-            num_images_per_trigger=1,
-            num_triggers=self.num_images,
-            use_roi_mode=self.use_roi_mode,
-            det_dist_to_beam_converter_path=self.det_dist_to_beam_converter_path,
-            trigger_mode=self.trigger_mode,
-            **optional_args,
-        )
-
-
-class RobotLoadThenCentre(GridCommon):
```
olliesilvester, line 77: Discussed with @rtuck99 - we're so far off any non-hyperion beamlines doing anything other than a grid scan, it's probably good to get rid of some of the common parameters like this one

pyproject.toml
```diff
@@ -45,7 +45,7 @@ dependencies = [
     "daq-config-server >= 0.1.1",
     "ophyd == 1.9.0",
     "ophyd-async >= 0.8a5",
-    "bluesky >= 1.13.0a4",
+    "bluesky >= 1.13",
```
olliesilvester, line 48: Alpha releases were causing linting issues with `set_mock_value`

rtuck99 (approved):
Approved

PR #748, opened 2025-01-15
Dodal 846 convert i24 to use device factory

rtuck99:
This PR updates mx-bluesky to work with dodal changes in

* DiamondLightSource/dodal#990

which addresses

* DiamondLightSource/dodal#846

for i24

This PR builds on top of

* #735 
* ~#743~


noemifrisina (approved):
Looks good, tests are happy, thanks.

Codecov seems to be complaining, but it's just the patch and looks to be outside of the scope of this PR so we may have to live with it for the moment.

Issue #533, opened 2024-09-26
 Hyperion unit tests failing on i24 


noemifrisina:
When running tests in the environment in `/dls_sw/i24/software/...` , several hyperion unit tests seem to be failing, eg:

```
FAILED tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py::TestFlyscanXrayCentrePlan::test_flyscan_xray_centre_sets_directory_stages_arms_disarms_unstages_the_panda - AssertionError: Attempt to open /dls_sw/i24/software/bluesky/mx-bluesky_v1.1.0/mx-bluesky/src/mx_bluesky/hyperion/resources/panda/panda-gridscan.yaml from inside a unit test
assert not True
 +  where True = is_relative_to(PosixPath('/dls_sw'))
 +    where is_relative_to = PosixPath('/dls_sw/i24/software/bluesky/mx-bluesky_v1.1.0/mx-bluesky/src/mx_bluesky/hyperion/resources/panda/panda-gridscan.yaml').is_relative_to
```
with stack trace:
```
_ TestFlyscanXrayCentrePlan.test_given_setting_stub_offsets_disabled_then_stub_offsets_not_set[panda] _
Traceback (most recent call last):
  File "/dls_sw/i24/software/bluesky/mx-bluesky_v1.1.0/mx-bluesky/.venv/lib/python3.11/site-packages/_pytest/runner.py", line 341, in from_call
    result: TResult | None = func()
                             ^^^^^^
  File "/dls_sw/i24/software/bluesky/mx-bluesky_v1.1.0/mx-bluesky/.venv/lib/python3.11/site-packages/_pytest/runner.py", line 242, in <lambda>
    lambda: runtest_hook(item=item, **kwds), when=when, reraise=reraise
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/dls_sw/i24/software/bluesky/mx-bluesky_v1.1.0/mx-bluesky/.venv/lib/python3.11/site-packages/pluggy/_hooks.py", line 513, in __call__
    return self._hookexec(self.name, self._hookimpls.copy(), kwargs, firstresult)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #749, opened 2025-01-15
Fix unit test banned path check to allow access to project folder

rtuck99:
Fixes
* #533 

This change adjusts the check for unit tests trying to open files so that files in the project folder are allowed

### DIFF
--- a/tests/unit_tests/hyperion/conftest.py
+++ b/tests/unit_tests/hyperion/conftest.py
@@ -4,7 +4,7 @@ from unittest.mock import patch
 
 import pytest
 
-BANNED_PATHS = [Path("/dls"), Path("/dls_sw")]
+PERMITTED_PATHS = [Path("/tmp")]
 
 
 @pytest.fixture(autouse=True)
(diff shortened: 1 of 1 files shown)


DominicOram (approved):
Looks good but as a suggestion how about we make it a whitelist instead? It looks like the only places the tests are writing to are:
* The package folder (which we're now checking)
* `/tmp`

So we could just check for one of those rather than check against the banned?

rtuck99:
> Looks good but as a suggestion how about we make it a whitelist instead? It looks like the only places the tests are writing to are:
> 
>     * The package folder (which we're now checking)
> 
>     * `/tmp`
> 
> 
> So we could just check for one of those rather than check against the banned?

I tried this, however it seems to run into problems, somehow running the test some packages using `importlib.resources.files` seem to resolve files from the workspace, others resolve to `site-packages`. I can't reproduce this on my local system.

However I think in any case to handle this properly we ought to whitelist files from every conceivable folder that python might resolve packages from, esp if 3rd party code ever loads any resources. Unfortunately this might be difficult to cover not only CI but also dev machines what with system packages, virtual environments etc.

DominicOram:
> > Looks good but as a suggestion how about we make it a whitelist instead? It looks like the only places the tests are writing to are:
> > ```
> > * The package folder (which we're now checking)
> > 
> > * `/tmp`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So we could just check for one of those rather than check against the banned?
> 
> I tried this, however it seems to run into problems, somehow running the test some packages using `importlib.resources.files` seem to resolve files from the workspace, others resolve to `site-packages`. I can't reproduce this on my local system.
> 
> However I think in any case to handle this properly we ought to whitelist files from every conceivable folder that python might resolve packages from, esp if 3rd party code ever loads any resources. Unfortunately this might be difficult to cover not only CI but also dev machines what with system packages, virtual environments etc.

Sure, happy to keep as is

PR #755, opened 2025-01-20
Remove unnecessary pydantic model_config arbitrary_types_allowed=True

rtuck99:
This removes the `arbitrary_types_allowed=True` fro `MxBlueskyParameters.model_config`


olliesilvester (approved):
Nothing broke so probably fine!

PR #761, opened 2025-01-24
Additional system tests for zocalo results bounds checks

rtuck99:
Additional system tests accompanying 
* DiamondLightSource/dodal#1004


DominicOram:
Tried to run this but getting an error of `bluesky.utils.FailedStatus: <AsyncStatus, device: undulator_dcm, task: <coroutine object UndulatorDCM.set at 0x7fcb682cf2e0>, errored: FileNotFoundError(2, 'No such file or directory')>` I think @rtuck99 is looking at this as part of another issue.

tests/system_tests/hyperion/external_interaction/conftest.py
```diff
@@ -255,13 +301,33 @@ async def zocalo_for_fake_zocalo():
...
 
-    with patch.object(zocalo, "trigger", side_effect=mock_zocalo_complete):
-        yield zocalo
+    with (
+        patch("dodal.devices.zocalo.zocalo_results.workflows") as workflows,
+        patch("dodal.devices.zocalo.zocalo_results._get_zocalo_connection"),
+    ):
+        workflows.recipe.wrap_subscribe.side_effect = mock_worfklow_subscribe
+        with patch.object(zocalo, "trigger", side_effect=mock_zocalo_complete):
+            yield zocalo
```
DominicOram, line 330: Should: I feel like this is overly complicated. How about something like:
```python
@pytest.fixture
def zocalo_for_system_test(zocalo) -> Generator[ZocaloResults, None, None]:
    with _zocalo_that_returns_result(zocalo, TEST_RESULT_MEDIUM):
        yield zocalo


@contextmanager
def _zocalo_that_returns_result(zocalo, my_result):
    @AsyncStatus.wrap
    async def mock_zocalo_complete():
        await zocalo._put_results(my_result, {"dcid": 1234, "dcgid": 123})

    with patch.object(zocalo, "trigger", side_effect=mock_zocalo_complete):
        yield zocalo
```
then in the test:
```python
with _zocalo_that_returns_result(load_centre_collect_composite.zocalo, zocalo_result):
   ...
```

DominicOram (changes requested):
Great, thanks, minor suggestions

tests/system_tests/hyperion/external_interaction/conftest.py
rtuck99: I think we already have some tests elsewhere that do this. However that then means that the code that is responsible for decoding the message from zocalo and setting the soft signals is then not tested. 
For a system test, it seemed more appropriate to test against the external interface with the zocalo message queue. This should have the added benefit of being more stable with respect to changes in our own code.

DominicOram: > However that then means that the code that is responsible for decoding the message from zocalo and setting the soft signals is then not tested.

We have tests for this in `dodal` using the fake zocalo. It also feels like it should be more of `dodal`'s concern anyway and `mx_bluesky` can assume that the internals of the device work as expected.

But sure, I agree a broader system test would be good. In which case should we be using the fake zocalo here? That will actually test the whole pipeline. I just feel like this test is fragile because it adds `my_callback` and `my_zocalo_result`, which don't exist on the production code. We then have to use `my_zocalo_result` in the test itself, where it's not clear that exists without reference back to this fixture.

rtuck99: I'm not sure how much extra adding the real zocalo is going to bring. After all as far as hyperion is concerned it's just a thing that sits on the message bus and reads the database. As long as the messages that get sent back and forth are realistic and sent at the same time it shouldn't matter. I doubt it's something that's likely to change without us knowing. Real zocalo does seem to have quite a lot of configuration and would also require a bunch of actual test image data to process (which would have to be simulated in a way that gives an actual result) so it's not trivial.

I suppose if someone did decide to use`my_callback` as an attribute name, or it did some funky introspection then there would be a problem. But it's really just a convenient way of keeping track of handlers, we could prefix it with something else to make collisions less likely, or move to some other method like having a global variable (ugh). At the end of the day python objects are just glorified hashmaps. Besides, it's not like we don't already have lots of unit tests that call private functions (or functions that should be private) in our code.

DominicOram: Sure, I still don't like it but willing to concede based on it being just test code.

DominicOram (approved):
Great thank you!

Issue #762, opened 2025-01-24
box_size_um cannot be passed from GDA in hyperion requests


rtuck99:
Currently the `box_size_um` parameter cannot be specified in request parameters to hyperion because it is only present in the `GridScanWithEdgeDetect` class which is not inherited by anything in the request parameter class hierarchy.

see e.g.

```
[31m[E 250122 13:57:00.429 __main__:267] ['Traceback (most recent call last):\n', '  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.3/mx-bluesky/src/mx_bluesky/hyperion/__main__.py", line 241, in compose_start_args\n    raise ValueError(f"Extra fields not allowed {parameters.model_extra}")\n', "ValueError: Extra fields not allowed {'box_size_um': 20.0}\n", '\nThe above exception was the direct cause of the following exception:\n\n', 'Traceback (most recent call last):\n', '  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.3/mx-bluesky/src/mx_bluesky/hyperion/__main__.py", line 259, in put\n    plan, params, plan_name, callback_type = compose_start_args(\n                                             ^^^^^^^^^^^^^^^^^^^\n', '  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.4.3/mx-bluesky/src/mx_bluesky/hyperion/__main__.py", line 243, in compose_start_args\n    raise ValueError(\n', 'ValueError: Supplied parameters don\'t match the plan for this endpoint b\'{"storage_directory": "/dls/i03/data/2025/cm40607-1/xraycentring/TestThaumatin/thau_4", "sample_puck": "1", "exposure_time_s": 0.004, "file_name": "thau_4", "sample_id": 6197098, "box_size_um": 20.0, "omega_start_deg": 0, "sample_pin": "4", "use_roi_mode": true, "chi_start_deg": 0.0, "features": {"compare_cpu_and_gpu_zocalo": true}, "detector_distance_mm": 196.6079939921457, "demand_energy_ev": null, "visit": "cm40607-1", "transmission_frac": 1.0, "parameter_model_version": "5.0.0"}\', for plan robot_load_then_centre\n'][0m
```

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/763

PR #763, opened 2025-01-24
Fix for box_size_um not included in robot_load_then_centre_plan.

rtuck99:
Fixes

* #762 

Depends on

* DiamondLightSource/dodal#997

`box_size_um` should no longer be rejected when specified in robot_load_then_centre


tests/unit_tests/hyperion/test_main_system.py
```diff
@@ -290,14 +292,53 @@ def test_when_started_n_returnstatus_interrupted_bc_RE_aborted_thn_error_reptd(
...
+    "endpoint, test_file",
+    [
+        [
+            START_ENDPOINT,
+            "tests/test_data/parameter_json_files/good_test_parameters.json",
+        ],
+        [
+            "/grid_detect_then_xray_centre/start",
+            "tests/test_data/parameter_json_files/good_test_grid_with_edge_detect_parameters"
+            ".json",
```
DominicOram, line 305: Nit: Having the `.json` on a separate line made me think it was a new arg at first glance. I think it's more readable together, even if it is longer than the line length

DominicOram (approved):
Great, thanks. Just some minor nit comments. I made https://github.com/DiamondLightSource/mx-bluesky/pull/768 to address them but you can just close that off if you like

Issue #764, opened 2025-01-24
Deployment issues discovered during mx-bluesky 1.4.3 testing


rtuck99:
This tracks deployment issues found in

* #758 

Specifically:

* `deploy_hyperion_to_k8s.sh` can't determine the correct deployment directory because `deploy_mx_bluesky.py --print-release-dir` prints an additional line 
* `deploy_hyperion_to_k8s.sh` can't determine the app version automatically because `setuptools_scm` was run and printed some additional output which got expanded into `$APP_VERSION`
* The `Dockerfile` is missing additional libGL dependencies required by cv2

## Acceptance Criteria
- `deploy_hyperion_to_k8s.sh` successfully deploys a fresh release of hyperion to kubernetes
- The deployed image successfully starts

PR #765, opened 2025-01-24
Fix deployment scripts

rtuck99:
This addresses
* #764

### DIFF
--- a/utility_scripts/deploy/deploy_mx_bluesky.py
+++ b/utility_scripts/deploy/deploy_mx_bluesky.py
@@ -356,5 +356,4 @@ def _parse_options() -> tuple[str, Options]:
 
 if __name__ == "__main__":
     beamline, options = _parse_options()
-    print(options.prune_deployments)
     main(beamline, options)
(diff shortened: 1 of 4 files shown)


utility_scripts/deploy/deploy_mx_bluesky.py
```diff
@@ -356,5 +356,4 @@ def _parse_options() -> tuple[str, Options]:
 
 if __name__ == "__main__":
     beamline, options = _parse_options()
```
olliesilvester, line 358: > deploy_hyperion_to_k8s.sh can't determine the correct deployment directory because deploy_mx_bluesky.py --print-release-dir prints an additional line

We should add a comment here explaining that we rely on the output here being in a certain format

olliesilvester (approved):

utility_scripts/deploy/deploy_mx_bluesky.py
rtuck99: I think it's adequately covered in the contract of `--print-release-dir` as documented:
"Print the path to the release folder and then exit."

Besides, why would you print anything at all between these two lines?

Issue #758, opened 2025-01-22
Testing schedule, mx-bluesky v1.4.3


rtuck99:
Testing schedule:

GDA changes:
* DiamondLightSource/mx-bluesky#703 - screen and collect changes https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/43072
DONE

Deployment:
* DiamondLightSource/mx-bluesky#271 - deployment script should delete ./tmp - N/A deployment fails before tests are run due to RH7 control server `confluent-kafka` build failure
* DiamondLightSource/mx-bluesky#242 
   - deployment script should give console output when running dls_dev_env.sh - DONE
   - deploy_hyperion_to_k8s.sh should now work - FAIL
   - the container image starts - FAIL
   - see #764 

Robot load:
* DiamondLightSource/mx-bluesky#698 - Hyperion should check the beamstop is in position, robot load after a robot error/puck load with beamstop out should produce an error

Gridscans:

* DiamondLightSource/mx-bluesky#242 - non-diffracting samples should now return no hits and be skipped
	       - chi is always applied when specified as 0 after previously being specified as non-zero
* DiamondLightSource/mx-bluesky#738 - GPU processing works with gda.mx.hyperion.xrc.compare_cpu_and_gpu=true in domain.properties - FIXED
* DiamondLightSource/mx-bluesky#221 - Crystals with x-coordinates near 0 on the end of the pin should not cause {grid_position} is outside the bounds of the grid error messages - FAIL
   * DiamondLightSource/mx-bluesky#759

Rotations:
* DiamondLightSource/mx-bluesky#618 - Aperture selection based on bounding box size - test large >= 50um and small crystals 
- confirmed for large crystals.
* DiamondLightSource/mx-bluesky#247 - Test with omega_flip = True feature flag, rotation_direction Positive in request. This should produce identical results to GDA - FIXED

Single collect:
* DiamondLightSource/mx-bluesky#687 - test samples with bad pin/no diffraction, single collection
* DiamondLightSource/mx-bluesky#687 - test samples with bad pin/no diffraction, multi-rotation
Screen and collect:
* DiamondLightSource/mx-bluesky#703 - screen and collect should actually collect as well as screen
* DiamondLightSource/mx-bluesky#687 - test samples with bad pin/no diffraction, S+C

Acceptance tests:

Neil to soak test energy changes, screen + collect

rtuck99:
> deploy_hyperion_to_k8s.sh should now work

[...]

Issue #516, opened 2024-09-20
 callbacks_factory in experiment_registry does not appear to ever be used


rtuck99:
`callbacks_factory` is a required member of `ExperimentRegistryEntry` but doesn't appear to ever be used.
`__main__.py.compose_start_args()`  seems to instead reference `callback_collection_type` which doesn't exist.
`wait_on_queue()` only looks at the callbacks if external callbacks are not used, which I believe is never the case, instead the callbacks are registered  in `setup_callbacks()` in `callbacks/__main__.py`

## Acceptance Criteria
- We either decide to use or get rid of `callback_collection_type` and associated methods in `callback_util.py`

DominicOram:
As part of this I think we should remove the flag to stop using external callbacks, we always use external ones now to the point where I suspect they won't work if we tried to use the internal ones again.

PR #766, opened 2025-01-24
Fix zocalo callback error

DominicOram:
See https://github.com/DiamondLightSource/mx-bluesky/issues/758#issuecomment-2612821684

This is a bit of a hotfix that avoids the race condition by making sure the zocalo callback is set up on initialisation rather than setting it up in the plan.

https://github.com/DiamondLightSource/mx-bluesky/issues/767 is to look at a long term fix but I don't think this hotfix is too bad anyway.

## To test
* Confirm tests still pass
* Confirm which plans trigger zocalo for rotation/grid are unchanged


olliesilvester (approved):
I'm happy with this, looks like a good solution

src/mx_bluesky/hyperion/external_interaction/callbacks/__main__.py
```diff
@@ -49,7 +49,7 @@
 
 
 def setup_callbacks():
-    zocalo = ZocaloCallback()
+    zocalo = ZocaloCallback("", "")
```
olliesilvester, line 52: Don't the `emit` parts on the lines below get overwritten whenever we use the callbacks?

DominicOram: Good spot, this is a mess... https://github.com/DiamondLightSource/mx-bluesky/issues/516. I think only this callback is used and not the others

PR #770, opened 2025-01-27
Use ophyd-async 0.9a2

### LINKED ISSUE #706: Update panda grid scans for new ophyd-async save-load
Ophyd-async v1 will use the new save-load design described in https://github.com/bluesky/ophyd-async/issues/16. We should update our panda grid scan accordingly

## Acceptance Criteria
- Panda setup plan updated to work with changes in ophyd-async

olliesilvester:
Fixes #706 

- Uses new save-load
- `DeviceCollector` renamed to `init_devices`

Requires https://github.com/DiamondLightSource/dodal/pull/1014

### DIFF
--- a/src/mx_bluesky/definitions.py
+++ b/src/mx_bluesky/definitions.py
@@ -1,0 +1,3 @@
+import os
+
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
--- a/src/mx_bluesky/definitions.py
+++ b/src/mx_bluesky/definitions.py
@@ -1,3 +1,4 @@
 import os
 
+# Change once Python<3.12 is dropped - see https://github.com/DiamondLightSource/mx-bluesky/issues/798
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
--- a/src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
@@ -1,6 +1,5 @@
 from datetime import datetime
 from enum import Enum
-from importlib import resources
 from pathlib import Path
 
 import bluesky.plan_stubs as bps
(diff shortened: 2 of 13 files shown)


src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -145,10 +144,11 @@ def setup_panda_for_flyscan(
 
     yield from bps.stage(panda, group="panda-config")
 
-    with resources.as_file(
-        resources.files(panda_resource) / "panda-gridscan.yaml"
-    ) as config_yaml_path:
-        yield from load_device(panda, str(config_yaml_path))
+    yield from load_panda_from_yaml(
```
rtuck99, line 147: What was the motivation from moving from `importlib.resources` to passing around paths?

tests/unit_tests/hyperion/device_setup_plans/test_setup_panda.py
```diff
@@ -209,8 +209,13 @@ def assert_set_table_has_been_waited_on(*args, **kwargs):
...
         ),
-        patch("mx_bluesky.hyperion.device_setup_plans.setup_panda.load_device"),
+        patch(
+            "mx_bluesky.hyperion.device_setup_plans.setup_panda.load_panda_from_yaml"
+        ),
         patch("mx_bluesky.hyperion.device_setup_plans.setup_panda.bps.abs_set"),
+        patch(
+            "mx_bluesky.hyperion.device_setup_plans.setup_panda.load_panda_from_yaml"
+        ),
     ):
```
rtuck99, line 219: We've patched this twice?

src/mx_bluesky/definitions.py
```diff
@@ -0,0 +1,3 @@
+import os
+
+ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
```
rtuck99, line 3: For things which we know are resources contained within our python project, I think I prefer to use `importlib.resources` to locate them rather than to use file paths. As far as I can tell it seems to be the "proper" way.

Python doesn't guarantee that every module has a `__file__` attribute since modules aren't always in files. Having said that I suspect that our packages will probably always be available as files.

I think most of my thoughts in this area derive from analogy with Java ClassLoaders which are a similar concept to python module, and it's best practice to bundle your resources in your jar file (usually in a separate hierarchy), and I think that is the overall intent of things like `importlib.resources` to encourage this, so that python packaging tools also know how to deploy your resources.

Having said all this

* `importlib.resources` does seem to be annoyingly immature as an API
* there's no functional imperative to use this mechanism at the moment

so I leave this here as an opinion you can choose to follow or ignore...

rtuck99 (approved):
Approved

src/mx_bluesky/definitions.py
olliesilvester: Discussed in person - as of Python 3.12, we can use this API to get a directory traversible, but we can't use it just yet. I will leave the code as-is but add a comment + issue to use `resources` again once we drop python < 3.12

olliesilvester: See https://github.com/DiamondLightSource/mx-bluesky/issues/798

PR #772, opened 2025-01-27
771 use gpu results

### LINKED ISSUE #771: Toggle to use GPU results over CPU ones
We are now somewhat confident that we can use the GPU results over the CPU ones. We should use them.

## Acceptance Criteria
- There is a property in GDA that, when on will take the GPU results from zocalo and not wait for the CPU ones to come in

DominicOram:
Fixes #771 

Needs https://github.com/DiamondLightSource/dodal/pull/1025

To Test
* Confirm tests pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd == 1.9.0",
     "ophyd-async >= 0.8a5",
     "bluesky >= 1.13",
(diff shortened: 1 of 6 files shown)


olliesilvester (changes requested):
Can we add a validator which doesn't let both `use_gpu_and_cpu` AND `use_gpu` be true? It doesn't make sense to use both and will make the dodal device do weird things

olliesilvester (approved):
Looks good!

PR #774, opened 2025-01-28
I24 serial: beamline test fixes

noemifrisina:
Testing version 1.4.4


pyproject.toml
```diff
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd == 1.9.0",
     "ophyd-async >= 0.8a5",
     "bluesky >= 1.13",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@cb12746cb3f658a6d4571a0cd870133a17980bb7",
+    "dls-dodal == 1.39.0",
```
DominicOram, line 49: Should: Better not to have the pin in that I did for the release i.e. not to include that commit.

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -626,6 +629,8 @@ def kickoff_and_complete_collection(pmac: PMAC, parameters: FixedTargetParameter
         pmac.collection_time, total_collection_time, group="setup_pmac"
     )
     yield from bps.wait(group="setup_pmac")  # Make sure the soft signals are set
+    _sig = yield from bps.rd(pmac.collection_time)
+    SSX_LOGGER.warning(f"This was set for collection time {_sig}")
```
DominicOram, line 633: Should: I suspect this may have been left in after debugging? In theory it should just be the same as what's set and logged above?

src/mx_bluesky/beamlines/i24/serial/write_nexus.py
```diff
@@ -54,6 +54,7 @@ def call_nexgen(
             pump_status = parameters.pump_status
 
     filename_prefix = parameters.filename
+    filename_prefix = cagetstring(Eiger.pv.filenameRBV)
```
DominicOram, line 57: Should: We should remove the first set of `filename_prefix` if we're immediately going to overwrite it

src/mx_bluesky/beamlines/i24/serial/dcid.py
```diff
@@ -213,7 +213,9 @@ def generate_dcid(
                 "transmission": transmission,
                 "visit": self.parameters.visit.name,
                 "wavelength": beam_settings.wavelength_in_a,
-                "group": {"experimentType": self.parameters.ispyb_experiment_type},
+                "group": {
+                    "experimentType": self.parameters.ispyb_experiment_type.value
+                },
```
DominicOram, line 218: Should: We should add a test that would have caught this. There are already tests that patch out the `request.post` below but they just assert it's called, we should assert it's called with reasonable values

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -707,7 +712,7 @@ def run_fixed_target_plan(
     parameters.collection_directory.mkdir(parents=True, exist_ok=True)
 
     if parameters.chip_map:
-        upload_chip_map_to_geobrick(pmac, parameters.chip_map)
+        yield from upload_chip_map_to_geobrick(pmac, parameters.chip_map)
```
DominicOram, line 715: Should: We should have a test that would have caught this

src/mx_bluesky/beamlines/i24/serial/write_nexus.py
DominicOram: Could: A test for this would be nice

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
```diff
@@ -433,7 +435,8 @@ def start_i24(
 
         # DCID process depends on detector PVs being set up already
         SSX_LOGGER.debug("Start DCID process")
-        filetemplate = f"{parameters.filename}.nxs"
+        complete_filename = cagetstring(pv.eiger_ODfilenameRBV)
+        filetemplate = f"{complete_filename}.nxs"
```
DominicOram, line 439: Could: A test for this would be nice

DominicOram (changes requested):
Great, thanks. Couple of comments that mostly boil down to tidying up and adding tests

pyproject.toml
noemifrisina: Sigh, I didn't even notice this had made it into the PR. Good spot, thanks!

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Collect_py3v1.py
noemifrisina: Yep, it was just to check that it had been correctly set

src/mx_bluesky/beamlines/i24/serial/write_nexus.py
noemifrisina: Facepalm, I though I had removed it, thanks. I'm actually wondering if it would be worth it to pass as an argument since we have to do the same cagetstring before starting dcid in the collection plan...

tests/unit_tests/beamlines/i24/serial/fixed_target/test_chip_manager.py
```diff
@@ -261,6 +263,47 @@ def test_scrape_mtr_directions():
...
 
 
+@patch(
+    "mx_bluesky.beamlines.i24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.scrape_mtr_directions"
+)
+@patch("mx_bluesky.beamlines.i24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.bps.rd")
+@patch(
+    "mx_bluesky.beamlines.i24.serial.fixed_target.i24ssx_Chip_Manager_py3v1.PARAM_FILE_PATH_FT"
+)
+def test_fiducial(fake_param_path, patch_read, patch_mtr, pmac, RE):
```
DominicOram, line 273: Could: This isn't a great test name. Test fiducial does what?

DominicOram (approved):
Great, thanks. Just one minor comment

PR #776, opened 2025-01-28
Remove ability to use internal callbacks

### LINKED ISSUE #516: callbacks_factory in experiment_registry does not appear to ever be used

### LINKED ISSUE #615: Update copier template
We should move to the latest copier template

## Acceptance Criteria
- We're on the latest copier template

DominicOram:
Fixes #516 

Whilst this goes a bit further than the original issue in that it removes functionality I think this is fine as it's functionality that hasn't been used in a while and so unsure if it actually even works. All it was doing was serving to confuse us more than anything else. If we want the functionality back we can reintroduce it at a later date.

## Acceptance Criteria
* Tests still pass (it's telling how little we had to change the tests)
* All the code around using external callbacks is removed


rtuck99:
I think we also need to update launcher script(s)
* `run_hyperion.sh`
* `entrypoint.sh` for docker

tests/unit_tests/hyperion/test_main_system.py
```diff
@@ -388,81 +374,6 @@ def test_cli_args_parse(arg_list, parsed_arg_values):
     assert test_args.dev_mode == parsed_arg_values[0]
     assert test_args.verbose_event_logging == parsed_arg_values[1]
     assert test_args.skip_startup_connection == parsed_arg_values[2]
-    assert test_args.use_external_callbacks == parsed_arg_values[3]
```
rtuck99, line 391: There's another test above this one that also uses `--external-callbacks`

src/mx_bluesky/hyperion/parameters/cli.py
```diff
@@ -64,5 +63,4 @@ def parse_cli_args() -> HyperionArgs:
         verbose_event_logging=args.verbose_event_logging or False,
```
rtuck99, line 63: `parse_cli_args` also needs changing to remove the arguments from the help

olliesilvester (changes requested):
Thanks, this looks good after addressing Rob's changes.

Non-Hyperion beamlines won't be using external callbacks initially, since they care less about speed and it's not currently supported by BlueAPI, but I will make sure that Hyperion's external callbacks work with the common gridscan

run_hyperion.sh
```diff
@@ -120,10 +116,8 @@ if [[ $START == 1 ]]; then
     #Add future arguments here
     declare -A h_only_args=(        ["SKIP_STARTUP_CONNECTION"]="$SKIP_STARTUP_CONNECTION"
                                     ["VERBOSE_EVENT_LOGGING"]="$VERBOSE_EVENT_LOGGING"
-                                    ["EXTERNAL_CALLBACK_SERVICE"]="$EXTERNAL_CALLBACK_SERVICE" )
```
rtuck99, line 123: missing )

run_hyperion.sh
```diff
@@ -120,10 +116,8 @@ if [[ $START == 1 ]]; then
...
     declare -A h_only_arg_strings=( ["SKIP_STARTUP_CONNECTION"]="--skip-startup-connection"
                                     ["VERBOSE_EVENT_LOGGING"]="--verbose-event-logging"
-                                    ["EXTERNAL_CALLBACK_SERVICE"]="--external-callbacks")
```
rtuck99, line 126: ditto

run_hyperion.sh
olliesilvester: Whoops

src/mx_bluesky/hyperion/parameters/cli.py
```diff
@@ -33,8 +33,7 @@ def parse_cli_args() -> HyperionArgs:
     """Parses all arguments relevant to hyperion. Returns an HyperionArgs dataclass with
     the fields: (verbose_event_logging: bool,
                  dev_mode: bool,
-                 skip_startup_connection: bool,
-                 external_callbacks: bool)"""
+                 skip_startup_connection: bool"""
```
rtuck99, line 36: missing )

utility_scripts/docker/entrypoint.sh
```diff
@@ -47,10 +44,8 @@ callback_start_log_path=$LOG_DIR/callback_start_log.log
 #Add future arguments here
 declare -A h_only_args=(        ["SKIP_STARTUP_CONNECTION"]="$SKIP_STARTUP_CONNECTION"
                                 ["VERBOSE_EVENT_LOGGING"]="$VERBOSE_EVENT_LOGGING"
-                                ["EXTERNAL_CALLBACK_SERVICE"]="$EXTERNAL_CALLBACK_SERVICE" )
 declare -A h_only_arg_strings=( ["SKIP_STARTUP_CONNECTION"]="--skip-startup-connection"
```
rtuck99, line 47: ) !

utility_scripts/docker/entrypoint.sh
```diff
@@ -47,10 +44,8 @@ callback_start_log_path=$LOG_DIR/callback_start_log.log
...
                                 ["VERBOSE_EVENT_LOGGING"]="--verbose-event-logging"
-                                ["EXTERNAL_CALLBACK_SERVICE"]="--external-callbacks")
```
rtuck99, line 53: )

Issue #633, opened 2024-11-11
 Get system tests running on kubernetes


olliesilvester:
Following on from https://github.com/DiamondLightSource/mx-bluesky/issues/632, it would be better to get our system tests running daily on a kubernetes cluster. This would be easier to maintain and update, and we wouldn't have to rely on a workstation in our office

## Acceptance Criteria
- System tests running (and passing) once a day in kubernetes

rtuck99:
Looking into this a bit more, I think the solution looks something like:

* Configure a self-hosted runner (see https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/adding-self-hosted-runners).
   * Create a self-hosted runner container image, probably using a `Dockerfile` similar to https://baccini-al.medium.com/how-to-containerize-a-github-actions-self-hosted-runner-5994cc08b9fb as a starting point.
   * Create a helm deployment that deploys the runner container image, and any other things that are needed by the system tests (fake zocalo? ActiveMQ/RabbitMQ?), and configured so that it can talk to the test ispyb database + test expeye
   * Deploy to k8s
 * Add github workflow .yml file to `mx-bluesky` so that the system tests are run using the self-hosted runner. Initially we might want to be able to run them on demand, and also nightly. Then once we are confident it works maybe we can then have them run as part of CI, but we will have to see how reliably it all runs.

rtuck99:
Blocked waiting on kubernetes namespace creation `mx-bluesky`, which is required to be able to request additional ephemeral-storage which exceeds the user-namespace cap of 2GiB

https://jira.diamond.ac.uk/servicedesk/customer/portal/2/SCHD-7284

PR #793, opened 2025-02-03
System tests on k8s

rtuck99:
This addresses
* #633 

The changes here mainly consist of changes to the unit tests to allow the ISPyB endpoint to be configured via the `ISPYB_CONFIG_PATH` during tests, which was previously overridden.

Also there are changes to allow sample ID, visit and suchlike to be configured as an interim so that it is still possible to run the system tests individually locally against the `ispyb_hyperion_dev.cfg` database.

The actual system test infrastructure can be found in the associated GitLab merge request
* https://gitlab.diamond.ac.uk/MX-GDA/hyperion-system-testing/-/merge_requests/1


pyproject.toml
```diff
@@ -165,7 +166,7 @@ legacy_tox_ini = """
 [tox]
 skipsdist=True
 
-[testenv:{pre-commit,type-checking,tests,docs}]
+[testenv:{pre-commit,type-checking,tests,docs,systemtests}]
```
callumforrester, line 169: Nit: We've been calling it `system-test` in other places, not really bothered but would be nice to be consistent

rtuck99: Unfortunately, dashes have a special meaning within tox:

https://tox.wiki/en/latest/user_guide.html

> Test environment names can consist of alphanumeric characters and dashes; for example: py311-django42. The name will be split on dashes into multiple factors, meaning py311-django42 will be split into two factors: py311 and django42.

docs/developer/hyperion/system-tests.rst
```diff
@@ -0,0 +1,20 @@
+System Tests
```
olliesilvester, line 1: If there are any system tests which we permanently want to leave in the option to locally run them, we should explain how to run them here

tests/system_tests/conftest.py
```diff
@@ -119,6 +122,15 @@
 }
 
 
+@pytest.fixture(autouse=True)
+def use_dev_ispyb_unless_overridden_by_environment():
+    ispyb_config_path = os.environ.get(
+        "ISPYB_CONFIG_PATH", CONST.SIM.DEV_ISPYB_DATABASE_CFG
```
olliesilvester, line 128: Where is `ISPYB_CONFIG_PATH` set? I can't find a corresponding name in the Dockerfiles

tests/unit_tests/conftest.py
```diff
@@ -52,10 +52,3 @@ def device_factories_for_beamline(beamline_module: ModuleType) -> set[AnyDeviceF
         for f in collect_factories(beamline_module, include_skipped=True).values()
         if hasattr(f, "cache_clear")
     }
-
-
-@pytest.fixture(scope="function", autouse=True)
-def clear_device_factory_caches_after_every_test(active_device_factories):
```
olliesilvester, line 58: Is this definitely not needed?

tests/system_tests/conftest.py
rtuck99: We don't override it for the containerised system tests - the test credentials are mounted to the default place.

tests/unit_tests/conftest.py
rtuck99: It got moved up a level

pyproject.toml
callumforrester: Hmm, should we also change `type-checking` then?

DominicOram (approved):
Great, thank you!

PR #796, opened 2025-02-04
Fix issues from testing

### LINKED ISSUE #769: Testing 27/01/25
- [ ] Group of full collections
- [ ] Changing energy between collections
- [ ] S+C

DominicOram:
Fixes #769 

Requires changes from https://github.com/DiamondLightSource/dodal/pull/1036

To test:
* Confirm tests pass
* Confirm all problems noted in issue are covered by this PR or other new issues


src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -460,3 +459,4 @@ def rotation_scan_core(
         _multi_rotation_scan(),
         group=CONST.WAIT.ROTATION_READY_FOR_DC,
     )
```
rtuck99, line 461: The run engine keeps track of staged devices, and automatically unstages them in the event of an exception. Do we care about this not happening any more? Also, as far as I can tell at the moment there isn't any validation about balanced stage/unstage calls but I don't know if that's something I would want to rely on.

DominicOram: In the unhappy path `start_preparing_data_collection_then_do_plan` will call `stop` on the detector (https://github.com/DiamondLightSource/mx-bluesky/blob/d490273c6662bd7210dc8b5ecc81107a7d1b43e7/src/mx_bluesky/hyperion/device_setup_plans/utils.py#L58) so that should be fine. Yes, we should balance the `stage`/`unstage`. With the fastCS Eiger we can stop using `yield from bps.abs_set(eiger.do_arm, 1, group=group)` to do an aysnchronous `stage` and just properly use `stage`/`unstage` again

rtuck99 (approved):
Approved

PR #801, opened 2025-02-06
Collect where we are if no centering needed

### LINKED ISSUE #791: Hyperion: If there is no need to center then collect at current co-ordinates
Currently if we decide that we do not need to centre (because the sample/chi has not changed) we end up collecting at 0, 0, 0. Instead we should collect at the current location.

## Note
* There is currently a bad hotfix on the beamline for this

## Acceptance Criteria
- When no centring is required then data is collected at the location we currently are

DominicOram:
Fixes #791

Note: I'm not 100% happy with the logic of this, it's not great that the assumption of what an empty `xray_centre_results` means is in `load_centre_collect_full`. However, the whole thing of how we work out if we need to do an XRC or not is a fudge and will get significantly tidier when hyperion is just called once per collection so I think this is a good compromise.

To test:
* Confirm new unit test passes


rtuck99 (approved):
Approved

Issue #632, opened 2024-11-11
 Fix system tests which don't require S03


olliesilvester:
Following on from https://github.com/DiamondLightSource/mx-bluesky/issues/631, we should make sure our useful system tests all pass, so that we can stop ignoring the daily email about them failing

## Acceptance Criteria
- Remaining system tests pass

DominicOram:
Some of them have been fixed in https://github.com/DiamondLightSource/mx-bluesky/pull/625

PR #802, opened 2025-02-07
632 Add external interaction system tests to the containerised system tests

rtuck99:
This addresses 

* #632 

and builds on

* #793

it also requires

* DiamondLightSource/dodal#1049
* https://gitlab.diamond.ac.uk/MX-GDA/hyperion-system-testing/-/merge_requests/2

These changes fix up and enable the following system tests when the containerised system tests are run:

* `test_external_callbacks`   -- external callback launching
* `test_exp_eye_dev` -- ExpEye integration
* `test_ispyb_dev_connection` -- ISPyB integration
* `test_nexgen` -- imginfo integration
* `test_zocalo_system` -- zocalo integration (using the "fake" zocalo

`test_config_service` is not currently included, to include this ideally we should make changes to the `daq-config-server` images so that instead of creating dev/production images with hard-coded configurations, it pushes a single set of images which are fully configurable for either production or development/test environments. See

* #817

### DIFF
--- a/tests/system_tests/hyperion/external_interaction/test_zocalo_system.py
+++ b/tests/system_tests/hyperion/external_interaction/test_zocalo_system.py
@@ -4,7 +4,6 @@ import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
 import numpy as np
 import pytest
-import pytest_asyncio
 from bluesky.run_engine import RunEngine
 from dodal.devices.zocalo import ZOCALO_READING_PLAN_NAME, ZocaloResults
 from dodal.utils import is_test_mode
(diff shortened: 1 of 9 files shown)


tests/system_tests/hyperion/external_interaction/test_zocalo_system.py
```diff
@@ -106,28 +98,33 @@ def inner_plan():
     return inner
 
 
-@pytest.mark.s03
+@pytest.mark.system_test
+def test_is_test_mode():
+    assert is_test_mode(), "DODAL_TEST_MODE must be set in environment before launch"
```
DominicOram, line 103: I think this could be cleaner but I've made a new issue https://github.com/DiamondLightSource/dodal/issues/1064

DominicOram (approved):
Great, thank you!

tests/system_tests/hyperion/external_interaction/test_zocalo_system.py
rtuck99: Yes, it's horrible isn't it.

PR #803, opened 2025-02-10
Get loop type from Agamemnon

### LINKED ISSUE #788: Multipin: Use loop type
Collecting a multipin in hyperion is the same as a single pin except:
* The width of the XRC grid to use is larger
* We want to select more centres, e.g. increase the `n` in `TopNByMaxCountSelection`

To allow a mix of single pins and multipins we need to pass on the `loopType` (which is currently in ispyb and the Agamemnon instruction) to hyperion so that it can work out how to change these parameters

Note:
* We need to confirm if we want this working with manual GDA collections or Agamemnon collections first

## Acceptance Criteria
- When a loop has a multipin loop type Hyperion uses the above different parameters

DominicOram:
Fixes #788 

To test:
* Confirm new unit tests and system test passes


src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def _single_pin() -> PinType:
+    return PinType(1, GridscanParamConstants.WIDTH_UM)
+
+
+def _get_parameters_from_url(url: str) -> dict:
+    response = requests.get(url, headers={"Accept": "application/json"})
+    response.raise_for_status()
+    return json.loads(response.content)["collect"]
```
olliesilvester, line 41: Could: Give an error/loggging message which includes the `json.loads(response.content)` as a string if the `collect` key doesn't exist

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def _get_pin_type_from_agamemnon_parameters(parameters: dict) -> PinType:
```
olliesilvester, line 44: If `loop_type_name` is `None` then we return single pin with no logging message - just checking this is intended behaviour?

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def update_params_from_agamemnon(parameters: T) -> T:
+    try:
+        pin_type = get_pin_type_from_agamemnon(parameters.beamline)
+        if isinstance(parameters, LoadCentreCollect):
+            parameters.robot_load_then_centre.grid_width_um = pin_type.full_width
+            parameters.select_centres.n = pin_type.expected_number_of_crystals
+    except Exception as e:
+        LOGGER.warning(f"Faild to get pin type from agamemnon, using single pin {e}")
```
olliesilvester, line 72:
```suggestion
        LOGGER.warning(f"Failed to get pin type from agamemnon, using single pin {e}")
```

tests/system_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -0,0 +1,12 @@
+from mx_bluesky.hyperion.external_interaction.agamemnon import (
+    AGAMEMNON_URL,
+    _get_parameters_from_url,
+    _get_pin_type_from_agamemnon_parameters,
+    _single_pin,
+)
+
+
+def test_given_test_agamemnon_instruction_then_returns_none_loop_type():
```
olliesilvester, line 9: Would be nice to have full containerised system tests for this (as a new issue)

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
olliesilvester: Also it would be good to validate somewhere that the input here is of the expected format. Eg `string-floatxfloat` (or whatever the regex is for that)

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -0,0 +1,138 @@
...
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.requests")
+def test_given_agamemnon_returns_multipin_when_get_next_pin_type_from_agamemnon_called_then_multipin_returned(
+    mock_requests: MagicMock,
+):
+    configure_mock_agamemnon(mock_requests, "multipin-6x50")
+    assert get_pin_type_from_agamemnon("i03") == PinType(6, 50)
+
+
+@pytest.fixture
+def load_centre_collect_params():
```
olliesilvester, line 100: Could: This same function is duplicated in some other test files, could move it somewhere common (probably conftest for now, although there is an issue somewhere to create a `testing` directory

olliesilvester (approved):
Looks good and nice test coverage. I have some comments but all optional

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Yep, I have also remembered that from testing on the beamline the format of the string describing the multipin has changed slightly. I will fix this now

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -38,16 +37,16 @@ def params_from_loop_type(loop_type: str | None):
...
     [
-        ("multipin-6x50", PinType(6, 50)),
-        ("multipin-6x25.8", PinType(6, 25.8)),
-        ("multipin-9x31", PinType(9, 31)),
+        ("multipin_6x50+9", PinType(6, 50, 9)),
+        ("multipin_6x25.8+8.6", PinType(6, 25.8, 8.6)),
+        ("multipin_9x31+90", PinType(9, 31, 90)),
     ],
 )
 def test_given_multipin_loop_type_in_parameteers_then_expected_pin_returned(
```
olliesilvester, line 52: You've typo'd `parameteers` on a bunch of these tests!

olliesilvester (approved):
Approving but a few comments again - nothing major

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,97 @@
...
+def _get_pin_type_from_agamemnon_parameters(parameters: dict) -> PinType:
+    loop_type_name: str | None = parameters["sample"]["loopType"]
+    if loop_type_name:
+        regex_search = re.search(MULTIPIN_REGEX, loop_type_name)
+        if regex_search:
+            wells, well_size, tip_to_first_well = regex_search.groups()
+            return PinType(int(wells), float(well_size), float(tip_to_first_well))
+        else:
+            warning_message = f"Agamemnon loop type of {loop_type_name} not recognised, assuming single pin"
+            if loop_type_name.startswith(MULTIPIN_PREFIX):
```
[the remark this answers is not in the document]
DominicOram: I agree, let's throw an Exception

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -0,0 +1,177 @@
...
+)
+def test_given_unrecognised_multipin_in_parameters_then_warning_logged_single_pin_returned(
+    loop_name: str,
+):
+    with pytest.raises(Exception) as e:
+        assert (
+            _get_pin_type_from_agamemnon_parameters(params_from_loop_type(loop_name))
+            == SinglePin()
+        )
+        assert "Expected multipin format" in str(e.value)
```
shihab-dls, line 95:
```suggestion
    with pytest.raises(Exception) as e:
        _get_pin_type_from_agamemnon_parameters(params_from_loop_type(loop_name))
    assert "Expected multipin format" in str(e.value)
```
I think it was agreed to throw an exception instead of returning `SinglePin()`, so this assertion should be removed. Moreover, `assert "Expected multipin format" in str(e.value)` is never actually ran, so should be outside the manager.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,99 @@
...
+from mx_bluesky.common.parameters.components import WithVisit
+from mx_bluesky.common.parameters.constants import GridscanParamConstants
+from mx_bluesky.common.utils.log import LOGGER
+from mx_bluesky.hyperion.parameters.load_centre_collect import LoadCentreCollect
+
+T = TypeVar("T", bound=WithVisit)
+AGAMEMNON_URL = "http://agamemnon.diamond.ac.uk/"
+MULTIPIN_PREFIX = "multipin"
+MULTIPIN_FORMAT_DESC = "Expected multipin format is multipin_{number_of_wells}x{well_size}+{distance_between_tip_and_first_well}"
+MULTIPIN_REGEX = rf"{MULTIPIN_PREFIX}_(\d+)x(\d*\.?\d*)\+(\d*\.?\d*)"
```
shihab-dls, line 18: Not sure how important this really is, but this regex allows for the following cases:
`"multipin_6x50+"`
`"multipin_6x50+98."`
`"multipin_6x50+.1"`
I've tried to make this stricter, and came up with this:
```suggestion
MULTIPIN_REGEX = rf"^{MULTIPIN_PREFIX}_(\d+)x(\d+(?:\.\d+)?)\+(\d+(?:\.\d+)?)$"
```
which should prevent invalid floats, and also anchors the string. Tuning this is probably never-ending though, so whatever you think is best! I've commented additional tests we can use, if you do use this regex.

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -0,0 +1,177 @@
...
+        == SinglePin()
+    )
+    mock_logger.warning.assert_called_once()
+
+
+@pytest.mark.parametrize(
+    "loop_name",
+    [
+        "multipin_67x56",
+        "multipin_90+4",
```
shihab-dls, line 84: Following my previous comment, these invalid pin types can be tests with a stricted regex, but again not sure how important this really is:
```suggestion
        "multipin_67x56",
        "multipin_90+4",
        "multipin_8",
        "multipin_6x50+",
        "multipin_6x50+98.",
        "multipin_6x50+.1",
        "multipin_6x.50+98",
        "multipin_6x50+98.1.2",
        "multipin_6x50.5.6+98",
        "multipin_6x50+98..1",
        "multipin_6x.50+.98",
        "multipin_6x+98",
```

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -0,0 +1,177 @@
...
+    ],
+)
+def test_given_unrecognised_multipin_in_parameters_then_warning_logged_single_pin_returned(
+    loop_name: str,
+):
+    with pytest.raises(Exception) as e:
```
shihab-dls, line 90: Nit: Should we be passing a general `Exception`? Maybe a `ValueError` or adjacent? Up to you!

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,99 @@
...
+        raise Exception(f"Unexpected json from agamemnon: {response_json}") from e
+
+
+def _get_pin_type_from_agamemnon_parameters(parameters: dict) -> PinType:
+    loop_type_name: str | None = parameters["sample"]["loopType"]
+    if loop_type_name:
+        regex_search = re.search(MULTIPIN_REGEX, loop_type_name)
+        if regex_search:
+            wells, well_size, tip_to_first_well = regex_search.groups()
+            return PinType(int(wells), float(well_size), float(tip_to_first_well))
```
shihab-dls, line 69: Related to my other regex comments, passing `multipin_6x50+` with the current regex would provide a valid `regex_search` but would raise a ValueError on `float(tip_to_first_well)`, as `tip_to_first_well` would be captured as an empty string. I think the stricter regex should deal with this though.

shihab-dls (changes requested):
Looks good to me! Just left a few comments on regex, but I'd go with your outlook on whether to apply them. Main change requested is to do with a test assertion.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Love it, thanks!

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Yep, thanks!

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def _get_parameters_from_url(url: str) -> dict:
+    response = requests.get(url, headers={"Accept": "application/json"})
+    response.raise_for_status()
+    response_json = json.loads(response.content)
+    try:
+        return response_json["collect"]
+    except KeyError as e:
+        raise Exception(f"Unexpected json from agamemnon: {response_json}") from e
```
shihab-dls, line 60: Nit: We could re-throw the `KeyError` here, after wrapping more context, for better stack trace integrity.
 ```suggestion
         raise KeyError(f"Unexpected json from agamemnon: {response_json}") from e
```
Definitely just a nit, since we propagate up to `update_params_from_agamemnon` which wraps this or the `HTTPError` from `raise_for_status()` (or some other error we don't know about) into an `Exception` anyways, and then swallow the error with a logging statement. So I think we'd lose information about the error type regardless, and would just rely on the f-string reason.

File: tests/unit_tests/hyperion/external_interaction/test_agamemnon.py [hunk 14]
```diff
@@ -0,0 +1,184 @@
...
+        headers={"Accept": "application/json"},
+    )
+
+
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.requests")
+def test_given_agamemnon_returns_an_unexpected_response_then_exception_is_thrown(
+    mock_requests: MagicMock,
+):
+    mock_requests.get.return_value.content = json.dumps({"not_collect": ""})
+    with pytest.raises(Exception) as e:
```
shihab-dls, line 128: Nit: Following my last nit, even though we swallow these errors, raising `KeyError` would let use do:
```suggestion
    with pytest.raises(KeyError) as e:
```
which I generally like more in unit tests for specificity.

shihab-dls (approved):
Thank you! Looks great now. I commented a few nits in regards to exception handling, but approving!

Issue #794, opened 2025-02-04
Common XRC gridscans: Incrementally merge the vmxm_fgs branch


olliesilvester:
I have written a common fast grid scan plan along with i03 and vmxm's implementations of this plan on a [separate branch](https://github.com/DiamondLightSource/mx-bluesky/tree/vmxm_fgs). To make the reviewing more effective, these should be broken down into a few separate PR's

## Acceptance Criteria - to merge:
- Make generic `read_hardware` plans
- Modify parameter model + config server
- Move composite devices to separate file + move other small bits to common (eg `XRayCentreEventHandler`) and `context.py`
- Make common top-level XRC plan along with the Hyperion entry to this plan
- Make VMXm entry to top-level XRC plan

PR #806, opened 2025-02-11
Common XRC: Modify parameters and composite devices

olliesilvester:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/794
- Add a general `feature_dependant_config` function to `FeatureFlags`, which handles configuration relying on the live feature settings. and is done before the XRC kicks off
- Add `detector_params` to `SpecifiedThreeDGridScan`, using a dictionary mapping beamline to detector size. This means `SpecifiedThreeDGridScan` is not abstract so can be used by other beamlines wanting to do XRC
- Start moving device composites to a separate file - this is just to avoid circular imports
- (TODO) also move a few small bits to `common`


src/mx_bluesky/hyperion/parameters/device_composites.py
```diff
@@ -0,0 +1,49 @@
...
+from dodal.devices.undulator import Undulator
+from dodal.devices.xbpm_feedback import XBPMFeedback
+from dodal.devices.zebra.zebra import Zebra
+from dodal.devices.zebra.zebra_controlled_shutter import ZebraShutter
+from dodal.devices.zocalo import ZocaloResults
+from ophyd_async.fastcs.panda import HDFPanda
+
+
+@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
+class HyperionFlyScanXRayCentreComposite:
```
olliesilvester, line 29: Not immediately obvious but we get circular imports without moving this

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -55,3 +55,6 @@ def update_self_from_server(self):
                 else self.overriden_features[flag]
             )
             setattr(self, flag, updated_value)
+
+    def feature_dependant_config(self, *args, **kwargs):
+        pass
```
rtuck99, line 60: This seems like a bit of a code smell.

We have an abstract method (which should be decorated as such btw), but it takes arbitrary parameters, and it's also not used by the base class.

So there is nothing that governs when this is called, or what any class that encounters a `FeatureFlags` should call it with  - the only methods that can call it have to know what to supply because they know what the concrete type is.

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -18,11 +21,12 @@
     XyzStarts,
 )
 from mx_bluesky.common.parameters.constants import (
+    DetectorParamConstants,
     GridscanParamConstants,
     HardwareConstants,
 )
 
-"""Parameter models in this file are abstract. They should be inherited by a top-level model"""
```
rtuck99, line 25: Is this no longer true?

src/mx_bluesky/common/external_interaction/config_server.py
olliesilvester: As discussed, since this function is currently only being used to update the composite device, we can do this before the common part of the plan is actually called.

I'll remove this bit from this PR!

src/mx_bluesky/common/parameters/gridscan.py
olliesilvester: This PR adds `detector_params` to `SpecifiedThreeDGridScan`, so this class isn't abstract anymore

rtuck99 (approved):
Approved

Issue #731, opened 2025-01-07
 Prevent Hyperion from being invoked from outside the beamline


rtuck99:
Once Hyperion is deployed to production beamline we will need some way of ensuring that plans can't be executed by calls from outside the beamline network.

As a minimum we should ensure that only HEAD/GET requests are permitted, it seems like this ought to be something that we can do in the kubernetes Ingress configuration.

If this isn't possible, we could just block all access from outside the beamline.

## Acceptance Criteria
- Clients from outside the beamline network cannot execute Hyperion plans or make other writes

rtuck99:
Currently waiting for
https://jira.diamond.ac.uk/servicedesk/customer/portal/2/SCHD-7502

rtuck99:
Blocked whilst Giles is away, need to deploy EPICS gateway in beamline cluster

rtuck99:
See also GitLab merge request:

https://gitlab.diamond.ac.uk/controls/containers/beamline/i03-services/-/merge_requests/11

rtuck99:
Currently waiting for investigation by scicomp into reasons why PVA gateway does not seem to work for OAV - Giles + Thomas currently looking into it

DominicOram:
@DominicOram to discuss with everyone involved

PR #809, opened 2025-02-12
731 Add source IP access control to k8s hyperion

rtuck99:
This addresses
* #731 

It adds an annotation to the `ingress.yml` that whitelists the beamline subnet, this in conjuction with a change to the DNS entry in SCHD-7502 servicedesk ticket that ensures source IPs are preserved in requests, enables requests to `i03-hyperion.diamond.ac.uk` to be only allowed for hosts on the beamline.

In addition this PR contains various other minor fixes to deployment / image build scripts / documentation to make it easier to deploy test images.

Acceptance criteria:

* Requests to `i03-hyperion.diamond.ac.uk` can't be made from outside the i03 beamline.

### DIFF
--- a/docs/developer/hyperion/deploying-hyperion.rst
+++ b/docs/developer/hyperion/deploying-hyperion.rst
@@ -100,9 +100,10 @@ The dev deployment bind-mounts the current ``hyperion`` workspace and ``../dodal
 run against your own development code. **Clusters do not allow bind mounts from arbitrary directories so
 your workspace will have to be in a permitted directory such as your home directory.**
 
-By default the script will log into the ``argus`` cluster, if you want to deploy to an alternate cluster,
-log in with ``kubectl set-context --current --namespace=<NAMESPACE>`` and then specify ``--no-login`` when running the
-script
+By default the script will log into the ``argus`` cluster if ``--dev`` is specified, otherwise the beamline cluster.
+If you want to deploy to an alternate namespace/cluster, change cluster with ``module load <cluster module>``
+and then set your namespace with ``kubectl config set-context --current --namespace=<NAMESPACE>`` and then
+specify ``--no-login`` when running the script.
 
 Please note, the deployment script is intended to be run from a checked-out matching version of the git repository.
 
--- a/helmchart/templates/ingress.yaml
+++ b/helmchart/templates/ingress.yaml
@@ -3,6 +3,8 @@ apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
   name: hyperion-ingress
+  annotations:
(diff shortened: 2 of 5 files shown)


DominicOram (changes requested):
This is probably my lack of understanding but it seems we're exposing the hyperion API port using a `LoadBalancer` under `service.yaml`? So is the ingress just giving us a nicer entry point so that we can send hyperion requests to `i03-hyperion.diamond.ac.uk`? Does this mean we could still send to `{hyperion's_k8s_ip}:5005` and because this protection is on the ingress, not the `LoadBalancer` it gets through?

DominicOram:
> > This is probably my lack of understanding but it seems we're exposing the hyperion API port using a `LoadBalancer` under `service.yaml`? So is the ingress just giving us a nicer entry point so that we can send hyperion requests to `i03-hyperion.diamond.ac.uk`? Does this mean we could still send to `{hyperion's_k8s_ip}:5005` and because this protection is on the ingress, not the `LoadBalancer` it gets through?
> 
> This is a good catch! I should have thought about this. As it turns out in theory we are ClusterIP service which ought to prevent us being reachable from outside but unfortunately we are also configured with `hostNetwork: true` (because reasons...) so when I actually tried this, you can actually reach hyperion on the nodes port.
> 
> I am currently investigating with Thomas Hartland to see whether there's any way round this.

Great, thanks. Could we get some more documentation on this as well then? For example:
* Why we need an ingress rather than just using the load balancer (if it's just so we don't have to remember the ip address that's fine)
* Why we need `hostNetwork: true`

rtuck99:
So an ingress is beneficial for the following reasons, (not necessarily immediately):

* To get a convenient domain name assigned to the service `i03-hyperion.diamond.ac.uk`
* As a convenient endpoint for https termination - the ingress is managed centrally and inherits the wildcard diamond.ac.uk certificate
* We get the benefit of being able to potentially delegate access control without code changes to hyperion, depending on how we decide what our ultimate solution is.
* Isolation of the service is better as we are able to use `ClusterIP` (effectively this means that the services are running on a private network that is not reachable, outbound requests are NAT'ed)

There are other benefits, it gives you most of the functionality of a reverse proxy (it is basically a wrapper around Nginx), things like load balancing, request rewriting, virtual hosting etc. but we aren't really interested in those.

The `hostNetwork` was originally required because:
> There's a reason for this, we block access to the instrumentation network from pods that aren't using hostNetwork: true.
This works around a bug in some GigE camera devices, where the IOC pod tells the camera to respond to it on the pod's kubernetes internal IP address, which the camera can never reach.

however this is no longer the case and we can turn it off, but this has introduced additional problems relating to PV access, hence now this GitLab merge request for a PV gateway:

https://gitlab.diamond.ac.uk/controls/containers/beamline/i03-services/-/merge_requests/11

Issue #784, opened 2025-01-30
Make it clear to a user that we're skipping collections after blank XRC


DominicOram:
Now that we're skipping bad samples (https://github.com/DiamondLightSource/mx-bluesky/issues/687) we should make this as clear as possible in the XRC comments so that a user knows what is going on. 

## Acceptance Criteria
- When we get no diffraction for an XRC we put something like `Diffraction not found, skipping sample.` in the XRC comments

DominicOram:
There's a hotfix on the beamline for this, needs tidying

PR #814, opened 2025-02-12
Make it clear to a user that we're skipping collections after blank XRC

shihab-dls:
Fixes issue #784 
Added the hotfix @DominicOram wrote in the beamline deployment, and a unit test to test the ispyb callback stop event.

### DIFF
--- a/src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
+++ b/src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
@@ -43,7 +43,11 @@ from mx_bluesky.common.parameters.constants import DocDescriptorNames, PlanNameC
 from mx_bluesky.common.parameters.gridscan import (
     GridCommon,
 )
-from mx_bluesky.common.utils.exceptions import ISPyBDepositionNotMade
+from mx_bluesky.common.utils.exceptions import (
+    CrystalNotFoundException,
+    ISPyBDepositionNotMade,
+    SampleException,
+)
 from mx_bluesky.common.utils.log import ISPYB_ZOCALO_CALLBACK_LOGGER, set_dcgid_tag
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


shihab-dls:
I assigned the conftest runstop event `test_gridscan3d_stop_document_with_crystal_exception` the values:
`"time": 1666604299.6149616`
`"uid": "65b2bde5-5740-42d7-9047-e860e06fbe15"`
arbitrarily, as they are required; I think this is fine, but commenting just in case it is not.

DominicOram (changes requested):
Thanks for the test. As discussed, I think there's a cleaner solution

DominicOram:
> I assigned the conftest runstop event `test_gridscan3d_stop_document_with_crystal_exception` the values: `"time": 1666604299.6149616` `"uid": "65b2bde5-5740-42d7-9047-e860e06fbe15"` arbitrarily, as they are required; I think this is fine, but commenting just in case it is not.

Yep, this is fine, thanks!

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -281,5 +285,10 @@ def activity_gated_stop(self, doc: RunStop) -> RunStop:
             )
             if self.ispyb_ids == IspybIds():
                 raise ISPyBDepositionNotMade("ispyb was not initialised at run start")
+            exception_type, message = SampleException.type_and_message_from_reason(
+                doc.get("reason", "")
+            )
+            if exception_type == CrystalNotFoundException.__name__:
+                doc["reason"] = message
```
DominicOram, line 292: Could: I think we can just do this for all exceptions e.g.
```suggestion
            if exception_type:
                doc["reason"] = message
```

DominicOram (approved):
Thank you! Much cleaner than my hotfix. Optional change, take it or leave it.

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
shihab-dls: Ah yeah, I think this is much better; no point restricting what errors we append reasons for.

Issue #785, opened 2025-01-30
Write the XRC comments such that synchweb displays the one with results


DominicOram:
When we do an XRC we put the results in the comment for the 1st grid. The syncweb logic for which comment to display on the main data collection page for data collection groups is to use the `min` of the individual collection's comments. This means that it will display the one that is alphabetically first e.g.

https://ispyb.diamond.ac.uk/dc/visit/mx23694-127/dcg/13879315 displays ```Comment: MX-Bluesky: Xray centring - Diffraction grid (30x16). Aperture: Small. Zocalo processing took 0.03 s. Crystal 1: Strength 19531; Position (grid boxes) ['6.43', '6.5', '5.86']; Size (grid boxes) [7 1 6]; Crystal 2: Strength 17079; Position (grid boxes) ['7.23', '8.5', '5.87']; Size (grid boxes) [5 1 6]; Crystal 3: Strength 13225; Position (grid boxes) ['6.76', '4.5', '5.79']; Size (grid boxes) [4 1 6]; ```

but https://ispyb.diamond.ac.uk/dc/visit/mx23694-127/dcg/13879651 displays ```Comment: MX-Bluesky: Xray centring - Diffraction grid (30x13). Aperture: Small. ```

We should prepend something on the comments so that synchweb chooses the one we're adding results to.

## Acceptance Criteria
- The alphabetically first comment is always the one that we add zocalo results to

PR #827, opened 2025-02-17
785 Write XRC results to alphabetically first comment so that they appear in Synchweb

rtuck99:
Fixes 

* #785

* Calls to `StoreInIspyb.update_deposition()` should now result in a call to the append comment sproc instead of blindly overwriting any existing comments.
* The initial upsert to the DataCollection table now has the comment prefix string including the gridscan number to ensure lexicographic sorting.

Unfortunately there has been significant churn to some of the unit tests because of the above changes, and also because some of the tests were written to emulate older behaviour whereby the second DataCollection in the gridscan was not created on the start event - this has not been true for some time and the tests no longer emulate this.

1. The XRC results (i.e. crystals found) comment should now always appear in the SynchWeb summary for the XRC.
2. Gridscan data collection comments should now be prefixed with `MX-Bluesky: Xray centring n - ` where n is 1 or 2
3. Tests pass

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
+++ b/src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
@@ -116,7 +116,7 @@ class GridscanISPyBCallback(BaseISPyBCallback):
             scan_data_infos = [
                 ScanDataInfo(
                     data_collection_info=populate_remaining_data_collection_info(
-                        None,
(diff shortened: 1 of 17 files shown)


shihab-dls (approved):
Looks great to me! Given the number of changes, I'd feel more confident with @olliesilvester or @DominicOram taking a look too, but I'll approve.

olliesilvester (approved):
Looks good! I didn't check the tests very thoroughly but I trust @shihab-dls  on those :grin:

Issue #826, opened 2025-02-17
Multipin tests 17/02


DominicOram:
* Need to trigger zocalo once at the end of all collections rather than in between each (https://github.com/DiamondLightSource/mx-bluesky/pull/850)
* Occassionally taking a collection in empty space https://ispyb.diamond.ac.uk/dc/visit/mx23694-130/id/16949455. This was because agamemnon gave us back every sublocation and so we were collecting on the same pin again. Will be fixed by a change in agamemnon to only give us the first sample

DominicOram:
Issue with long pins being out of range, where pin tip detection continually inverts step size (see code)

DominicOram:
Weird streaking on lines https://ispyb.diamond.ac.uk/dc/visit/mx23694-130/dcg/14178031

DominicOram:
I hacked in something to skip snapshots and invert every other collection

DominicOram:
When calling `get_pin_type_from_agamemnon(parameters.beamline)` in https://github.com/DiamondLightSource/mx-bluesky/pull/803. `parameters.beamline` gives the PV prefix `BL03I` but for agamamnon we need the beamline name i.e. `i03`

DominicOram:
We're looking in the wrong place for panda settings, should be `f"{ROOT_DIR}/hyperion/resources/panda/"` rather than `f"{ROOT_DIR}/hyperion/resources/panda/{PANDA_FLYSCAN_SETTINGS_FILENAME}"`

PR #829, opened 2025-02-17
Keep moving in the same direction if pin tip not found

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/826

Previously we were doing `step_size * -1` in each loop of the search, meaning we would alternate between -0.5 and +0.5 search direction.

1. Confirm test now passes

### Checks for reviewer


src/mx_bluesky/hyperion/experiment_plans/pin_tip_centring_plan.py
```diff
@@ -83,18 +83,18 @@ def pin_tip_valid(pin_xy: Tip):
         if pin_tip_valid(tip_xy_px):
             return (int(tip_xy_px[0]), int(tip_xy_px[1]))
 
-        if tip_xy_px[0] == 0:
-            # Pin is off in the -ve direction
-            step_size_mm = -step_size_mm
+        # Pin is off in the -ve direction
```
shihab-dls, line 86:
```suggestion
        # Pin is off in the -ve direction if the returned tip x pixel value is 0
```
Could: Amend this comment to specify which condition suggests a -ve pin direction

Issue #780, opened 2025-01-30
Remove results info from XRC comment in ispyb


DominicOram:
We currently list the results we get from zocalo in the ispyb comment during XRC. This is quite verbose and is _mostly_ a repeat of what is in the banner below. The only thing missing is the strength, which is being put in.

![Image](https://github.com/user-attachments/assets/7c5f7b3a-7174-417b-96eb-0529df0e677b)

Once the strength is in we can remove putting the results in the comment.

## Acceptance Criteria
- The XRC ispyb comment doesn't contain the zocalo results

DominicOram:
@ndg63276 do you have a link to the synchweb issue for this?

ndg63276:
https://jira.diamond.ac.uk/browse/LIMS-1637

DominicOram:
Strength is now there in `staging`. Size isn't there yet  but @ndg63276 is putting it in

PR #832, opened 2025-02-18
Use group comments for XRC

### LINKED ISSUE #782: Use data collection group comments for XRC
Currently for data collection groups synchweb selects the comments from one of the two individual collections in the group. For XRC this means the comments only contain information on one of the 2D grids. There is a comment field for the group as a whole, which will be used after https://jira.diamond.ac.uk/browse/LIMS-1637. We should start filling this out in preparation containing information that relates to the whole grid e.g. 
* x, y, z number of steps
* zocalo timings
* zocalo results (depending on status of https://github.com/DiamondLightSource/mx-bluesky/issues/780)
* Whether there was diffraction/skipping the sample

## Acceptance Criteria
- The comments in the data collection group for XRC contains a sensible comment

shihab-dls:
Fixes #782 

This PR, branching off of #785, adds an entrypoint, `update_data_collection_group_table`, to update the data collection group table, which simply calls `_store_data_collection_group_table`.

An instance of `DataCollectionGroupInfo` is now maintained in `GridscanISPyBCallback`, such that comments from the OAV and Zocalo events are appended to its `comments` field. `update_data_collection_group_table` is then called at the end of an event to update the comment. Similarly, a run stop with an exception sets the exception reason as the group comment. Only `sample_id` is required when updating the group table; thus, a better approach may be to only maintain a comment, passing in null for all other fields but `sample_id`. However, `visit_string` and `experiment_type` in `DataCollectionGroupInfo` are required fields, so only `sample_barcode` would not need to be maintained., and thus this is not done.

1. Check that group comments are in the correct form
2. Confirm group comments are updated appropriately

### Checks for reviewer


shihab-dls:
So far, these comments are included as group info:

- x, y, z number of steps
- zocalo timings
- zocalo results (As #780 is still unresolved)
- skipping sample

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -289,5 +317,11 @@ def activity_gated_stop(self, doc: RunStop) -> RunStop:
             )
             if exception_type:
                 doc["reason"] = message
+                self.data_collection_group_info.comments = message
```
olliesilvester, line 320: Line 320 makes line 319 not needed

src/mx_bluesky/common/external_interaction/ispyb/ispyb_store.py
```diff
@@ -152,6 +152,19 @@ def append_to_comment(
                 data_collection_id, comment, delimiter
             )
 
+    def update_data_collection_group_table(
+        self,
+        dcg_info: DataCollectionGroupInfo,
+        data_collection_group_id: int | None = None,
+    ) -> None:
+        with ispyb.open(self.ISPYB_CONFIG_PATH) as conn:
+            assert conn is not None, "Failed to connect to ISPyB!"
```
olliesilvester, line 161:
```suggestion
            assert conn, "Failed to connect to ISPyB!"
```

tests/unit_tests/common/external_interaction/xray_centre/test_ispyb_callback.py
```diff
@@ -294,6 +296,12 @@ def test_activity_gated_event_oav_snapshot_triggered(self, mock_ispyb_conn):
             },
         )
 
+        group_dc_cols = remap_upsert_columns(
+            mx_acq.get_data_collection_group_params(),
+            mx_acq.upsert_data_collection_group.mock_calls[1].args[0],
+        )
+        assert group_dc_cols["comments"] == "Diffraction grid scan of 40 by 20 by 10."
```
olliesilvester, line 314: @DominicOram Is it obvious to whoever's looking at the ispyb comments that this means a 40x20 grid followed by a 40x10 grid?

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -85,6 +88,15 @@
 ReWithSubs = tuple[RunEngine, tuple[GridscanNexusFileCallback, GridscanISPyBCallback]]
 
 
+@pytest.fixture
+def dummy_rotation_data_collection_group_info():
+    return DataCollectionGroupInfo(
+        visit_string="cm31105-4",
+        experiment_type="SAD",
+        sample_id=364758,
+    )
```
olliesilvester, line 97: Should: Instead of making the same fixture in two files it's better to put it in a `conftest` file which both of the files can use

tests/unit_tests/hyperion/experiment_plans/test_grid_detection_plan.py
```diff
@@ -39,6 +42,15 @@
 from ...conftest import assert_event
 
 
+@pytest.fixture
+def dummy_rotation_data_collection_group_info():
```
olliesilvester, line 46: See other comment regarding this fixture

olliesilvester (changes requested):
Thanks, this is nice! A few comments but nothing major.

I think there's also room for a few more tests here. Maybe something to check your new assertions, and to check that `data_collection_group_table` looks correct if we get an exception before we see any events - i.e:

```
if exception_type:
      doc["reason"] = message
      self.data_collection_group_info.comments = message
      self.ispyb.update_data_collection_group_table(
          self.data_collection_group_info,
          self.ispyb_ids.data_collection_group_id,
      )
```

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -145,7 +145,9 @@ def activity_gated_start(self, doc: RunStart):
         return super().activity_gated_start(doc)
 
     def activity_gated_event(self, doc: Event):
-        assert self.data_collection_group_info, "No data collection group info"
+        assert self.data_collection_group_info, (
+            f"No data collection group info - event document has been emitted before a {PlanNameConstants.GRID_DETECT_AND_DO_GRIDSCAN} start document"
```
olliesilvester, line 149: Should: pull this string into a constant

src/mx_bluesky/common/external_interaction/callbacks/xray_centre/ispyb_callback.py
```diff
@@ -185,13 +198,19 @@ def _handle_zocalo_read_event(self, doc):
...
+        ) + crystal_summary
+
         self.ispyb.append_to_comment(
             self.ispyb_ids.data_collection_ids[0], crystal_summary
         )
 
     def _handle_oav_grid_snapshot_triggered(self, doc) -> Sequence[ScanDataInfo]:
         assert self.ispyb_ids.data_collection_ids, "No current data collection"
         assert self.params, "ISPyB handler didn't receive parameters!"
+        assert self.data_collection_group_info, "No data collection group"
```
olliesilvester, line 213: Could: Use the constant mentioned earlier for this assertion

tests/unit_tests/common/external_interaction/xray_centre/test_ispyb_callback.py
```diff
@@ -326,3 +345,36 @@ def test_plan():
...
+    @patch(
+        "mx_bluesky.common.external_interaction.callbacks.xray_centre.ispyb_callback.GridscanISPyBCallback._handle_oav_grid_snapshot_triggered",
+    )
+    @patch(
+        "mx_bluesky.common.external_interaction.ispyb.ispyb_store.StoreInIspyb.update_deposition",
+    )
+    @patch(
+        "mx_bluesky.common.external_interaction.ispyb.ispyb_store.StoreInIspyb.update_data_collection_group_table",
+    )
+    def test_given_even_doc_before_start_doc_recieved_then_exception_raised(
```
olliesilvester, line 361:
```suggestion
    def test_given_event_doc_before_start_doc_recieved_then_exception_raised(
```

olliesilvester (approved):
Few minor things, but looks good :)

olliesilvester (changes requested):
I was about to merge this, but I noticed we don't have any tests which do the full process here, which is:
1. Fire a start ispyb
2. Two OAV snapshot events gives the first part of the group comment
3. Zocalo is triggered and the event from this adds the last part of the comment
4. Stop document clears the comment field

Also part 3 doesn't require part 2, the comment makes sense either way - I guess this will be useful for plans which don't do grid detect. Also would be good to have a test for that part. I'll do these though

olliesilvester (approved):

Issue #787, opened 2025-01-31
Sample error not going into database correctly


DominicOram:
Following https://github.com/DiamondLightSource/mx-bluesky/pull/625 we are now correctly putting `LOADED` into ispyb when we load a sample and `ERROR - beamline` if we get a general exception. However, we are not getting the `ERROR - sample` when we fail to diffract.

## Acceptance Criteria
- `ERROR - sample` is correctly put in database when we fail to diffract

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/836

PR #836, opened 2025-02-18
787 Update sample status field correctly

rtuck99:
Fixes
* #787 

1. `BLSampleStatus` should now be logged against the correct sample instead of only the first sample
2. BLSampleStatus updates should now only be triggered once on an exception instead of once for each level of bluesky run nesting it escapes.

### Checks for reviewer


shihab-dls (approved):
Changes and tests look comprehensive to me! Approving.

PR #843, opened 2025-02-20
Fix path to panda settings

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/826

1. Confirm new test fails on `main` and passes here

### Checks for reviewer


olliesilvester (approved):
Thanks!

PR #846, opened 2025-02-20
Add to docs on code reviews

olliesilvester:
As per recent slack thread, it would be good to let others know when a review is in process by assigning yourself the PR when you begin to review it. Added docs which say this

### Checks for reviewer


DominicOram (approved):
Great, thanks!

PR #852, opened 2025-02-21
Remove aperture change callback

### LINKED ISSUE #849: Remove aperture change callback
The aperture change callback was made so that we could tell GDA which aperture to select for rotation scans after doing an XRC. Given that we now do everything with the one plan this is no longer needed and so we should remove it.

## Acceptance Criteria
- Aperture change callback and associated tests are removed

DominicOram:
Fixes #849

1. Confirm tests still pass and all references to aperture change callback have been removed

### Checks for reviewer


rtuck99 (approved):
Can rebase and remove the redundant test file, approved

Issue #773, opened 2025-01-27
Read collection parameters from agamemnon


DominicOram:
To remove our dependence on GDA we should instead get the collection information from agamemnon (e.g. http://agamemnon.diamond.ac.uk/example/collect).

When we are triggered from GDA we should be able to read the same information from agamemnon and convert it into a parameter object like from GDA the confirm that they are the same.

## Acceptance Criteria
- When hyperion is triggered we also read from http://agamemnon.diamond.ac.uk/getnextcollect/i03
- The parameters from here are converted into a `LoadCentreCollect` object
- We compare this conversion with what GDA sent
- We still use the GDA provided parameters for the collection
- New issues spun out for any discrepancies or hard conversions

DominicOram:
From a first pass and comparing the two:
```
{
    "parameter_model_version": "5.0.0", -> fixed
    "beamline": "BL03S", -> fixed
    "det_dist_to_beam_converter_path": "tests/test_data/test_lookup_table.txt", -> fixed
    "insertion_prefix": "SR03S", -> fixed
    "visit": "cm31105-4", -> derive from collect.prefix or ask to add explicitly
    "detector_distance_mm": 255, -> equal to collect.collection.distance
    "sample_id": 12345, -> equal to collect.sample.id
    "sample_puck": 40, -> equal to collect.sample.container
    "sample_pin": 3, -> equal to collect.sample.position
    "robot_load_then_centre": {
        "storage_directory": "/tmp/dls/i03/data/2024/cm31105-4/auto/123458/xraycentring", -> derive from collect.prefix or ask to add explicitly
        "file_name": "robot_load_centring_file", -> created based on other params
        "comment": "Robot load and centre", -> fixed
        "exposure_time_s": 0.004, -> fixed for xrc
        "use_roi_mode": false, -> fixed for xrc
        "demand_energy_ev": 11100, -> derive from collect.sample.wavelength for now
        "run_number": 0, -> fixed
        "features": {"use_panda_for_gridscan": false, "compare_cpu_and_gpu_zocalo": true}, -> will need to use the config server
        "panda_runup_distance_mm": 0.17, -> fixed or config server
        "box_size_um": 20 -> fixed or config server
    },
    "multi_rotation_scan": {
        "comment": "Rotation", -> equal to collect.collection.comment
        "storage_directory": "/tmp/dls/i03/data/2024/cm31105-4/auto/123458/", -> equal to collect.prefix
[...]

PR #854, opened 2025-02-24
Get visit parameters from agamemnon

shihab-dls:
Part of #773 

These changes should parse visit-related parameters (i.e., fields of `WithVisit`) from Agamemnon JSON blob.

1. Check we correctly parse all visit parameters
2. Confirm tests pass, and cover edge cases (e.g., regex strictness)
3. Check error handling is appropriate

### Checks for reviewer


DominicOram (changes requested):
Thanks! So I think it would be good if we could get this sent to us explicitly, I've started https://diamondlightsource.slack.com/archives/G017HTC55M1/p1740413060948359 to discuss it. As discussed, I also think that it would be good if we, instead of updating the parameters coming in from GDA we make an entirely new set of params and compare them to what GDA is giving us.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -6,16 +6,41 @@
...
+    sample_puck: int | None
+    sample_pin: int | None
+    beamline: str | None
+    visit: str | None
+    det_dist_to_beam_converter_path: str | None
+    insertion_prefix: str | None
+    detector_distance_mm: int | None
+    parameter_model_version: str | None
+    robot_load_then_centre: RobotLoadThenCentre | None
+    multi_rotation_scan: MultiRotationScan | None
```
shihab-dls, line 43: I decided to add this temporarily to allow `None`. The idea is to populate this from Agamemnon as we finish up #773 related tickets for the comparison aspect of the issue.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -81,15 +106,46 @@ def get_next_instruction(beamline: str) -> dict:
...
+        # Log differences against GDA populated parameters
+        differences = []
+        for key, value in load_centre_collect_params.__dict__.items():
+            if value != getattr(parameters, key, None):
+                differences.append(key)
+        LOGGER.info(f"Differences found for: {differences}")
+    except (ValueError, KeyError) as e:
+        LOGGER.warning(f"Failed to compare parameters: {e}")
+    except Exception as e:
+        LOGGER.warning(f"Unexpected error occurred. Failed to compare parameters: {e}")
```
shihab-dls, line 141: Not sure about this error handling, but wanted to swallow the errors here instead of in `update_params_from_agamemnon`

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -99,6 +147,10 @@ def update_params_from_agamemnon(parameters: T) -> T:
                 # Before we do https://github.com/DiamondLightSource/mx-bluesky/issues/226
                 # this will give no snapshots but that's preferable
                 parameters.multi_rotation_scan.snapshot_omegas_deg = []
+            compare_params(agamemnon_params, parameters)
```
DominicOram, line 150: Could: I think it might be cleaner to have the `compare` as a separate step that is called from `main` as they are two different things that aren't necessarily related:
* Add the pin type in
* Compare the parameters

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Yh, i think swallowing them here is fine :)

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -148,11 +166,14 @@ def test_given_agamemnon_fails_when_update_parameters_called_then_parameters_unc
...
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.compare_params")
 @patch("mx_bluesky.hyperion.external_interaction.agamemnon.requests")
 def test_given_agamemnon_gives_single_pin_when_update_parameters_called_then_parameters_changed_to_single_pin(
-    mock_requests: MagicMock, load_centre_collect_params: LoadCentreCollect
+    mock_requests: MagicMock,
+    mock_compare_params: MagicMock,
+    load_centre_collect_params: LoadCentreCollect,
 ):
-    configure_mock_agamemnon(mock_requests, None)
+    configure_mock_agamemnon(mock_requests, "")
```
DominicOram, line 176: Should: I think what's returned from agamemnon in this case is `None` not `""`

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -81,15 +91,53 @@ def get_next_instruction(beamline: str) -> dict:
...
+
+    if match:
+        return (match.group("visit"), detector_distance)
+
+    raise ValueError(
+        f"Agamemnon prefix '{prefix}' does not match MX-General root structure"
+    )
+
+
+def populate_parameters_from_agamemnon(agamemnon_params):
```
DominicOram, line 112: Should: I think it would be good to have a test that covers this, you could either change some of your current tests to mock out `get_withvisit_parameters_from_agamemnon` instead of `get_withvisit_parameters_from_agamemnon` or make an entirely new test

DominicOram (changes requested):
Great, thanks. Couple of mostly quite minor suggestions

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -81,15 +92,59 @@ def get_next_instruction(beamline: str) -> dict:
...
 
-def get_pin_type_from_agamemnon(beamline: str) -> PinType:
-    params = get_next_instruction(beamline)
-    return _get_pin_type_from_agamemnon_parameters(params)
+def get_withvisit_parameters_from_agamemnon(parameters: dict) -> tuple:
+    try:
+        prefix = parameters["prefix"]
+        collection = parameters["collection"]
+        # Assuming distance is identical for multiple collections. Remove after https://github.com/DiamondLightSource/mx-bluesky/issues/773
+        detector_distance = collection[0]["distance"]
```
shihab-dls, line 100: Now taking the first collections `detector_distance`. Will add a comment to #773 to address this before closing. I can't see collections from http://agamemnon.diamond.ac.uk/getnextcollect/i03 (I assume due to shutdown), but I assume if one collection is provided then it also within an array (otherwise, would need to check before indexing).

tests/system_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -1,12 +1,30 @@
...
+
+
+def test_given_test_agamemnon_instruction_then_load_centre_collect_parameters_populated():
+    params = _get_parameters_from_url(AGAMEMNON_URL + "/example/collect")
+    load_centre_collect = populate_parameters_from_agamemnon(params)
+    difference = DeepDiff(
+        load_centre_collect,
+        EXPECTED_PARAMETERS,
+    )
+    assert not difference
```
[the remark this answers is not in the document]
DominicOram: I think loosened, I can do this now quickly.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Yes, it will be in a list so should be fine :)

DominicOram (approved):
Looks good, thanks. Just need to get the system test working, which I'll do now

PR #857, opened 2025-02-24
Enforce pytest.raises() is simple

### LINKED ISSUE #851: Enforce single line pytest.raises
We should enforce https://docs.astral.sh/ruff/rules/pytest-raises-with-multiple-statements/ in our CI to avoid occasions where we are missing an `assert` in a `pytest.raises` context

## Acceptance Criteria
- `PT012` is enforced in CI

shihab-dls:
Fixes #851 

This PR should amend all tests that use `with pytest.raises()..`, such that the context manager only contains a single line of code (i.e., the line that raises the expected exception). This follows the newly added ruff linting rule [PT012](https://docs.astral.sh/ruff/rules/pytest-raises-with-multiple-statements/). The main reason we are doing this is to avoid accidentally asserting within a `pytest.raises()` block, as this never runs, but looks like it does.

1. Check all instances of `with pytest.raises()..` accounted for.
2. Confirm tests still pass.

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -205,6 +205,7 @@ lint.select = [
     "I",   # isort - https://docs.astral.sh/ruff/rules/#isort-i
     "UP",  # pyupgrade - https://docs.astral.sh/ruff/rules/#pyupgrade-up
     "SLF", # self - https://docs.astral.sh/ruff/settings/#lintflake8-self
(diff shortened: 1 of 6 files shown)


DominicOram (approved):
Great, thank you! Minor nit, take it or leave it

PR #858, opened 2025-02-24
Enable the zocalo bounds check system test

rtuck99:
Enables the test in
* #761 
to actually be selected for running in the containerised system tests


DominicOram (approved):
:+1:

PR #870, opened 2025-02-26
Put full murko results in redis as pickled objects

### LINKED ISSUE #683: Murko: Store all the returned data from murko
Currently the redis_to_murko forwarder is only gathering the best click from murko to put in redis. We get back lots of other information like masks of where the crystal are. We should store these too.

## Acceptance Criteria
- All the data returned from murko is put in redis

DominicOram:
Fixes #683

Needed to use pickle instead of json as the murko results contain numpy arrays.

1. Confirm tests still pass

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
+++ b/src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
@@ -1,6 +1,7 @@
 import io
 import json
 import pickle
+from datetime import timedelta
 from typing import TypedDict
 
 import numpy as np
@@ -41,20 +42,10 @@ def send_to_murko_and_get_results(request: MurkoRequest) -> dict:
     return results
 
 
-def correlate_results_to_uuids(request: MurkoRequest, murko_results: dict) -> list:
-    results = []
-    uuids = request["prefix"]
-
-    width, height = get_image_size(request["to_predict"][0])
-
-    for uuid, prediction in zip(uuids, murko_results["descriptions"], strict=False):
-        coords = prediction["most_likely_click"]
-        y_coord = coords[0] * height
-        x_coord = coords[1] * width
-        results.append(
-            {"uuid": uuid, "x_pixel_coord": x_coord, "y_pixel_coord": y_coord}
-        )
-    return results
+def correlate_results_to_uuids(
+    request: MurkoRequest, murko_results: dict
+) -> list[tuple]:
+    return list(zip(request["prefix"], murko_results["descriptions"], strict=False))
 
 
 class BatchMurkoForwarder:
@@ -98,12 +89,12 @@ class BatchMurkoForwarder:
         results = correlate_results_to_uuids(request_arguments, predictions)
         self._send_murko_results_to_redis(sample_id, results)
 
-    def _send_murko_results_to_redis(self, sample_id: str, results: list):
-        for result in results:
-            self.redis_client.hset(
-                f"murko:{sample_id}:results", result["uuid"], json.dumps(result)
-            )
-        self.redis_client.publish("murko-results", json.dumps(results))
+    def _send_murko_results_to_redis(self, sample_id: str, results: list[tuple]):
+        for uuid, result in results:
+            redis_key = f"murko:{sample_id}:results"
+            self.redis_client.hset(redis_key, uuid, str(pickle.dumps(result)))
+            self.redis_client.expire(redis_key, timedelta(days=7))
+        self.redis_client.publish("murko-results", pickle.dumps(results))
 
     def add(self, sample_id: str, uuid: str, image: NDArray):
         """Add an image to the batch to send to murko."""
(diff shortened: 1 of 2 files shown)


src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -98,12 +89,12 @@ def _handle_batch_of_images(self, sample_id, images, uuids):
...
-            self.redis_client.hset(
-                f"murko:{sample_id}:results", result["uuid"], json.dumps(result)
-            )
-        self.redis_client.publish("murko-results", json.dumps(results))
+    def _send_murko_results_to_redis(self, sample_id: str, results: list[tuple]):
+        for uuid, result in results:
+            redis_key = f"murko:{sample_id}:results"
+            self.redis_client.hset(redis_key, uuid, str(pickle.dumps(result)))
+            self.redis_client.expire(redis_key, timedelta(days=7))
+        self.redis_client.publish("murko-results", pickle.dumps(results))
```
olliesilvester, line 97: How come we need to store everything in a redis hash and then publish all the results over separately?

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -41,20 +42,10 @@ def send_to_murko_and_get_results(request: MurkoRequest) -> dict:
...
-        y_coord = coords[0] * height
-        x_coord = coords[1] * width
-        results.append(
-            {"uuid": uuid, "x_pixel_coord": x_coord, "y_pixel_coord": y_coord}
-        )
-    return results
+def correlate_results_to_uuids(
+    request: MurkoRequest, murko_results: dict
+) -> list[tuple]:
+    return list(zip(request["prefix"], murko_results["descriptions"], strict=False))
```
olliesilvester, line 48: Nit: My static type checker reads this return value as `list[tuple[Unknown, ...]]`, which makes it hard to see what's going on afterwards - I would've expected it to be able to work out the type for `request["prefix"]` since you used a `TypedDict`

if we explicitly say the return type is, eg, `tuple[list[str], MurkoResults]`, this would help to see what's going on later in the code. Not super important though

olliesilvester (approved):
Thanks, just some comments regarding readability.

At some point, it would be quite useful to have an overall (system?) test for this just to see an example of what's going on and what the data looks like

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -41,20 +42,10 @@ def send_to_murko_and_get_results(request: MurkoRequest) -> dict:
...
-    for uuid, prediction in zip(uuids, murko_results["descriptions"], strict=False):
-        coords = prediction["most_likely_click"]
-        y_coord = coords[0] * height
-        x_coord = coords[1] * width
-        results.append(
-            {"uuid": uuid, "x_pixel_coord": x_coord, "y_pixel_coord": y_coord}
-        )
-    return results
+def correlate_results_to_uuids(
+    request: MurkoRequest, murko_results: dict
```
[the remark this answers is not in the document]
DominicOram: I have added a minor comment pointing to where there's some info but I think ultimately we should do https://github.com/MartinSavko/murko/issues/8 instead

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -98,12 +89,12 @@ def _handle_batch_of_images(self, sample_id, images, uuids):
...
         self._send_murko_results_to_redis(sample_id, results)
 
-    def _send_murko_results_to_redis(self, sample_id: str, results: list):
-        for result in results:
-            self.redis_client.hset(
-                f"murko:{sample_id}:results", result["uuid"], json.dumps(result)
-            )
-        self.redis_client.publish("murko-results", json.dumps(results))
+    def _send_murko_results_to_redis(self, sample_id: str, results: list[tuple]):
+        for uuid, result in results:
```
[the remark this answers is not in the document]
DominicOram: Yes, they're the same. We need to use `prefix` for when we send to murko but I don't really want to use it everywhere as I don't think it's a very good name. I've added a comment to `_correlate_results_to_uuids`, does this clarify it better?

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
DominicOram: Added a comment, let me know if it still doesn't makes sense

src/mx_bluesky/beamlines/i04/redis_to_murko_forwarder.py
```diff
@@ -86,10 +96,14 @@ def _handle_batch_of_images(self, sample_id, images, uuids):
             "prefix": uuids,
         }
         predictions = send_to_murko_and_get_results(request_arguments)
-        results = correlate_results_to_uuids(request_arguments, predictions)
+        results = _correlate_results_to_uuids(request_arguments, predictions)
```
olliesilvester, line 99: Nit: variable names `predictions` and `results` could be renamed to something like `results` and `results_with_uuids`

olliesilvester (approved):
This is clearer now, thanks!

PR #872, opened 2025-02-27
Create XBPM feedback preprocessor which can be inserted anywhere in a plan

### LINKED ISSUE #866: Make "xbpm and transmission for collection" optional in flyscan_xray_centre_plan
We currently have a decorator which we attach to our data collections/FGS which:
- Sets transmission to 1
- Checks if XBPM feedback is stable
- Pauses XBPM feedback and sets transmission to the desired value
- Does collection/scan
- Unpauses feedback and sets transmission to 1

Some beamlines using our xray-centre FGS will not want to do any of the above, so we need to make this an optional feature. Some extra constraints are:
- We only want the XBPM feedback to be paused for the minimal time - we should only be wrapping the actual collection part
- We want to avoid having a bunch of things like `if using XBPM wrapper then do XBPM decorator` in our common FGS plan

Proposed solution is to use a [bluesky preprocessor](https://nsls-ii.github.io/bluesky/plans.html#custom-preprocessors) to listen for a message which marks the collection beginning, and then inject the xbpm pausing logic. This preprocessor can then be optionally used at the entry points for the beamline-specific FGS plans

## Acceptance Criteria
- `@transmission_and_xbpm_feedback_for_collection_decorator` within the FGS is made optional 
- We aren't leaving the XBPM feedback paused for any longer than we were before

olliesilvester:
Fixes #866 

1. Check preprocessor does what it claims to do in its docstring - docstring for plan mutator will help to review this
2. Confirm tests pass

### Checks for reviewer


src/mx_bluesky/common/preprocessors/preprocessors.py
```diff
@@ -0,0 +1,63 @@
...
+            elif (
+                msg.command == "close_run"
+                and PlanNameConstants.GRIDSCAN_OUTER in msg.run
+            ):
+                return None, tail()
+        return None, None
+
+    # Ensure unpausing xbpm feedback occurs if there's an exception during the run
+    return (
+        yield from contingency_wrapper(
```
olliesilvester, line 52: Not 100% sure if it's better to unpause using `except_plan` or `final_plan` here. Final plan means we will always unpause at the end of the decorated plan. 

Except plan means we only do an extra unpause if there's an error. However, it means XBPM may not ever be unpaused if the relevant `close_run` is never seen

DominicOram: As discussed, it may be that we get a `close_run` even if there is an exception in the plan, in which case we don't need to put this in a wrapper at all. Can you try this and write a test against it?

src/mx_bluesky/common/device_setup_plans/xbpm_feedback.py
```diff
@@ -98,6 +99,21 @@ def _inner_plan():
     )
 
 
+def _check_and_pause_feedback_and_verify_undulator_gap(
```
DominicOram, line 102: Should: It's not clear to me what moving this out here has done other than just made things more descriptive? Which I'm all for and happy to keep, just wanted to check my understanding. However, my point with the undulator check was that all beamlines will need the undulator check but not all will need the pause/feedback check and this doesn't solve that. Happy to pull that into a new issue though?

DominicOram (changes requested):
Thanks, I think the preprocessor is a nice way to do this even if it's not pretty.

As discussed I think it would be good to make the wrapper generic by making `transmission_and_xbpm_feedback_for_collection_for_fgs_wrapper` optionally take a run name to start/stop on. If no name is given it should just do it on the first run it sees. I think you can then replace all uses of `@transmission_and_xbpm_feedback_for_collection_decorator` and `transmission_and_xbpm_feedback_for_collection_wrapper` with this new wrapper.

src/mx_bluesky/common/device_setup_plans/xbpm_feedback.py
olliesilvester: This was moved so that the decorator would still work without having to duplicate any functions. As we're planning on removing the original decorator now though, this change isn't needed.

Undulator gap was a separate thing yeah, I will address that in the PR that adds the VMXm implementation

src/mx_bluesky/common/preprocessors/preprocessors.py
olliesilvester: This works in _almost_ all cases, but It looks like if an old-fashioned `Ophyd` `Status` sets an exception on itself (eg using `set_exception`), then the plan terminates and, `close_run` is never seen.

It's a bit annoying, but given that the unpausing just does
`yield from bps.mv(xbpm_feedback.pause_feedback, Pause.RUN, attenuator, 1.0)`, I think we should probably just put it as an `except_plan`

src/mx_bluesky/common/device_setup_plans/xbpm_feedback.py
olliesilvester: Might be worth investigating if this is intended behaviour in Bluesky. Seems like it's unique to the `FailedStatus` exception

src/mx_bluesky/common/preprocessors/preprocessors.py
olliesilvester: Okay, I figured out why this happens, can discuss tomorrow

src/mx_bluesky/common/preprocessors/preprocessors.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if msg.command:
+            # If we specified a run key, did we encounter it
+            # If we didn't specify, then insert the plans and track the name of the run
+            if msg.command == "open_run" and (
+                run_key_to_wrap
+                and run_key_to_wrap is msg.run
+                or (not (run_key_to_wrap or _wrapped_plan_name))
+            ):
+                _wrapped_plan_name = msg.run if msg.run else "unnamed_run"
```
DominicOram, line 80: Must: I think you need a test on the correct stuff being inserted if the message has no run key. Currently I can change this line to `_wrapped_plan_name = msg.run` and all the tests pass

src/mx_bluesky/common/preprocessors/preprocessors.py
```diff
@@ -0,0 +1,108 @@
...
+        nonlocal _wrapped_plan_name
+
+        if msg.command:
+            # If we specified a run key, did we encounter it
+            # If we didn't specify, then insert the plans and track the name of the run
+            if msg.command == "open_run" and (
+                run_key_to_wrap
+                and run_key_to_wrap is msg.run
+                or (not (run_key_to_wrap or _wrapped_plan_name))
+            ):
```
DominicOram, line 79: Should: If you swap the order of these you can reduce it a bit:
```suggestion
            if msg.command == "open_run" and (
                not (run_key_to_wrap or _wrapped_plan_name)
                or run_key_to_wrap is msg.run
            ):
```

src/mx_bluesky/common/preprocessors/preprocessors.py
```diff
@@ -0,0 +1,108 @@
...
+                _wrapped_plan_name = msg.run if msg.run else "unnamed_run"
+
+                return head(msg), None
+
+            # Check if the run tracked from above was closed
+            elif msg.command == "close_run" and (
+                _wrapped_plan_name == "unnamed_run"
+                or (msg.run and _wrapped_plan_name and _wrapped_plan_name is msg.run)
+            ):
+                return None, tail()
```
DominicOram, line 89: Could: This may be a bit cleaner using pattern matching:

```suggestion
        match msg.command:
            case "open_run":
                # If we specified a run key, did we encounter it
                # If we didn't specify, then insert the plans and track the name of the run
                if (
                    not (run_key_to_wrap or _wrapped_plan_name)
                    or run_key_to_wrap is msg.run
                ):
                    _wrapped_plan_name = msg.run if msg.run else "unnamed_run"

                    return head(msg), None
            case "close_run":
                # Check if the run tracked from above was closed
                if _wrapped_plan_name == "unnamed_run" or (
                    msg.run and _wrapped_plan_name and _wrapped_plan_name is msg.run
                ):
                    return None, tail()
```

src/mx_bluesky/common/preprocessors/preprocessors.py
DominicOram: Nit: I would pull `unnamed_run` into a constant

src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -434,18 +433,16 @@ def multi_rotation_scan(
...
-        composite.attenuator,
-        composite.dcm,
-        parameters.transmission_frac,
-    )
     @bpp.finalize_decorator(lambda: _cleanup_plan(composite))
     def _multi_rotation_scan():
         for single_scan in parameters.single_rotation_scans:
 
-            @bpp.set_run_key_decorator("rotation_scan")
+            @transmission_and_xbpm_feedback_for_collection_decorator(
```
DominicOram, line 440: Must: Sorry, I feel like you may have asked me this and I gave you a different answer but... This now changes behaviour so that we are rechecking feedback between each collection. I think we want to only do it once on the full set of collections so it needs to be tied to `multi_rotation_scan`

src/mx_bluesky/hyperion/experiment_plans/set_energy_plan.py
```diff
@@ -33,6 +36,17 @@ class SetEnergyComposite:
...
+@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
+class XBPMWrapperComposite:
+    undulator: Undulator
+    xbpm_feedback: XBPMFeedback
+    attenuator: BinaryFilterAttenuator
+    dcm: DCM
+
+
+@bpp.set_run_key_decorator(PlanNameConstants.SET_ENERGY)
+@bpp.run_decorator()
```
DominicOram, line 49: Hmm.. I forgot that we do this on setting energy too. Having to open a run to set an energy isn't great :/ I think this is fine and a step in the right direction but we should revisit at some point (https://github.com/DiamondLightSource/mx-bluesky/issues/889)

tests/unit_tests/common/device_setup_plans/test_xbpm_feedback.py
```diff
@@ -0,0 +1,125 @@
...
+    mock = MagicMock()
+
+    @transmission_and_xbpm_feedback_for_collection_decorator(composite, 0.1)
+    @bpp.run_decorator()
+    def my_collection_plan():
+        mock()
+        yield from bps.null()
+
+    status = Status()
+    status.set_exception(Exception())  # error occurs before contingency wrapper?
```
DominicOram, line 94: Should: I don't really like question marks in comments like this. It just gives doubt to the later reader

tests/unit_tests/hyperion/experiment_plans/test_flyscan_xray_centre_plan.py
```diff
@@ -1010,3 +1013,115 @@ def test_run_gridscan_and_fetch_results_discards_results_below_threshold(
...
+        mock_wait,
+        mock_unpause_and_set_transmission: MagicMock,
+        mock_check_and_pause: MagicMock,
+        fake_fgs_composite: HyperionFlyScanXRayCentreComposite,
+        test_fgs_params: HyperionSpecifiedThreeDGridScan,
+        RE: RunEngine,
+    ):
+        RE(flyscan_xray_centre(fake_fgs_composite, test_fgs_params))
+        mock_check_and_pause.assert_called_once()
+        mock_unpause_and_set_transmission.assert_called_once()
```
DominicOram, line 1071: Could: I think this is covered by the test below so you could remove

DominicOram (changes requested):
Thanks, a few more comments, sorry. Also, we currently don't ever actually call it with no `run_key_to_wrap`, right? Is this because we have inner runs and so we end up with:
1. open outer run -> this pauses
2. open inner run
3. close inner run -> this unpauses
4. close outer run -> this should actually be where we unpause

I think we should probably think about this more in https://github.com/DiamondLightSource/mx-bluesky/issues/889 but it's ok for now. But can you add something in the docstring to say that you have to specify the run key if you have nested runs?

olliesilvester:
> Thanks, a few more comments, sorry. Also, we only call it with no `run_key_to_wrap` during `set_energy`, right? Is this because we have inner runs and so we end up with:
> 
>     1. open outer run -> this pauses
> 
>     2. open inner run
> 
>     3. close inner run -> this unpauses
> 
>     4. close outer run -> this should actually be where we unpause
> 

No, using it with inner runs should be fine. If you don't specify the run key, it pauses XBPM on the first run, and then marks the name of that run using `_wrapped_plan_name`, (which I should rename `_wrapped_run_name`), and then waits for that plan to unpause. Any inner plans that close before that won't unpause things

tests/unit_tests/common/device_setup_plans/test_xbpm_feedback.py
olliesilvester: Whoops, meant to remove that

DominicOram:
> > Thanks, a few more comments, sorry. Also, we only call it with no `run_key_to_wrap` during `set_energy`, right? Is this because we have inner runs and so we end up with:
> > ```
> > 1. open outer run -> this pauses
> > 
> > 2. open inner run
> > 
> > 3. close inner run -> this unpauses
> > 
> > 4. close outer run -> this should actually be where we unpause
> > ```
> 
> No, using it with inner runs should be fine. If you don't specify the run key, it pauses XBPM on the first run, and then marks the name of that run using `_wrapped_plan_name`, (which I should rename `_wrapped_run_name`), and then waits for that plan to unpause. Any inner plans that close before that won't unpause things. I'll adjust one of my tests to show this.
> 
> We can call it on `set_energy_plan` without a `run_key_to_wrap`, but we do still need a open a run

Ok, great, can we have some tests on this behaviour around inner functions then? i.e. that it doesn't unpause on the inner run.

If it always works on the outer run then I'm not sure I understand why we need to specify `run_key_to_wrap` in most of these places then as we're just specifying the outer run?

src/mx_bluesky/common/preprocessors/preprocessors.py
olliesilvester: `test_xbpm_preprocessor_wraps_one_run_only_if_no_run_specified` tests this now

tests/unit_tests/common/preprocessors/test_preprocessors.py
```diff
@@ -127,24 +129,66 @@ def except_plan():
...
     mock_pause.assert_called_once()
     mock_unpause_xbpm.assert_called_once()
+
+
+def test_xbpm_preprocessor_cant_unpause_on_wrong_run(
+    xbpm_and_transmission_wrapper_composite: XBPMAndTransmissionWrapperComposite,
+    RE: RunEngine,
+):
+    # Logic in the preprocessor relies on the assumption that Bluesky doesn't let us have
+    # multiple unnamed runs open, or multiple runs with the same name
```
DominicOram, line 163: Nit: I would put this in the preprocessor itself.

DominicOram (approved):
Great, thank you! Minor nitpick, take it or leave it.

tests/unit_tests/common/preprocessors/test_preprocessors.py
olliesilvester: I do already have this comment along the same lines in the preprocessor

```
# An exception is raised in the RunEngine if two unnamed runs are opened
# at the same time, so we are safe from unpausing on the wrong run
```

I put the other comment in the test as well since it's not very obvious what the test is doing otherwise

Issue #881, opened 2025-03-03
Roll back to bluesky 1.13


DominicOram:
Same as https://github.com/DiamondLightSource/dodal/issues/1082

## Acceptance Criteria
- `mx-bluesky` is on bluesky version 1.13

PR #882, opened 2025-03-03
Roll back bluesky version

DominicOram:
Fixes https://github.com/DiamondLightSource/mx-bluesky/issues/881

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/1083

1. Confirm tests still pass and we're on 1.13.0

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,6 @@ dependencies = [
     #
     "annotated_types",
     "caproto",
-    "confluent-kafka==2.8.0",
     "fastapi[all]",
     "flask-restful",
     "ispyb",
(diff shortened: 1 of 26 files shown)


pyproject.toml
```diff
@@ -18,7 +18,6 @@ dependencies = [
     #
     "annotated_types",
     "caproto",
-    "confluent-kafka==2.8.0",
```
olliesilvester, line 21: Maybe we should leave this pin in? Unless the latest `confluent-kafka` is working now

olliesilvester (approved):

pyproject.toml
DominicOram: I think that the `confluent-kafka` is a dependency of `bluesky` and `bluesky==1.13.0` has it pinned at `confluent-kafka==2.8.0` so we should be fine

Issue #860, opened 2025-02-24
Get WithSample parameters from Agamemnon


shihab-dls:
Part of #773 

We need to populate `WithSample` from Agamemnon. See #773 for more information.
 
## Acceptance Criteria
- `WithSample` is populated directly from Agamemnon.

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/890

PR #890, opened 2025-03-07
860 Decode sample information from Agamemnon response

rtuck99:
Fixes

* #860 

(remember to update `setup.cfg` with the dodal commit tag if you need it for tests to pass!)

The information in the `WithSample` mx-bluesky parameter mixin should now be retrieved from the Agamemnon request and included in the comparison with the GDA request

### Checks for reviewer


DominicOram (approved):
Great, thank you!

PR #893, opened 2025-03-10
Move undulator check to separate preprocessor

### LINKED ISSUE #891: Add undulator check as preprocessor
We now have put the xbpm feedback check in as preprocessor. Currently this code also checks the undulator gap before collection. However, many other beamlines would also like the undulator gap to be checked but not have the feedback check. We should pull this out into a new preprocessor so that it can be used separately. This is also useful outside of MX (https://diamondlightsource.slack.com/archives/CJ6127Z8T/p1741600816914069) so we should think about putting it in dodal

## Acceptance Criteria
- The undulator gap is pulled into a new preprocessor

olliesilvester:
Fixes #891 

Link to dodal PR (if required): (https://github.com/DiamondLightSource/dodal/pull/1099)

1. Confirm undulator preprocessor works
2. Confirm we are using this preprocessor before all data collections, as before
3. Confirm tests pass with the dodal change and that new tests cover everything

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd == 1.9.0",
     "ophyd-async >= 0.9.0a2",
     "bluesky == 1.13",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@b119d38c70c67925a998f9bcfc09689db9043eaf",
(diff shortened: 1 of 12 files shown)


DominicOram (approved):
Great, thank you!

Issue #151, opened 2024-08-12
I24 SSX; Replace EDMs with web gui


dperl-dls:
Some preliminary/exploratory work is being done in https://github.com/dperl-dls/i24-ssx-gui

This will require a number of things to work:

- [ ] Set up https://github.com/DiamondLightSource/coniql on the beamline such that it only accepts mutate requests from the beamline network
- [ ] Prepare (as generic and reusable as practicable) React components to match the features in the EDMs
  - [ ] generic buttons etc.
  - [ ] clickable OAV stream
  - [ ] ...
- [ ] Prepare (as generic and reusable as practicable) React components to replace the state currently managed by the general purpose PVs

PR #907, opened 2025-03-11
I24 serial: update web gui plans

noemifrisina:
Part of #151 

Adds/updates some more plans for the web gui.
Needed by [MX-daq-ui#29](https://github.com/DiamondLightSource/mx-daq-ui/pull/29)

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


tests/unit_tests/beamlines/i24/serial/parameters/test_utils.py
```diff
@@ -62,3 +62,21 @@ def test_get_chip_map(fake_caget: MagicMock):
...
     assert len(chip_map) == 2
     assert chip_map == [1, 3]
+
+
+def test_get_chip_format_for_custom_from_web_ui_fails_if_no_data_passed():
+    with pytest.raises(ValueError):
+        get_chip_format(ChipType.Custom, None, "web")
+
+
+def test_get_chip_format_for_custom_from_web_ui():
```
olliesilvester, line 72: I think this test test along with `test_get_chip_format_for_custom_chips` and `test_get_chip_format_for_oxford_chips` could be combined with a big `pytest.mark.parametrize`

tests/unit_tests/beamlines/i24/serial/web_gui/test_general_plans.py
```diff
@@ -48,13 +54,46 @@ def fake_generator(value):
     fake_mv.assert_called_with(ANY, 0.0125, ANY, 0.025)
 
 
+@patch("mx_bluesky.beamlines.i24.serial.web_gui_plans.general_plans.get_detector_type")
+def test_gui_set_parameters_raises_error_for_empty_map(mock_det_type, RE):
```
olliesilvester, line 58: I can't see any tests for `gui_set_parameters` which handle the case where it doesn't error - although codecov isn't complaining so maybe I've missed something?

olliesilvester (changes requested):
Thanks, in general looks good but I have some questions and suggestions (feel free to ignore any of the suggestions though, there's no 'musts')

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
```diff
@@ -13,7 +13,11 @@ class EmptyMapError(Exception):
     pass
 
 
-def get_chip_format(chip_type: ChipType) -> ChipDescription:
+def get_chip_format(
+    chip_type: ChipType,
+    format: list[int | float] | None = None,
```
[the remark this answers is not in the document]
olliesilvester: Having looked at more of the code I'm not certain that it does always need to be length 4 though

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
```diff
@@ -33,10 +37,19 @@ def get_chip_format(chip_type: ChipType) -> ChipDescription:
...
-            defaults["y_step_size"] = float(caget(pv.me14e_gp99))
+            if origin == "edm":
+                defaults["x_num_steps"] = int(caget(pv.me14e_gp6))
+                defaults["y_num_steps"] = int(caget(pv.me14e_gp7))
+                defaults["x_step_size"] = float(caget(pv.me14e_gp8))
+                defaults["y_step_size"] = float(caget(pv.me14e_gp99))
+            else:
+                # NOTE Test for WEB GUI
+                if not format:
+                    raise ValueError("Format for custom chip not passed")
```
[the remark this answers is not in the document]
noemifrisina: For a custom chip it makes little sense to have a default as whatever it is chances are it will be completely different from what the users bring in. Ideally the UI would error before getting here in the case of no format set, but it doesn't right now so better to stop it here than to start a collection with a wrong chi size

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
```diff
@@ -96,14 +124,14 @@ def gui_set_parameters(
...
+        "map_type": mapping,
+        "chip_map": chip_map,
         "pump_repeat": PumpProbeSetting[pump_probe],  # pump_repeat,
         "laser_dwell_s": laser_dwell,
         "laser_delay_s": laser_delay,
         "checker_pattern": checker_pattern,
         "pre_pump_exposure_s": pre_pump,
     }
-    print(FixedTargetParameters(**params))
     # This will then run the run_fixed_target plan
```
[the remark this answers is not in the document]
noemifrisina: It will when the UI collection panel is finished and merged... for the purpose of testing while it's a WIP it was good to just have a confirmation that the bluesky plan got the correct parameters - which it does. Running the actual plan will be a separate PR because that involves changes on both sides from how things are now

tests/unit_tests/beamlines/i24/serial/web_gui/test_general_plans.py
noemifrisina: You haven't missed anything, it's just that it's kind of difficult - and not super useful - to test the non-erroring bit while the function doesn't do anything but print. I can add a test that checks the printed parameters but that will change as soon was the rest of the functionality is added (ie run the collection plan)

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
```diff
@@ -84,7 +102,17 @@ def gui_set_parameters(
     # NOTE still a work in progress, adding to it as the ui grows
     detector_stage = i24.detector_motion()
     det_type = yield from get_detector_type(detector_stage)
-    chip_params = get_chip_format(ChipType[chip_type])
+    _format = chip_format if ChipType[chip_type] is ChipType.Custom else None
+    chip_params = get_chip_format(ChipType[chip_type], _format)
+    if ChipType[chip_type] in [ChipType.Oxford, ChipType.OxfordInner]:
+        mapping = MappingType.Lite if map_type == "Lite" else MappingType.NoMap
+        if mapping is MappingType.Lite and len(chip_format) == 0:
+            # this logic should go in the gui with error message.
```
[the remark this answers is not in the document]
noemifrisina: Not yet, need to work out what exactly it's possible to do on the ui side and what here first...

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
```diff
@@ -33,10 +37,19 @@ def get_chip_format(chip_type: ChipType) -> ChipDescription:
...
+                defaults["x_num_steps"] = format[0]
+                defaults["y_num_steps"] = format[1]
+                defaults["x_step_size"] = format[2]
+                defaults["y_step_size"] = format[3]
```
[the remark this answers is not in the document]
noemifrisina: Yes, that would work both list and tuple, but `format` might be None which will cause linting to rightly fail - tried a type ignore to make it happy and it still failed because that made the ValueError seem unreachable. I can't really say what the right solution is here...

tests/unit_tests/beamlines/i24/serial/parameters/test_utils.py
noemifrisina: Mmm I think I'd rather keep them separate mostly for readability, as they all require slightly different patches and arguments - we'd end up with a few patches that are only needed in one of the tests (eg. the caget one)

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
```diff
@@ -13,8 +14,27 @@ class EmptyMapError(Exception):
...
-def get_chip_format(chip_type: ChipType) -> ChipDescription:
-    """Default parameter values."""
+def get_chip_format(
+    chip_type: ChipType,
+    format: Sequence[int | float] | None = None,
+    origin: Literal["edm", "web"] = "edm",
+) -> ChipDescription:
+    """Get the default parameter values for the requested chip type.
+
+    For an Oxford-type chip, the default values are hard coded as the dimensions are \
```
olliesilvester, line 24: You don't need backslashes for docstrings when you're inside the `"""` bit

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
```diff
@@ -75,16 +96,54 @@ def gui_set_parameters(
...
     pump_probe: str,
     laser_dwell: float,
     laser_delay: float,
     pre_pump: float,
 ) -> MsgGenerator:
+    """Set the parameter model for the data collection.
+
+    Args:
+        sub_dir (str): subdirectory of the visit to write data in.
+        chip_name (str): a name identifying the current chip collection, will be used \
```
olliesilvester, line 111: See other comment on backslashes

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
```diff
@@ -13,8 +14,27 @@ class EmptyMapError(Exception):
...
+    always the same. For a Custom chip instead, the number of steps and step size in \
+    each direction must be entered throught the GUI - web or edm. If the collection is \
```
olliesilvester, line 26:
```suggestion
    each direction must be entered through the GUI - web or edm. If the collection is
```

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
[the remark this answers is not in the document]
olliesilvester: If you do `assert format("Format for custom chip not passed")` instead of the if statement it might fix linting? No big deal if not though, don't let it slow you down

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
```diff
@@ -75,16 +96,54 @@ def gui_set_parameters(
...
+            lite collection.
+        checker_pattern (bool): whether checker_pattern is turned on, ie. only every \
+            other window in a block gets collected
+        pump_probe (str): pump probe setting.
+        laser_dwell (float): laser exposure time for pump probe collections, in s.
+        laser_delay (float): delay between laser exposure and collection, in s.
+        pre_pump (float): pre-pump exposure time for a pump probe short2 collection, \
+            ie a pump-in-probe where the collection starts during the pump.
+    """
     # NOTE still a work in progress, adding to it as the ui grows
```
olliesilvester, line 133: Could we link to an issue here if there is one?

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
```diff
@@ -75,16 +96,54 @@ def gui_set_parameters(
...
     det_type = yield from get_detector_type(detector_stage)
-    chip_params = get_chip_format(ChipType[chip_type])
+    _format = chip_format if ChipType[chip_type] is ChipType.Custom else None
+    chip_params = get_chip_format(ChipType[chip_type], _format)
+    if ChipType[chip_type] in [ChipType.Oxford, ChipType.OxfordInner]:
+        mapping = MappingType.Lite if map_type == "Lite" else MappingType.NoMap
+        if mapping is MappingType.Lite and len(chip_format) == 0:
+            # this logic should go in the gui with error message.
+            raise EmptyMapError("No blocks chosen")
+        chip_map = chip_format
```
olliesilvester, line 143: Codecov is complaining that lines 143 and 146 aren't covered by tests

src/mx_bluesky/beamlines/i24/serial/web_gui_plans/general_plans.py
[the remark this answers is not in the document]
olliesilvester: okay, makes sense. Comment is a bit misleading though, maybe 
```suggestion
    # run_fixed_target plan can be run after parameters are set
```

tests/unit_tests/beamlines/i24/serial/web_gui/test_general_plans.py
```diff
@@ -1,15 +1,26 @@
...
     gui_gonio_move_on_click,
+    gui_move_backlight,
     gui_move_detector,
+    gui_set_parameters,
     gui_sleep,
     gui_stage_move_on_click,
 )
 
 
+def fake_generator(value):
```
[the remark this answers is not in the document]
olliesilvester: I still suggest moving this to a fixture which all tests can see, as we have 6 of the same functions in i24 tests

olliesilvester (changes requested):
Sorry, a few more small comments

src/mx_bluesky/beamlines/i24/serial/parameters/utils.py
noemifrisina: Well, just for the docstring no. But for API docs built from docstrings, the backslash avoids readability issues. 
To be fair, we don't have API docs in here so may as well remove it

olliesilvester: Ah fair enough I didn't realise that, ignore my comment there then!

olliesilvester (approved):

PR #911, opened 2025-03-13
Ignore sqlalchemy deprecation warning in CI

noemifrisina:
Needed until there's a fix for  [ispyb-api#233](https://github.com/DiamondLightSource/ispyb-api/issues/233)


DominicOram (approved):
Great, thanks!

Issue #916, opened 2025-03-14
Test sometimes takes longer than a second.


adaudon:
Tests are supposed to take less than 1 second to pass. This test usually runs under 1 second, however it occasionally runs a little longer than 1 second. Doesn't appear often but does appear from time to time. Might be worth investigating?

`FAILED tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py::test_when_full_grid_scan_run_then_parameters_sent_to_fgs_as_expected - Failed: Timeout >1.0s`

olliesilvester:
A simple single unit test now takes a few seconds to complete when running locally. Running a pytest in a fresh repo with no setup/teardown is pretty instant. 

We should try to streamline the bits of code which run for every test. We have a bunch of fixtures which are used always, as well as logic in `pytest_runtest_setup` and `pytest_runtest_teardown`

PR #921, opened 2025-03-18
663 Pytest and tox CI timeout added for 2 seconds

### LINKED ISSUE #663: Add CI job that fails PRs if tests are taking >1s
All unit tests should be fast running and so not take >1s each. If they are taking longer it is most likely due to an issue with poor mocking. To make sure we enforce this we should add a CI check on a PR that confirms that no tests individually took too long.

The durations can be found with `pytest -m "not s03" --durations=0` but this isn't very machine readable so it may be worth a look around to see if we can get them out in a nicer way

We should use the approach discussed in https://github.com/bluesky/ophyd-async/issues/729#issue-2780647882. Then feed it back into the copier template

## Acceptance Criteria
- CI blocks PRs where an individual test is taking longer than 1s

adaudon:
Fixes #663 

1. Run pytest -m "not dlstbx" in root directory of repo.
2. Push code to git hub to check work flow.

### Checks for reviewer

## Notes

Timeout changed to 2s instead of 1s. Following tests pass locally but fail on GitHub workflows when timeout set to 1s instead of 2s.
- tests/unit_tests/hyperion/experiment_plans/test_robot_load_and_change_energy.py::test_given_smargon_disabled_for_longer_than_timeout_when_plan_run_then_throws_exception
- tests/unit_tests/hyperion/experiment_plans/test_robot_load_and_change_energy.py::test_when_prepare_for_robot_load_called_then_moves_as_expected
- tests/unit_tests/hyperion/experiment_plans/test_multi_rotation_scan_plan.py::test_full_multi_rotation_plan_nexus_files_written_correctly[False]
- tests/unit_tests/hyperion/experiment_plans/test_robot_load_then_centre.py::test_box_size_passed_through_to_gridscan
- tests/unit_tests/hyperion/experiment_plans/test_grid_detect_then_xray_centre_plan.py::test_when_full_grid_scan_run_then_parameters_sent_to_fgs_as_expected

See follow up issue https://github.com/DiamondLightSource/mx-bluesky/issues/916


DominicOram (changes requested):
Thanks! I think I like the suggestion on https://github.com/bluesky/ophyd-async/issues/729#issue-2780647882 to mark tests that are longer than expected with the decorator, could you reduce the time down to 1 second again and mark the longer ones with the decorator? I think the failing tests are unrelated, see https://github.com/DiamondLightSource/mx-bluesky/issues/941

PR #850, opened 2025-02-21
Trigger Zocalo and turn on feedback after all collections in MSP

DominicOram:
Fixes #826 and fixes #848

Note, this is probably a breaking change for single rotation scans but given https://github.com/DiamondLightSource/mx-bluesky/issues/847 I think that's ok.

### Instructions to reviewer on how to test:

1. Confirm new tests make sense and pass
2. Confirm new docs make sense
3. Confirm all other changes in #826 have been solved or have been spun into new issues

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

Issue #848, opened 2025-02-20
Multi rotations: Wait for unstage to happen before changing transmission


DominicOram:
Until we finish unstaging the eiger we may still be taking data. Therefore we should make sure not to change transmission before then. There is a test for doing this with single rotations (`test_rotation_scan_does_not_change_transmission_back_until_after_data_collected`) but not for multiple rotations. If there was it would fail

## Acceptance Criteria
- We do not change transmission until the eiger has finished unstaging on multiple rotation scans
- There is a test for this

PR #935, opened 2025-03-20
(#847) remove single rotation plan

### LINKED ISSUE #847: Hyperion: Remove single rotation scan plan
The single rotation plan is functionally equivalent to calling a single rotation plan but with a list of one rotations. To save on complexity we should remove the single plan and instead use the multiple plan. It should be up to the client to wrap this in a list of one.

Steps to this are:
* Remove `rotation_scan` plan
* Find any references and remove those or change to multi-rotation
* For unit tests switch them over to use the multi rotation plan
* Once done the above have a chat and maybe rename `multi_rotation_plan` -> `rotation_plan`

## Acceptance Criteria
- There is no longer a single rotation plan
- The endpoints are also removed

jacob720:
Fixes #847 
Removes the `rotation_scan` plan, as `multi_rotation_plan` can be used with just one rotation. One test is failing, which is because of this [issue](https://github.com/DiamondLightSource/mx-bluesky/issues/848) and should be fixed in this [PR](https://github.com/DiamondLightSource/mx-bluesky/pull/850).

1. Confirm `rotation_scan` does not exist.
2. Confirm `multi_rotation_scan` can be used to do a scan with a single rotation
3. Check all tests pass other than `test_rotation_scan_does_not_change_transmission_back_until_after_data_collected`.

Some thoughts:
- Could change the name of `multi_rotation_scan` to `rotation_scan`.
- Could combine the tests in `test_rotation_scan_plan.py` and `test_multi_rotation_scan_plan.py` into one file.

### Checks for reviewer


DominicOram (approved):
This is great, thanks. Can you do the merge and confirm the tests pass. Then I think we should do your suggested rename + combining of tests but can do in a separate PR to be less messy

jacob720:
Couple more tests failing now, I'm looking into it

DominicOram:
I fixed the test by adding the `BeamDrawingCallback`on to the multi rotation as well. I think this is correct but we may need to check with @rtuck99 when he's back.

Issue #646, opened 2024-11-13
 I24 serial: cs_maker should wait for the motors to be finished moving before setting the next pmac_string


noemifrisina:
The `cs_maker` function callse `trigger` on the pmac string to home/move the motors multiple times in a row. A sleep has now been added in between to allow to the motors to finish moving but we should actually check that they've stopped. 

Maybe this could be done in the dodal device?

PR #942, opened 2025-03-21
925 use bps not time for sleep

### LINKED ISSUE #925: Use `bps.sleep` rather than `time.sleep`
There are a number of places (particularly in the I24 SSX code) where we use `sleep` rather than `bps.sleep`.

This is bad because it means we're blocking the run RE from doing other work, such as callbacks, changing things on devices.

We probably want to generically stop this, we should look into using something like https://docs.astral.sh/ruff/rules/banned-api/ to not allow you to import `time.sleep`

## Acceptance Criteria
- `sleep` is replaced with `bps.sleep`
- CI stops people importing `time.sleep` - ideally with a message explaining why
- There is a way to overrule the CI check for specific lines

jacob720:
Fixes #925 

1. Add in a time.sleep, try to commit
2. Confirm ruff doesn't let you

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
+++ b/src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
@@ -6,10 +6,8 @@ This version changed to python3 March2020 by RLO
 import json
 import re
 import sys
-import time
 from pathlib import Path
 from pprint import pformat
-from time import sleep
 
 import bluesky.plan_stubs as bps
 import numpy as np
@@ -100,7 +98,7 @@ def initialise_stages(
 
     caput(pv.pilat_cbftemplate, 0)
 
-    sleep(0.1)
+    bps.sleep(0.1)
     SSX_LOGGER.info("Clearing General Purpose PVs 1-120")
     for i in range(4, 120):
         if i == 100:
@@ -259,7 +257,7 @@ def upload_chip_map_to_geobrick(pmac: PMAC, chip_map: list[int]) -> MsgGenerator
         SSX_LOGGER.debug(f"Set {pvar_str} for block {block}")
         yield from bps.abs_set(pmac.pmac_string, pvar_str, wait=True)
         # Wait for PMAC to be done processing PVAR string
-        sleep(0.02)
+        bps.sleep(0.02)
     SSX_LOGGER.debug("Upload parameters done.")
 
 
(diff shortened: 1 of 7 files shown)


src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -832,16 +830,16 @@ def check_dir(val):
     SSX_LOGGER.info(f"{sqfact1:1.4f} \n {sqfact2:1.4f} \n {sqfact3:1.4f}")
     SSX_LOGGER.debug("Long wait, please be patient")
     yield from bps.trigger(pmac.to_xyz_zero)
-    sleep(2.5)
+    bps.sleep(2.5)
```
noemifrisina, line 833: In most cases in I24 ssx code changing to bps.sleep should be okay. This particular plan instead poses a few issues in the sense that we need something that blocks here and doesn't allow the next plan to run, because this would send another pmac string too early, stopping the motor move started by bps.trigger and messing up the coordinate system creation... 

There may be a couple other places with this particular issue, but cs_maker is a really big one and we need to keep this blocking until #646  is fixed.

noemifrisina: This is the one use case where we'd need a noqa-like thing to allow time.sleep

jacob720: That's good to know, I've changed back the sleeps in this plan to `time.sleep`. Are there any other plans where you'd like me to do the same?

noemifrisina (commented):
Aside from my previous comment, this looks good once tests are fixed thanks. 

I'd like another opinion on this, I think we could merge it keeping it mind that it would break i24 - meaning until the other issue is fixed we can't update the version on that beamline. I don't really like the option but it might be the needed prompt to finally fix the issue...

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
noemifrisina: I think that's the big problematic one thanks. The only other one I'm not 100% sure about may actually be fine, will just need some beamline testing. As the PR is actually to ban time.sleep, do we already have a way of allowing it in one use case?

callumforrester (dismissed):
All instances of `bps.sleep` should be `yield from bps.sleep` (assuming they are actually inside plans, they are a no-op outisde)

noemifrisina (commented):
Looks like the tests for the i24 plans that used to use `time.sleep` need to be updated too (it was mocked to avoid having very slow tests)

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
DominicOram: I'm not sure I understand this usecase. Do you mean you need to block before running `set_pmac_strings_for_cs`? A `yield from bps.sleep` will do that

noemifrisina: It needs to block in between everything that sends a command to set the pmac_string, so essentially this whole block 

```python
 yield from bps.trigger(pmac.to_xyz_zero)
    time.sleep(2.5)
    yield from set_pmac_strings_for_cs(pmac, {"cs1": cs1, "cs2": cs2, "cs3": cs3})
    yield from bps.trigger(pmac.to_xyz_zero)
    time.sleep(2.5)
    yield from bps.trigger(pmac.home, wait=True)
    time.sleep(2.5)
    SSX_LOGGER.debug(f"Chip_type is {chip_type}")
    if chip_type == 0:
        yield from bps.abs_set(pmac.pmac_string, "!x0.4y0.4", wait=True)
        time.sleep(2.5)
        yield from bps.trigger(pmac.home, wait=True)
    else:
        yield from bps.trigger(pmac.home, wait=True)
```
I just put the comment on the first trigger.

The problem is that all triggers will send a pmac_string that moves the motor but `bps.trigger` doesn't know that and returns as completed once the pmac_string is set. So the next pmac_string will happen while the motors are moving, stopping them, which causes the final coordinate system to be wrong. The fix, since we're already overriding trigger in the PMAC device, would be to monitor the motors in there... only we haven't gotten around it that yet.

DominicOram: As discussed offline, I think we should be able to use `bps.sleep` here. @jacob720 can you make it use `bps.sleep` again (sorry) and I will find some time to test on the beamline

DominicOram (approved):
Great, thank you! Happy to merge despite `codecov` issues as that's outside of scope

PR #946, opened 2025-03-24
Fill in more data from agamemnon

DominicOram:
Fixes #945

### Instructions to reviewer on how to test:

1. Confirm tests pass, including agamemnon system tests

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

PR #937, opened 2025-03-20
Extracted and converted wavelength into demand_energy

adaudon:
Added demand_energy_ev to agamemnon collection

Fixes #862 

### Instructions to reviewer on how to test:

1. Get output of parameters from the populate_parameters_from_agamemnon.
2. Test test_given_test_agamemnon_instruction_then_returns_none_loop_type() and test_given_test_agamemnon_instruction_then_load_centre_collect_parameters_populated()
3. Confirm demand_energy_ev is shown.

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

Issue #864, opened 2025-02-24
Get scan_width_deg from Agamemnon


shihab-dls:
Part of #773 

We need to calculated `scan_width_deg` from `collect.collection.number_of_images` & `omega_increment` (i.e., `scan_width_deg = number_of_images * omega_increment`)

## Acceptance Criteria
- `scan_width_deg` is calculated and set in `LoadCentreCollect`

DominicOram:
For the initial comparison we should use the first rotation that agamemnon gives, this will work for ligand collections but we need to something more for native. This will be a follow up issue.

PR #947, opened 2025-03-25
Create robot load and centre params from agamemnon instruction

### LINKED ISSUE #861: Get RobotLoadThenCentre parameters from Agamemnon
Part of #773 

We need to populate `RobotLoadThenCentre` from Agamemnon.
 
## Acceptance Criteria
- `RobotLoadThenCentre` fields are directly populated from Agamemnon

DominicOram:
Fixes #861

Built on top of https://github.com/DiamondLightSource/mx-bluesky/pull/946 so review that first.

Will need energy added from https://github.com/DiamondLightSource/mx-bluesky/pull/937

This also changes the paths for xraycentring data and snapshots, I think we should roll with it and see what issues it causes

1. Confirm tests pass, including system tests

### Checks for reviewer


DominicOram:
> Looks ok as far as it goes. However I wonder where are we ultimately going with this? We now seem to have duplicate values populated for things like `sample_puck` etc in the model.
> 
> `LoadCentreCollect` already has logic for populating these values from the minimal set supplied in the json. If ultimately we are going to be using this to populate LoadCentreCollect to call the experiment plan, is there any sense in populating both sets, why not just let `model_validate` do all the work?

There was a couple reasons for this:
* It would be hard to do the stepping stone of just doing the robot load first if we use the `LoadCentreCollect` now, we would also need to add in everything for the rotations
* This gives us typing etc. the alternative would be to construct a json object that describes part of the `RobotLoadThenCentre` and pass that in so we would only get runtime checking?

We can revisit once we've done https://github.com/DiamondLightSource/mx-bluesky/issues/864

PR #949, opened 2025-03-25
Fixes for dodal 1085 breakage, pydantic deprecation warnings

rtuck99:
Fixes breakages caused by upstream changes:

* DiamondLightSource/dodal#1085
* bluesky/ophyd-async#838

plus a test failure in our own code caused by our use of `DeepDiff` on pydantic models which also triggers the problematic deprecation warnings

N.b. pydantic 2.11 is currently in pre-release but I have it accidentally installed due to having to revert some other package in my venv.

1. Tests pass

### DIFF
--- a/src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
@@ -1,10 +1,12 @@
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
+from typing import cast
 
 import bluesky.plan_stubs as bps
 from bluesky.utils import MsgGenerator
 from dodal.common.beamlines.beamline_utils import get_path_provider
+from dodal.common.types import UpdatingPathProvider
 from dodal.devices.fast_grid_scan import PandAGridScanParams
 from dodal.devices.smargon import Smargon
 from ophyd_async.fastcs.panda import (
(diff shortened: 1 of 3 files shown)


src/mx_bluesky/hyperion/device_setup_plans/setup_panda.py
```diff
@@ -222,6 +224,8 @@ def set_panda_directory(panda_directory: Path) -> MsgGenerator:
     suffix = datetime.now().strftime("_%Y%m%d%H%M%S")
 
     async def set_panda_dir():
-        await get_path_provider().update(directory=panda_directory, suffix=suffix)
+        await cast(UpdatingPathProvider, get_path_provider()).update(
```
olliesilvester, line 227: Fine for this PR, but it seems bad/annoying to have to cast every time we use `get_path_provider()`.

tests/system_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -27,8 +29,14 @@ def test_given_test_agamemnon_instruction_then_returns_none_loop_type():
...
-    difference = DeepDiff(
-        load_centre_collect,
-        EXPECTED_PARAMETERS,
-    )
+    difference = True
+    with warnings.catch_warnings():
+        # Suppress warning exceptions due to Pydantic 2.11 deprecation warnings
+        warnings.filterwarnings(
+            "ignore", "Accessing this attribute on the instance is deprecated"
+        )
```
DominicOram, line 37: I don't think we need this if we have the exception in the `pyproject.toml`. I will make a new PR to remove this and put an issue in with `deepdiff`

Issue #226, opened 2024-07-18
Take one snapshot for multiple rotations


DominicOram:
When doing multiple rotations in one arm we do not need to keep moving and taking snapshots at each location. Instead, after we have worked out where each rotation will be we take the snapshots that we took for the grid scan and then we draw the different beam locations on it for each collection. i.e. each collection will only have one beam position marked but it will be the same underlying snapshot image that they are drawn on.

## Acceptance Criteria
* When doing multiple rotations the beam positions are drawn on the grid snapshots and these are put in the correct place in ispyb

DominicOram:
We want to optionally use this behaviour on single pins too.

DominicOram:
Discussed architecture with @rtuck99 and @olliesilvester. We came up with the following. This should mean that the current behaviour can still be achieved with the same callbacks/devices and only a change to plans.
![Image](https://github.com/user-attachments/assets/d37245b4-6339-417a-b49b-ee190cb612eb)
Other things discussed:
* The snapshots with crosshairs will not be re-orientated so that the crosshair is in the middle. This will mean that the crosshair will jump around in the y when flipping between different omegas in synchweb
* There is a middle ground where we take all the snapshots once before we do all the rotations, should discuss with the scientists if we want this option but the above architecture should deal with it ok
* We should split this issue into making the callback do the drawing but leaving the plans basically unchanged (i.e. the left hand side) then doing the full thing

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/950

PR #950, opened 2025-03-25
226 one snapshot multiple rotations

rtuck99:
Fixes 

* #226 

This PR does the following:
* Introduces a new parameter `use_grid_snapshots` when specifying snapshots in the request parameters for rotations
* When the parameter is `False` (the default) then rotation snapshots are captured as normal using the OAV
* When the parameter is `True` then
   * The only valid values for `snapshot_omegas_deg` are 0 and 270 degrees
   * Rotation snapshots are not taken, instead they are generated from the rotation snapshots by rendering the crosshair on top of the 0 and -90 degree gridscan snapshots in the appropriate position
   * Rotation snapshots for non-zero chi will be rendered using the zero-chi snapshots

### Checks for reviewer


src/mx_bluesky/hyperion/external_interaction/callbacks/snapshot_callback.py
```diff
@@ -39,69 +58,185 @@ class BeamDrawingCallback(PlanReactiveCallback):
...
+                (current_sample_pos_mm[1] - snapshot_info.sample_pos_mm[1]),
+                (current_sample_pos_mm[2] - snapshot_info.sample_pos_mm[2]),
+            ),
+            snapshot_info.omega,
+        )
+
+    def _project_xyz_to_xy(
+        self, xyz: tuple[float, float, float], omega_deg: float
+    ) -> tuple[float, float]:
+        return (xyz[0], xyz[1] * cos(omega_deg) + xyz[2] * sin(omega_deg))
```
DominicOram, line 209: FYI: https://github.com/DiamondLightSource/mx-bluesky/issues/1021

src/mx_bluesky/hyperion/external_interaction/callbacks/snapshot_callback.py
```diff
@@ -39,69 +58,185 @@ class BeamDrawingCallback(PlanReactiveCallback):
...
+            )
+            data["oav-snapshot-last_saved_path"] = output_snapshot_path
         return doc
 
+    def _image_plane_offset_mm(
+        self,
+        snapshot_info: _SnapshotInfo,
+        current_sample_pos_mm: tuple[float, float, float],
+    ) -> tuple[float, float]:
+        assert snapshot_info.omega in (0, -90), "Unexpected base snapshot angle"
```
DominicOram, line 196: Should: Why does this matter? It should be generic for any omega?

tests/system_tests/hyperion/external_interaction/test_load_centre_collect_full_plan.py
```diff
@@ -656,15 +674,190 @@ def test_execute_load_centre_collect_rotation_snapshots(
...
+                "tests/test_data/test_images/thau_1_91_expected_0.png",
+                "tests/test_data/test_images/thau_1_91_expected_270.png",
+                "tests/test_data/test_images/thau_1_91_expected_270.png",
+            ],
+            [
+                fetch_datacollection_attribute(rotation_dc_ids[i], col)
+                for col in ["xtalSnapshotFullPath1", "xtalSnapshotFullPath2"]
+                for i in (
+                    0,
+                    1,
```
DominicOram, line 858: Nit: Remove the trailing `,` and it will format more nicely

tests/test_data/test_daq_configuration/display.configuration
```diff
@@ -0,0 +1,42 @@
+zoomLevel = 1.0
```
DominicOram, line 1: Should: Why not re-use https://github.com/DiamondLightSource/mx-bluesky/blob/main/tests/test_data/test_display.configuration?

tests/test_data/test_daq_configuration/OAVCentring_hyperion.json
```diff
@@ -0,0 +1,23 @@
+{
```
DominicOram, line 1: Should: Why not re-use https://github.com/DiamondLightSource/mx-bluesky/blob/main/tests/test_data/test_OAVCentring.json?

tests/unit_tests/common/external_interaction/callbacks/common/test_snapshot_callback.py
```diff
@@ -39,38 +73,191 @@ def simplified_experiment_plan(
...
+            "display_config": "tests/test_data/test_daq_configuration/display.configuration",
+        }
+
+    @pytest.fixture()
+    def snapshot_oav_with_1x_zoom(self, oav_with_snapshots):
+        set_mock_value(oav_with_snapshots.zoom_controller.level, "1.0x")
+        return oav_with_snapshots
+
+    @pytest.mark.parametrize("chis", [[0], [0, 30]])
+    def test_snapshot_callback_generate_snapshot_from_gridscan(
```
DominicOram, line 182: Should: I feel like having one big test here means that we're potentially missing a lot of coverage. For example, it would be good to see a range of tests where given different initial `x, y, z, omega` and different final `x, y, z, omega`, `draw_crosshair` is called with the expected values

DominicOram (changes requested):
Thanks, some comments in code

src/mx_bluesky/hyperion/experiment_plans/oav_snapshot_plan.py
```diff
@@ -62,12 +65,25 @@ def _setup_oav(
     )
 
 
+def _generate_oav_snapshots(composite: OavSnapshotComposite, params: WithSnapshot):
+    """Generate rotation snapshots from previously captured grid snapshots"""
+    for omega in 0, 270:
+        snapshot_path = str(
+            params.snapshot_directory / (_snapshot_filename(omega) + ".png")
+        )
```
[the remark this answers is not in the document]
rtuck99: We have to emit multiple docs here, as the docs have to be forwarded to the ispyb callback. I don't feel comfortable reusing the same document as the documents contain sequence numbers generated by the run engine.

We can opt not to populate the file name here and allow the callback to compute the filename though.

tests/unit_tests/common/external_interaction/callbacks/common/test_snapshot_callback.py
rtuck99: I have added some more tests, which did find an issue with the z-axis conversion which wasn't discovered in the test due to the z-axis offset not being big enough to discriminate in the images.

src/mx_bluesky/hyperion/experiment_plans/oav_snapshot_plan.py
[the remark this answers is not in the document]
DominicOram: We could produce one doc here and get the snapshot callback to emit twice but no worries, lets keep it as emitting two here. I think constructing the filename in the callback would be better, I don't like that we have to assume what omega values were used in grid detect here as if we want to change it it's not clear that we will need to change this code too

DominicOram (changes requested):
Great, thanks couple more comments but mostly there.

tests/test_data/test_daq_configuration/display.configuration
rtuck99: See my other comment

https://github.com/DiamondLightSource/mx-bluesky/pull/950/files#r2063768324

DominicOram: The values for the other files are arbitrary though so we could just change those so that it works for this?

src/mx_bluesky/hyperion/external_interaction/callbacks/snapshot_callback.py
```diff
@@ -10,17 +15,35 @@
 from mx_bluesky.common.external_interaction.callbacks.common.plan_reactive_callback import (
     PlanReactiveCallback,
 )
-from mx_bluesky.common.parameters.constants import DocDescriptorNames
+from mx_bluesky.common.parameters.components import WithSnapshot
+from mx_bluesky.common.parameters.constants import DocDescriptorNames, PlanNameConstants
 from mx_bluesky.common.utils.log import ISPYB_ZOCALO_CALLBACK_LOGGER as CALLBACK_LOGGER
 
+COMPRESSION_LEVEL = 6
```
DominicOram, line 22: Should: Can we have a comment about why 6 was chosen?

DominicOram (approved):
Great, thanks. Just the one comment, feel free to merge after addressing

Issue #358, opened 2023-06-30
Have a baton between Hyperion and GDA


DominicOram:
We should have an EPICS PV for the baton and who is controlling the beamline. Hyperion/GDA should set this appropriately

DominicOram:
Pull this into a REST service instead?

graeme-winter:
For the record, I have always thought a PV based solution would be a good fit - not worked through the flow chart, but this also means we can (at least in theory) also have a direct EPICS GUI taking the baton to stop either GDA or Hyperion from interacting

DominicOram:
I think all that should be able to be handled under the "Usual GDA handling" box, where we fallback to the existing code

DominicOram:
Comment from discussion with scientistst is "Generally fine but should also check `CURRENT_USER` during Hyperion operation"

DominicOram:
Hyperion side in https://github.com/DiamondLightSource/mx-bluesky/pull/956

Need to do:
- [ ] GDA side
- [ ] Actually integrate the logic, needs https://jira.diamond.ac.uk/browse/I03-894 at least

PR #956, opened 2025-03-26
Add baton handling between Hyperion and GDA

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/358

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/1144

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


tests/unit_tests/hyperion/test_baton_handler.py
```diff
@@ -0,0 +1,140 @@
...
+@pytest.fixture()
+async def baton() -> Baton:
+    async with init_devices(mock=True):
+        baton = Baton("")
+    set_mock_value(baton.requested_user, HYPERION_USER)
+    return baton
+
+
+def test_loop_forever_if_not_hyperion_requested():
+    pass
```
rtuck99, line 23: !!!

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+        requested_user = yield from bps.rd(baton.requested_user)
+
+
+def move_to_default_state():
+    # To be filled in in https://github.com/DiamondLightSource/mx-bluesky/issues/396
+    yield from bps.null()
+
+
+def baton_handler(baton: Baton, composite: LoadCentreCollectComposite):
```
rtuck99, line 57: Can this be named something different? `baton_handler` makes it sound like an event handler, but it's more of a wrapper around a critical section.

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -0,0 +1,71 @@
...
+    yield from wait_for_hyperion_requested(baton)
+    yield from bps.abs_set(baton.current_user, HYPERION_USER)
+
+    def default_state_then_collect():
+        yield from move_to_default_state()
+        yield from main_hyperion_loop(baton, composite)
+
+    def release_baton():
+        yield from bps.abs_set(baton.requested_user, NO_USER)
```
rtuck99, line 66: If GDA requested the baton, its requested baton value will be overwritten

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -0,0 +1,71 @@
...
+
+def move_to_default_state():
+    # To be filled in in https://github.com/DiamondLightSource/mx-bluesky/issues/396
+    yield from bps.null()
+
+
+def baton_handler(baton: Baton, composite: LoadCentreCollectComposite):
+    yield from wait_for_hyperion_requested(baton)
+    yield from bps.abs_set(baton.current_user, HYPERION_USER)
+
```
rtuck99, line 60: Technically there is a potential race condition here whereby we ought to check that the requested user is still hyperion after we set the current user, in case another process changed the requested user between the two calls. However I think that this is probably fine since in reality the odds of this happening are vanishingly small especially as requesting the baton is essentially a product of human intervention.

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -0,0 +1,71 @@
...
+NO_USER = "None"
+
+
+def wait_for_hyperion_requested(baton: Baton):
+    SLEEP_PER_CHECK = 0.1
+    while True:
+        requested_user = yield from bps.rd(baton.requested_user)
+        if requested_user == HYPERION_USER:
+            break
+        yield from bps.sleep(SLEEP_PER_CHECK)
```
rtuck99, line 25: The fact that this isn't the only place where we call sleep in a loop waiting for a value makes me think maybe this is something that either bluesky should support natively, or maybe we are doing something wrong and we should be doing it at the device level?

DominicOram: Yep, this is a long standing discussion with `bluesky` https://github.com/bluesky/bluesky/issues/1618. I think it adds weight to us needing that but is outside the scope of this issue.

src/mx_bluesky/hyperion/baton_handler.py
DominicOram: I guess the outcome would be that hyperion moves to the default state then the main loop does nothing and the `finally` releases the baton. This seems fine and, as you said, probably not going to ever really happen in prod so I'll leave it. Good spot though, thanks.

jacob720 (approved):
Looks good to me!

olliesilvester:
This looks fine as-is, but I'd suggest a slightly different way to watch the `requested_user` signal

- Do a `bps.monitor` on `requested_user`, and subscribe to a new callback which raises `RequestedUserInterruption` if the value is not `hyperion` while the hyperion loop is still running
- Run the main hyperion loop in a contingency wrapper with an except plan `handle_requested_user_interruption`. Not sure if that function should just log some warnings or actually stop Hyperion early

As mentioned though, probably not too important since it's only an issue with manual intervention

DominicOram:
What's the advantage of monitoring it? We don't actually care if it changes during a collection. We care at the end of the collection, where we will pass control back if we need to.

olliesilvester:
It would just mean we can react a bit faster if it does change, but given
> We don't actually care if it changes during a collection

then yeah I guess it doesn't really give us anything except a slightly more detailed log message

PR #957, opened 2025-03-27
Plan for changing aithre goniometer omega velocity

### LINKED ISSUE #930: Convert omega velocity control to bluesky
The GUI has two buttons for controlling the speed of the goniometer; "slow turn" and "fast turn". These are linked to a lambda which does a caput to change the value:

        self.ui.buttonSlowOmegaTurn.clicked.connect(lambda: ca.caput(pv.omega_velo, 15))
        self.ui.buttonFastOmegaTurn.clicked.connect(lambda: ca.caput(pv.omega_velo, 40))

These should be converted to bluesky

JamesDoingStuff:
Fixes #930

1. Confirm new test passes
2. Optionally, run plan on goniometer in laser lab

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
+++ b/src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
@@ -1,0 +1,10 @@
+import bluesky.plan_stubs as bps
+from bluesky.utils import MsgGenerator
+from dodal.devices.aithre_lasershaping.goniometer import Goniometer
+
+
+def change_goniometer_turn_speed(
+    goniometer: Goniometer, velocity: float
+) -> MsgGenerator:
+    """Set the velocity of the goniometer"""
+    yield from bps.mv(goniometer.omega.velocity, velocity)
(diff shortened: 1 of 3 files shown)


src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -0,0 +1,10 @@
+import bluesky.plan_stubs as bps
+from bluesky.utils import MsgGenerator
+from dodal.devices.aithre_lasershaping.goniometer import Goniometer
+
+
+def change_goniometer_turn_speed(
+    goniometer: Goniometer, velocity: float
```
DominicOram, line 7: Should: Once we get to exposing this in `BlueAPI` we can use `inject` to fill in the devices, e.g. see https://github.com/DiamondLightSource/mx-bluesky/blob/main/src/mx_bluesky/beamlines/i04/thawing_plan.py. This will mean that a caller needn't specify the goniometer. The inject takes the name of the device (which you have set in `aithre.py` so in case `goniometer`

DominicOram (changes requested):
Great, thanks. One comment in the code. 

It may be that for small plans like this I think we can probably expose `mv` via `BlueAPI` and so the client can call `run_plan("mv", "goniometer.omega.velocity", 100)` instead of `run_plan("change_goniometer_turn_speed", 100)` and we don't need so much code on this side. However, now that I've written that out I'm not sure blueAPI understands nested things like that, we would need to check. I think this is fine as is though, just a thing to try out once BlueAPI is deployed

DominicOram (approved):
Great, thank you!

PR #961, opened 2025-03-31
i23 Laser: Plan to go to maximum gonio position

### LINKED ISSUE #932: convery gonio goTopm3600 to bluesky
On the GUI, there is a button labelled "+/-3600". Once clicked, it runs https://github.com/DiamondLightSource/aithre/blob/e6bcfa3915a6df59bf609681c6a3258c3a4a07c1/bin/guiv4_prod.py#L454 which checks the gonio omega value and moves to 3600 or -3600, whichever is further away. This is used as a prolonged rotation to cut samples in a lathe fashion as the gonio doesn't like going further than this (no idea why).

This should be converted to a bluesky plan, but also it would be good to have the option to "bounce" between the two values so it continuously rotates without needing to press the button when it reaches one of the extremes. There would need to be a "stop rotation" button added to stop the motion when needed.

JamesDoingStuff:
Fixes first half of #932 

1. Check new test is passing
2. Ideally, test on laser shaping setup in lab 18

### Checks for reviewer


src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -16,3 +16,12 @@ def change_goniometer_turn_speed(
...
     yield from bps.mv(goniometer.omega.velocity, velocity)
+
+
+def go_to_furthest_maximum(
+    goniometer: Goniometer = inject("goniometer"),
+) -> MsgGenerator:
+    """Go to +/-3600, whichever is further away"""
+    current_value: float = yield from bps.rd(goniometer.omega.user_readback)
+
+    yield from bps.mv(goniometer.omega, -3600 if current_value > 0 else 3600)
```
DominicOram, line 27: Could: Would be good to pull the 3600 out into a constant called like `LIMIT_OF_TRAVEL`.

DominicOram: Or, @co2e14, is it possible to get this set as the `LLM`/`HLM` of the motor? Then we can just read from that?

DominicOram (approved):
Great, thank you! One optional comment

DominicOram (approved):
Great, thank you!

PR #969, opened 2025-04-02
(#931) Plans to move Aithre goniometer sample stage

### LINKED ISSUE #931: convert jog buttons to bluesky
Buttons on the gui are used to jog the goniometer stages up/down/left/right/forward/backward. These are found at: https://github.com/DiamondLightSource/aithre/blob/e6bcfa3915a6df59bf609681c6a3258c3a4a07c1/bin/guiv4_prod.py#L311

The lambda function feeds to jogSample: https://github.com/DiamondLightSource/aithre/blob/e6bcfa3915a6df59bf609681c6a3258c3a4a07c1/bin/guiv4_prod.py#L420 which is used to move the gonio stages. For movement up and down, the current omega value must be taken into account to determine how much to move y and z. 

This should be converted to a bluesky plan with blueapi support.

### LINKED ISSUE #979: Device logic for vertical sample movement on Aithre
Logic for moving the sample stage up and down - see #931 - should exist in the device rather than the plan.  Depends on https://github.com/bluesky/ophyd-async/pull/852. WIP code can be found on https://github.com/DiamondLightSource/dodal/pull/1164

## Acceptance Criteria
- Logic for converting vertical movement to y and z components is at the device level
- Vertical position can be set using the same jog plan that left-right movement uses

JamesDoingStuff:
Fixes #931
Fixes #979 

Requires Dodal PR  [#1164](https://github.com/DiamondLightSource/dodal/pull/1164)

1. Check new tests are passing
2. Test on Aithre setup

### Checks for reviewer


src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -16,3 +18,32 @@ def change_goniometer_turn_speed(
...
+
+def jog_sample(
+    direction: str, increment_size: float, goniometer: Goniometer = inject("goniometer")
+) -> MsgGenerator:
+    """Adjust the goniometer stage positions vertically"""
+    direction_map = {
+        "right": (goniometer.x, 1),
+        "left": (goniometer.x, -1),
+        "z_plus": (goniometer.z, 1),
+        "z_minus": (goniometer.z, -1),
```
DominicOram, line 31: Should: The jog that Chris has written only moves 0.005. Can we pull the amount moved into a constant and put it at 0.005?

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -16,3 +18,32 @@ def change_goniometer_turn_speed(
 ) -> MsgGenerator:
     """Set the velocity of the goniometer"""
     yield from bps.mv(goniometer.omega.velocity, velocity)
+
+
+def jog_sample(
+    direction: str, increment_size: float, goniometer: Goniometer = inject("goniometer")
+) -> MsgGenerator:
+    """Adjust the goniometer stage positions vertically"""
+    direction_map = {
```
DominicOram, line 27: Should: Can you make the options an enum rather than a string? Will lead to less possibility of accidents where people mistype the direction

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -16,3 +18,32 @@ def change_goniometer_turn_speed(
...
+    }
+
+    if direction in direction_map:
+        axis, sign = direction_map[direction]
+        yield from bps.mvr(axis, sign * increment_size)
+    elif direction in {"up", "down"}:
+        omega: float = yield from bps.rd(goniometer.omega)
+        z_component = (math.cos(math.radians(omega))) * increment_size
+        y_component = (math.sin(math.radians(omega))) * increment_size
+        sign = 1 if direction == "up" else -1
```
DominicOram, line 41: Could: You could argue this logic should live in the `ophyd_async` device, maybe as a derived signal (https://blueskyproject.io/ophyd-async/main/how-to/derive-one-signal-from-others.html). Basically, I think if this move in `up` and `down` is one that @co2e14 wants to do in lots of places (not just jog) it should go in the device

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
```diff
@@ -16,3 +18,32 @@ def change_goniometer_turn_speed(
...
+
+        yield from bps.rel_set(
+            goniometer.sampz, sign * z_component, group="gonio_stage"
+        )
+        yield from bps.rel_set(
+            goniometer.sampy, sign * y_component, group="gonio_stage"
+        )
+        yield from bps.wait("gonio_stage")
```
DominicOram, line 49: Should: I was originally worried that we're waiting for the `up`/`down` move here but not for the other moves. But we are because `mvr` waits but `rel_set` doesn't. To avoid this confusion, I think we should try and limit ourselves to either `mv` or `set` in this plan. You can do the `mv` for the `up`/`down` by doing `yield from bps.mvr(goniometer.sampz, sign * z_component, goniometer.sampy, sign * y_component)` or you can do `rel_set` in the other moves, putting it in the `gonio_stage` group and wait on that all at the end. Up to you

tests/unit_tests/beamlines/aithre_lasershaping/test_goniometer_controls.py
```diff
@@ -37,3 +52,34 @@ def test_change_goniometer_turn_speed(
         and msg.obj.name == "goniometer-omega-velocity"
         and msg.args[0] == 40,
     )
+
+
+@pytest.mark.parametrize(
```
DominicOram, line 57: Should: Good set of tests in general, thanks! But would be good to test the case where a motor is already at a position and you jog. Basically, I think the current tests will work if you replace everything in the plan with `abs_set` rather than `rel_set`, so we should make sure tests would catch that

DominicOram (changes requested):
Thanks, a few comments in the code

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
JamesDoingStuff: I did initially have it as a constant, but the existing Z+ and Z- buttons actually use 0.05 instead, so I wondered if it'd just be simpler to have a variable increment size. I'd be happy to implement two constants if you think that'd be better though

DominicOram: @co2e14 why are these different? Naively, the units look the same:
```
$ caget LA18L-MO-LSR-01:SAMPZ.EGU
LA18L-MO-LSR-01:SAMPZ.EGU      mm
$ caget LA18L-MO-LSR-01:Z.EGU
LA18L-MO-LSR-01:Z.EGU          mm
```
It's just that you want Z+ to be a bigger move because it tends to be more out?

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
JamesDoingStuff: I've had a look at this and it looks like it'd be a good idea, so I had a go at creating a derived signal for vertical position, but I _think_ I'm running into the same issue that you're discussing [here](https://github.com/DiamondLightSource/dodal/pull/1148/files#r2020872446). I can look at using the multi derived signal method, but that looks a little more involved, so I wonder if it might be better to leave this as it is, just for now?

DominicOram: [This](https://github.com/DiamondLightSource/dodal/pull/1148#discussion_r2019028282) issue? Or [this](https://github.com/DiamondLightSource/dodal/pull/1148#discussion_r2019051284)?

JamesDoingStuff: Oops sorry, I think the link broke. The former, where it requires the raw signals to all be rw

DominicOram: As discussed, please make a new issue that is converting this into a device, linking whatever you have so far.

JamesDoingStuff:
Changes since last review:

- Map of jog directions converted to enum
- All `rel_set` converted to `mvr`
- Tests now move the device twice in the same direction, thus checking relative motion behaves correctly
- The logic for converting the vertical motion values into `y` and `z` values has been moved to the device ([dodal#1164](https://github.com/DiamondLightSource/dodal/pull/1164))

src/mx_bluesky/beamlines/aithre_lasershaping/goniometer_controls.py
JamesDoingStuff: Had a chat with @co2e14 about this a while ago. Essentially it's that movement in Z equates to moving in and out of focus on the video feed, which is harder to judge than the other directions and therefore doesn't need as much precision.

DominicOram (approved):
Great, thank you!

PR #970, opened 2025-04-03
968 code map and tidy for grid detect xrc

### LINKED ISSUE #968: Reinstate the code map and tidy for grid detect and xrc
In artemis we used to have a [code map](https://github.com/DiamondLightSource/hyperion/wiki/Code-Map). Which we lost when moving everything into `mx-bluesky`. For doing https://github.com/DiamondLightSource/mx-bluesky/issues/918 it would be good to redo this and flesh it out again. Additionally, in writing this documentation we should think about if there are smaller functions that could be tidied to make them more generic/readable e.g. https://github.com/DiamondLightSource/mx-bluesky/blob/c960431be7a9f23bec637d872ba61b05142c5530/src/mx_bluesky/hyperion/experiment_plans/change_aperture_then_move_plan.py#L13C5-L13C38 should take whether or not to do the stub offsets, not a full Parameter object.

## Acceptance Criteria
- The code map is copied back into `mx-bluesky`
- The parts for grid detect and xrc are updated if they are out of date and made detailed enough so that a new-ish starter can understand them
- In doing this, if there are places where code can be made more generic/cleaner we should do so

jacob720:
Fixes #968 

Adds a code map into the docs for `grid_detect_then_xray_centre_plan`. 
Also minor changes to simplify params passed into plans.

1. Build the docs
2. Confirm they make sense

### Checks for reviewer


DominicOram (approved):
This is great, thank you. I think my only comment is that in your diagram you have order of operations running left to right but to fit things in you've spread them out so `setup_general_oav` is further left than `set_detector_z` but actually called after. It might not matter, I think it is still pretty readable, but maybe we can think of a cleaner way of showing it? If not, don't worry

PR #980, opened 2025-04-07
847 rename multi_rotation_plan to rotation_plan

### LINKED ISSUE #847: Hyperion: Remove single rotation scan plan

jacob720:
Fixes #847, following along from #935.

`multi_rotation_scan` is renamed to `rotation_scan`, and the tests that were found in `test_rotation_scan_plan.py` and `test_multi_rotation_scan_plan.py` are combined into one file. 

1. Run tests
2. Confirm they all pass
3. Check naming decisions make sense

### Checks for reviewer


rtuck99 (commented):
Technically looks fine. 

Not sure whether it's a good thing or not as it stands - this now means that `rotation_scan` takes a `MultiRotationScan` and not a `RotationScan`, I think I preferred the previous naming consistency.

For better or worse, we already have a `RotationScan` and it is still used internally. Probably we can clean it up a bit now that it's not exposed in the public API.

src/mx_bluesky/hyperion/experiment_plans/experiment_registry.py
```diff
@@ -47,7 +47,7 @@ class ExperimentRegistryEntry(TypedDict):
         "setup": pin_centre_then_xray_centre_plan.create_devices,
         "param_type": PinTipCentreThenXrayCentre,
     },
-    "multi_rotation_scan": {
```
rtuck99, line 50: I think we can now remove RotationScan from param_type?

DominicOram:
>Not sure whether it's a good thing or not as it stands - this now means that rotation_scan takes a MultiRotationScan and not a RotationScan, I think I preferred the previous naming consistency.

My concern is that the external API isn't that clear. To a scientist wanting to run a rotation scan if we only have `multi_rotation_scan` they may disregard it as not what they want whereas having a `rotation_scan` that allows you to specify both types seems better. How about renaming the `MultRotationScan` too then?

jacob720:
I've renamed `MultiRotationScan` to `RotationScan`, and what was previously `RotationScan` is now `SingleRotationScan`. in `param_types`, there's now only `RotationScan` (before there was both `RotationScan` and `MultiRotationScan`.

rtuck99 (approved):
Approved

PR #985, opened 2025-04-10
Robin: fixes 10Apr

robinowen-i24:
Fixes #ISSUE

edm can't handle long file paths. Limit is ~130chars so images weren't being displaced properly in the pump probe edm. Added two new png files (s1l.png and s2l.png) with shorter filenames which circumvents this for now. edm updated to reflect this.
Also fixed a last hangup resulting from replacement of BL24I-MO-DCM-01:LAMBDA with BL24I-MO-DCM-01:WAVELENGTH
There was a remaining instance of lambda in setup eiger. Fixed this/replaced with wavelength and added wavelength PV to py.py


src/mx_bluesky/beamlines/i24/serial/setup_beamline/pv.py
```diff
@@ -154,7 +154,7 @@ def __which__():
 dcm_roll1 = "BL24I-MO-DCM-01:XTAL1:ROLL"
 dcm_roll2 = "BL24I-MO-DCM-01:XTAL2:ROLL"
 dcm_pitch2 = "BL24I-MO-DCM-01:XTAL2:PITCH"
-dcm_lambda = "BL24I-MO-DCM-01:LAMBDA"
```
DominicOram, line 157: Should: This PV isn't used anywhere, in fact `pv.dcm_lambda` is used in a comment but no where else. Can we not remove it and remove the use of `pv.dcm_lambda` in the comment?

DominicOram (changes requested):
Thanks, one comment

src/mx_bluesky/beamlines/i24/serial/setup_beamline/pv.py
robinowen-i24: Yes, that should be fine/sounds like the right thing to do - BL24I-MO-DCM-01:LAMBDA now retired.

DominicOram: Ok, given Noemi is now away I will just do this to get it closed off.

DominicOram (approved):

PR #989, opened 2025-04-14
Update i24 roadmap docs

noemifrisina:
How to test:
- Generate docs with `tox -e docs`
- Check page is working


docs/developer/serial-crystallography-on-i24/how-to/pmac-docs.rst
```diff
@@ -1,3 +1,16 @@
+=================================
+PMAC-run operations on I24 serial
+=================================
+
+Several operations in a fixed target (FT) collection are run vie the PMAC, by sending a command to the ``PMAC_STRING`` PV
```
rtuck99, line 5: Perhaps this should reference the PMAC dodal device and the `pmac_string` signal since that's presumably what users will be interacting with.

docs/developer/serial-crystallography-on-i24/how-to/pmac-docs.rst
```diff
@@ -18,6 +31,8 @@ to motors. The axes allocations defined for the chip stages set up are:
 
 When an X-axis move is executed, the #1 motor will make the move.
 
+It should be noted that the PMAC coordinate system used by I24 serial is number 2, which can be specified in a PMAC string by starting the command with ``&2``.
+
 When running chip collections, the stage motors are moved via the `PMAC
 device <https://github.com/DiamondLightSource/dodal/blob/main/src/dodal/devices/i24/pmac.py>`__
 in a couple of different ways.
```
rtuck99, line 38: Is there an authoritative reference for the PMAC instruction set/language?
It would be good to reference it here

rtuck99 (approved):
Approved, 

However I think maybe it would be worth considering whether some of this information should be moved to the `dodal` documentation - it's a bit unclear exactly who the audience is for this documentation, there are a number of different potential readers:

* Users of the mx-bluesky experiment plans
* Implementers of the mx-bluesky experiment plans / people wanting to use the PMAC directly
* Dodal developers making changes to the `PMAC` device

As such I think it may be worth separating the concerns a bit, and possibly moving some of the documentation to the `dodal` project to document the PMAC device more fully.

PR #1005, opened 2025-04-17
Create an mx-bluesky-blueapi image and add to CI

### LINKED ISSUE #922: Run `BlueAPI` from our own container
Currently we use the `BlueAPI` container to run plans on i04/vmxm.

This works for simple plans but doesn't when we have dependencies outside of `BlueAPI` as it does not install our dependencies.

For some of our dependencies we just need to do a `pip install` (e.g. for ispyb) for others there are system dependencies e.g. `cv2` in https://github.com/DiamondLightSource/mx-bluesky/issues/920 needs additional apt-gets (https://stackoverflow.com/questions/55313610/importerror-libgl-so-1-cannot-open-shared-object-file-no-such-file-or-directo).

As discussed with @callumforrester and @DiamondJoseph the proposed solution to this is that instead we deploy a specific `mx-bluesky` image that uses `blueAPI` as a base image.

## Acceptance Criteria
- I can deploy an `mx-bluesky` image that works as the `blueAPI` image does now except has our external dependencies in it
- I can do this using ArgoCD
- There is documentation on the `mx-bluesky` side on how to do this deployment
- There is documentation on the `BlueAPI` side on how to set this up in a generic way
- Building this container is done as part of our CI

olliesilvester:
Fixes #922

### Checks for reviewer
1. Check new dockerfile looks sensible
2. Check that CI looks like it should do a build and push upon a release
3. Check that additions to docs makes sense


docs/developer/hyperion/images/code-map.png
rtuck99, line 1: does this get referred to anywhere?

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
rtuck99, line 1: I'm in two minds about this. On the one hand, if mx-bluesky is mostly a vanilla blueapi application, then I think it would be better to not have our own deployment.yaml and instead reference their blueapi chart with our own Values.yaml e.g. as per https://gitlab.diamond.ac.uk/controls/containers/beamline/i02-1-services

That way configuration of the base blueapi wouldn't need to be duplicated here, apart from whatever we would need to supply in `Values.yaml`

However I think in order for us to get all the various folders mounted we would need `blueapi` to insert additional customisation points in their Helm chart templates.

On the other hand, at the moment I think we aren't using most of the features in the blueapi helm charts - service account creation, stateful sets, config maps, initcontainers, so perhaps there is little benefit.

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+          hostPath:
+            type: Directory
+            path: "{{ .Values.application.dataDir }}"
+      containers:
+      - name: hyperion
+        image: {{ .Values.application.imageRepository}}/mx-bluesky:{{ .Values.application.appVersion }}
+        securityContext:
+          capabilities:
+            # Required for hostNetwork 
+            drop:
```
rtuck99, line 63: I think we can just drop `- ALL` - I think there is another low-priority issue for doing this for hyperion

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+    matchLabels:
+      app: mx-bluesky-blueapi
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        app: mx-bluesky-blueapi
+    spec:
+      securityContext:
+        runAsUser: {{  .Values.securityContext.runAsUser }}
```
rtuck99, line 16: is there going to be a Values.yaml somewhere for this?

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+              - SYS_TIME
+              - SYS_TTY_CONFIG
+              - SYSLOG
+              - WAKE_ALARM            
+        resources:
+          limits:
+            cpu: "1"
+            memory: "1Gi"
+        ports:
+          - name: hyperion-api
```
rtuck99, line 104: I think we want to remove references to `hyperion-` in order to avoid confusion, probably best to rename to `mx-bluesky-`

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+          periodSeconds: 5
+        volumeMounts:
+          - mountPath: "/dls_sw/{{ .Values.application.beamline }}"
+            name: dls-sw-bl
+            readOnly: true
+            mountPropagation: HostToContainer
+          - mountPath: "/dls_sw/apps"
+            name: dls-sw-apps
+            readOnly: true
+            mountPropagation: HostToContainer
```
rtuck99, line 134: I have raised DiamondLightSource/mx-bluesky#1057 which I think we should make a priority as it was OK for a one off but now we have two containers both doing this and I don't want to make it a pattern.

helmcharts/mx-bluesky-blueapi/templates/deployment.yaml
```diff
@@ -0,0 +1,151 @@
...
+          - mountPath: "/dls_sw/dasc"
+            name: dls-sw-dasc
+            readOnly: true
+            mountPropagation: HostToContainer
+          - mountPath: "/app/hyperion/src"
```
rtuck99, line 139: These folders won't exist in your container image as they aren't created in the Dockerfile. If I understand how the blueapi image is meant to be used, it will have created its virtual environment in /venv and our Dockerfile is trying to install mx-bluesky into it.

rtuck99 (changes requested):
I think need to talk to core team about how the scratch folder blueapi deployment works.

docs/developer/hyperion/images/code-map.png
olliesilvester: Nope, this was from the PR being out of date

helmcharts/mx-bluesky-blueapi/Chart.yaml
callumforrester, line 1: FYI You should still be able to use the vanilla bluesky helm chart and override the container image configuration to use your own image.

```yaml
image:
  repository: ghcr.io/diamondlightsource/mx-bluesky-blueapi
```

PR #1020, opened 2025-04-25
I24 SSX: Make detector readback human readable

### LINKED ISSUE #905: I24 serial: make detector choice human readable
The edm dropdown for the detector choice will set the value of the general purpose PV associated with the detector name to either 0 or 1 (because of how dropdowns in edms work). This makes it difficult for the user to be sure which detector is currently in use by looking at the monitor.

Easiest solution - although quite horrible - for this would be to have the `setup_detector_stage` run by the `Move stage` button on the Detector choice edm to a caput on the PV after `_move_detector_stage`

```python

def setup_detector_stage(...) -> MsgGenerator:
    # Whatever the code does here
    # ...
    yield from _move_detector_stage(...)
    caput(det_type_pv, det_name)
```

Alternatively we could set one of the oother general PVs and show that value in the monitor.

Eventually this will be solved by moving to the web gui.

adaudon:
Fixes #905 

1. Confirm tests still pass

### Checks for reviewer


tests/unit_tests/beamlines/i24/serial/setup_beamline/test_setup_detector.py
```diff
@@ -44,13 +44,23 @@ def test_get_requested_detector_raises_error_for_invalid_value(fake_caget):
...
+        (DetRequest.pilatus.value, SSXType.EXTRUDER, Pilatus.det_y_target),
+    ],
+)
 async def test_setup_detector_stage(
-    fake_caget, detector_stage: DetectorMotion, RE: RunEngine
+    fake_caput,
+    fake_caget,
+    RequestedDetectorValue,
+    SerialType,
+    DetectorTarget,
```
DominicOram, line 60: Should: Parameter/variable names should be `snake_case` in python

tests/unit_tests/beamlines/i24/serial/setup_beamline/test_setup_detector.py
```diff
@@ -44,13 +44,23 @@ def test_get_requested_detector_raises_error_for_invalid_value(fake_caget):
 
 
 @patch("mx_bluesky.beamlines.i24.serial.setup_beamline.setup_detector.caget")
+@patch("mx_bluesky.beamlines.i24.serial.setup_beamline.setup_detector.caput")
+@pytest.mark.parametrize(
```
DominicOram, line 48: Nice use of parametrize

tests/unit_tests/beamlines/i24/serial/setup_beamline/test_setup_detector.py
```diff
@@ -44,13 +44,23 @@ def test_get_requested_detector_raises_error_for_invalid_value(fake_caget):
...
+    "RequestedDetectorValue, SerialType, DetectorTarget",
+    [
+        (DetRequest.eiger.value, SSXType.FIXED, Eiger.det_y_target),
+        (DetRequest.pilatus.value, SSXType.EXTRUDER, Pilatus.det_y_target),
+    ],
+)
 async def test_setup_detector_stage(
```
DominicOram, line 55: Should: As this test stands it will pass on `main`, which means it's not testing the new logic you added in this PR. You should assert that `det_type_pv` gets set to the correct string

DominicOram (changes requested):
Thank you, a few comments in code

tests/unit_tests/beamlines/i24/serial/setup_beamline/test_setup_detector.py
```diff
@@ -46,21 +46,23 @@ def test_get_requested_detector_raises_error_for_invalid_value(fake_caget):
...
+    expected_detector,
     detector_stage: DetectorMotion,
     RE: RunEngine,
 ):
-    fake_caget.return_value = RequestedDetectorValue
-    RE(setup_detector_stage(SerialType, detector_stage))
-    assert await detector_stage.y.user_setpoint.get_value() == DetectorTarget
+    fake_caget.return_value = requested_detector_value
+    assert _get_requested_detector(str(requested_detector_value)) == expected_detector
+    RE(setup_detector_stage(serial_type, detector_stage))
```
DominicOram, line 67: Should: This is checking that `requested_detector_value` corresponds to the detector name, it's not checking that the name is getting written to the expected PV

DominicOram (changes requested):
I still think this test would pass on `main`, in general you want to aim to write tests that fail on `main` and pass on the feature branch, this proves that you have fixed the change

DominicOram (approved):
Great, thank you!

PR #1026, opened 2025-04-28
Fix unit tests broken by DiamondLightSource/dodal#1193

rtuck99:
Fixes unit tests broken by
* DiamondLightSource/dodal#1193

by patching the i03 default beamline_parameters file


olliesilvester (approved):
Thanks!

PR #1027, opened 2025-04-28
Add better logging on i24 SSX failure

### LINKED ISSUE #1000: I24 Serial: Log exception traceback
On error, in `run_fixed_target_plan` the contingency wrapper runs `run_aborted_plan` and does not re-raise. This means that the exception that caused the error does not get reliably logged. We should also log the error as part of the wrapper.

## Acceptance Criteria
- Errors are logged when they occur in `run_fixed_target_plan`

DominicOram:
Fixes #1000 

1. Confirm modified test passes

### Checks for reviewer


adaudon (approved):
Looks good and tests pass

PR #1035, opened 2025-04-30
Unit tests: Ensure that run engine event loop always terminates

rtuck99:
Fixes an issue where `RunEngine` fixtures were creating event loops that were never cleaned up.

This led to 
```
  File "/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/socket.py", line 608, in socketpair
    a, b = _socket.socketpair(family, type, proto)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [Errno 24] Too many open files
```

being reported in unit tests where `ulimit -n` was set too low (1024)

1. Unit tests pass
### Checks for reviewer

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -377,7 +377,7 @@ def done_status():
 
 
 @pytest.fixture
-def eiger(done_status):
+def eiger(done_status, RE: RunEngine):
     eiger = i03.eiger(connect_immediately=True, mock=True)
     eiger.stage = MagicMock(return_value=done_status)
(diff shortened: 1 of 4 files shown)


DominicOram (approved):
Thanks, I can't reproduce your error locally but this improves things and doesn't slow anything down so happy to merge. I wonder if https://github.com/DiamondLightSource/dodal/issues/1204 is the better long term solution

Issue #842, opened 2025-02-20
Hyperion: Document rotation_direction stuff


DominicOram:
If I set `rotation_direction` to be `POSITIVE` I get an invalid nexus file because the omega axis is not flipped. Instead if I set `omega_flip` to True I get a correct nexus file, however I also get my `omega_start` inverted. This is confusing to a user who will likely just set `rotation_direction` and expect everything to follow through.

## Acceptance Criteria
- We discuss how these parameters should work together to avoid user error
- We implement the decision

DominicOram:
Dom to look at this, maybe I'm being an idiot

rtuck99:
Firstly a clarification regarding the purpose of `omega_flip` and `rotation_direction`:
`rotation_direction` is positive or negative to change the physical direction of rotation and is always with respect to the same `omega` axis orientation. `omega_flip` specifies the orientation of the `omega` axis in the hardware layer and therefore any internal conversions required in Hyperion from the rotations specified in the request parameters.

https://diamondlightsource.slack.com/archives/G017HTC55M1/p1740412555625209?thread_ts=1739890720.794019&cid=G017HTC55M1

> So the omega_flip parameter is something that I added during the whole investigation into omega rotation that I did a while back. Essentially omega_flip controls whether we consider the positive rotation axis to be oriented the same in "hyperion request space" as in hardware-space.
So we should never need to change this.
Note that this means that changing it inverts the motor omega coordinates for all  motor commands for any given request.
As it was originally set up, the default value of True for the beamline and with the requests as specified by GDA it should result in identical behaviour to GDA for single rotations.
As omega_flip determines the hardware +ve rotation axis orientation, any PCAP data written out to the nexus files will be interpreted with respect to this axis, therefore omega_flip  controls the axis vector written out in the nexus file.

The above semantics have been documented in various places in the `mx-bluesky` source but given the above confusion I think additional documentation wouldn't hurt.

After the above clarification, a meeting was held to discuss the implementation of reversing rotation direction on every other sweep, 

* #839 

the outcome of which was as follows:

>    * There are two approaches to rotating backwards:
>        *   Keep the vectors of rotation the same in the nexus file and count omega backwards
>        *   Swap the vector of rotation and keep omega counting forwards
>    * These are mathematically equivalent
>    * Historically third-party analysis packages assumed omega was counting forwards and so this is mildly preferred by analysis
>    * In practice in-house analysis pipelines correct for this even if counting backwards
[...]

PR #1036, opened 2025-04-30
Additional documentation on coordinate systems to document omega_flip

rtuck99:
Also carry over coordinate system documentation from old Hyperion Wiki

Fixes 
* #842 

### Checks for reviewer


docs/developer/hyperion/reference/coordinate-systems.rst
```diff
@@ -0,0 +1,76 @@
...
+------------------------
+
+.. include:: <isogrk1.txt>
+
+The diagram below shows the x,y,z directions of the smargon motor coordinate system. The smargon y, z axes change with the :math:`\omega`. x,y,z are measured in millimetres.
+
+
+* The pin runs from -ve x to +ve x
+* y is +ve towards the ceiling when :math:`\omega = 0`.
+* z is +ve towards the detector/oav and parallel to the beam when :math:`\omega = 0`.
```
DominicOram, line 49: Should: The detector/oav are at different sides of the sample. The OAV is upstream and the detector is downstream. It may be better to define this as which direction is downstream/upstream at omega 0.

DominicOram (changes requested):
Great, thanks. Just one comment

docs/developer/hyperion/reference/coordinate-systems.rst
rtuck99: I've updated the diagrams and changed the text slightly. Hopefully that should be clearer.

DominicOram (approved):
Great, thank you!

PR #1054, opened 2025-05-08
Create I04 grid detect xrc plan

### LINKED ISSUE #918: Create a common grid detect then xray centre plan
As part of the project to get mx-bluesky fast grid scans on i04, we need to create a `common_grid_detect_then_xray_centre_plan`. This should look similar to the grid detect plan that currently exists - we need to remove the reliance of the PandA, as well as parametrise what plans come next.

It might be helpful to use https://github.com/DiamondLightSource/mx-bluesky/pull/828 as a guideline here

Brief steps:
1. Move the current grid detect plan to `mx_bluesky.common.plans`. Anything that this file imports from Hyperion also needs to be moved into common - we aren't allowed to have anything from common import from Hyperion.
    - For this, the device composite should instead become a protocol of minimum devices - like `FlyscanEssentialDevices`
    - A parameter class `GridScanWithEdgeDetect` should be made into common, inheriting from `GridCommon` instead of `GridCommonWithHyperionDetectorParams` 
2. Parametrise the plan which is called after the grid detect is complete, using something similar to `BeamlineSpecificFGSFeatures` and `plan_after_getting_xrc_results`
3. Create a Hyperion entry point to this plan which should call the common version, and have unchanged functionality from what it currently does
4. Move tests appropriately so that we have a file testing the common parts, and a file testing the hyperion-specific parts

In a ticket after this, we can create the i04 entry point.

## Acceptance Criteria
- A common grid detect plan is made, and Hyperion uses this plan with unchanged functionality

jacob720:
Fixes #918, requires https://github.com/DiamondLightSource/mx-bluesky/pull/1081

Creates a common grid detect then XRC plan, and Hyperion and I04 entry points to this plan.
The `grid_detect_then_xray_centre` plan is now common and takes some additional parameters. There is a Hyperion entry point called `hyperion_grid_deetct_then_xray_centre` and an i04 entry point called `i04_grid_detect_then_xray_centre`. 

1. Check Hyperion plan runs the same as before.

### Checks for reviewer


src/mx_bluesky/common/experiment_plans/read_hardware.py
rtuck99, line 1: Should this module really be in `experiment_plans`? Feels more like an inner plan to me

src/mx_bluesky/common/experiment_plans/write_sample_status.py
rtuck99, line 1: Ditto

src/mx_bluesky/hyperion/experiment_plans/__init__.py
rtuck99, line 1: If we are renaming the external plan name here, if we aren't bothering to update GDA at this point, we might as well remove it from here, make it a private inner plan.

tests/unit_tests/hyperion/experiment_plans/conftest.py
```diff
@@ -341,3 +350,13 @@ def grid_detection_callback_with_detected_grid():
             "z_step_size_um": 0.1,
         }
         yield callback
+
```
rtuck99, line 353: This is just duplicating the one in `unit_tests/conftest.py`

tests/unit_tests/hyperion/conftest.py
```diff
@@ -74,6 +74,7 @@ def test_multi_rotation_params(tmp_path):
     )
 
 
+@pytest.fixture
 def oav_parameters_for_rotation(test_config_files) -> OAVParameters:
```
rtuck99, line 78: This duplicates the fixture in `tests/conftest.py`

src/mx_bluesky/beamlines/i04/parameters/device_composites.py
```diff
@@ -0,0 +1,45 @@
...
+from mx_bluesky.common.experiment_plans.common_flyscan_xray_centre_plan import (
+    FlyScanEssentialDevices,
+)
+
+
+@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
+class I04FlyScanXRayCentreComposite(FlyScanEssentialDevices):
+    """All devices which are directly or indirectly required by this plan"""
+
+    aperture_scatterguard: ApertureScatterguard
```
rtuck99, line 31: we don't need to override the eiger, synchrotron or the zocalo here, we can just inherit them

src/mx_bluesky/beamlines/i04/parameters/device_composites.py
```diff
@@ -0,0 +1,45 @@
...
+from dodal.devices.zebra.zebra import Zebra
+from dodal.devices.zebra.zebra_controlled_shutter import ZebraShutter
+from dodal.devices.zocalo import ZocaloResults
+
+from mx_bluesky.common.experiment_plans.common_flyscan_xray_centre_plan import (
+    FlyScanEssentialDevices,
+)
+
+
+@pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
```
rtuck99, line 27: Do we actually need this type at all???

As far as I can see, the only reason we use it is to satisfy the type constraints of `ConstructBeamlineSpecificFeatures`.

But, `GridDetectThenXRayCentreComposite` already extends `FlyScanEssentialDevices` and so still satisfies this.

`detect_grid_and_do_gridscan` already ignores any constraints `construct_beamline_specific`  might have in reality and just calls it with a `GridDetectThenXRayCentreComposite`, which is fine because in practical usage, GDTXRCC is always a superset of everything anyway.

I feel like all the type-parameter gymnastics is not really getting us anything useful and maybe we should just get rid of them.

tests/unit_tests/beamlines/i04/test_i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,418 @@
...
+    )
+    msgs = assert_message_and_return_remaining(
+        msgs,
+        predicate=lambda msg: msg.command == "set"
+        and msg.obj.name == "detector_motion-shutter"
+        and msg.args[0] == 0,
+    )
+
+
+@patch(
```
rtuck99, line 80: Rather than patch all these methods in `common_grid_detect_then_xray_centre_plan` just patch out `grid_detect_then_xray_centre` to return `iter([])`

rtuck99:
Overall plan looks ok, although would appreciate some input from @DominicOram as he knows more about what i04 actually needs to do in the plan.

A few things that can be removed/simplified - esp in the unit tests I think a lot of the patching can be cut down just by patching out the high level plans rather than all the things they call.

olliesilvester:
Thanks for reviewing @rtuck99 . I will finish this PR off at some point in the next week or so on @jacob720 's behalf.

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    An plan which combines the collection of snapshots from the OAV and the determination
+    of the grid dimensions to use for the following grid scan.
+    """
+    yield from setup_beamline_for_OAV(
+        composite.smargon,
+        composite.backlight,
+        composite.aperture_scatterguard,
+        wait=True,
+    )
```
olliesilvester, line 83: I think there was a reason why we had to call this here even though it's called by a later plan. Something to do with the aperture scatterguard not moving properly? Do you remember @jacob720 ? I will add a comment in code once we have the reason...

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+
+    callbacks = create_gridscan_callbacks()
+
+    @bpp.subs_decorator(callbacks)
```
olliesilvester, line 87: I think the reason Hyperion doesn't need this on its equivalent plan is because Hyperion's callbacks run in a separate process. When Hyperion switches to BlueAPI we need to think about how this should work

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+    @verify_undulator_gap_before_run_decorator(composite)
+    @transmission_and_xbpm_feedback_for_collection_decorator(
+        composite, parameters.transmission_frac, PlanNameConstants.GRIDSCAN_OUTER
+    )
+    def grid_detect_then_xray_centre_with_callbacks():
+        yield from grid_detect_then_xray_centre(
```
olliesilvester, line 93: I'm going to try and think of better namings for these functions: `grid_detect_then_xray_centre`  calls `detect_grid_and_do_gridscan`

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+    context: BlueskyContext,
+) -> GridDetectThenXRayCentreComposite:
+    return device_composite_from_context(context, GridDetectThenXRayCentreComposite)
+
+
+def i04_grid_detect_then_xray_centre(
+    composite: GridDetectThenXRayCentreComposite,
+    parameters: GridCommon,
+    oav_config: str = OavConstants.OAV_CONFIG_JSON,
+    udc: bool = False,
```
olliesilvester, line 72: We should have a docstring for this plan. Also, I need to check if GDA knows how to pass in the udc flag

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+    if not udc:
+        yield from get_ready_for_oav_and_close_shutter(
+            composite.smargon,
+            composite.backlight,
+            composite.aperture_scatterguard,
+            composite.detector_motion,
+        )
+
+
+def get_ready_for_oav_and_close_shutter(
```
olliesilvester, line 112: I think we need a comment here to explain why we need to do this when we aren't running UDC - I expect it's to get the beamline in a state which works with GDA?

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+    )
+    yield from bps.abs_set(
+        detector_motion.shutter,
+        0,
+        group=group,
+    )
+    yield from bps.wait(group)
+
+
+def create_gridscan_callbacks() -> tuple[
```
olliesilvester, line 131: Add comment here explaining what the callbacks do

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -0,0 +1,187 @@
...
+)
+from mx_bluesky.common.parameters.device_composites import (
+    GridDetectThenXRayCentreComposite,
+)
+from mx_bluesky.common.parameters.gridscan import GridCommon, SpecifiedThreeDGridScan
+from mx_bluesky.common.preprocessors.preprocessors import (
+    transmission_and_xbpm_feedback_for_collection_decorator,
+)
+from mx_bluesky.common.utils.context import device_composite_from_context
+from mx_bluesky.hyperion.experiment_plans.hyperion_flyscan_xray_centre_plan import (
```
olliesilvester, line 56: We shouldn't be importing from hyperion into i04 like this. I will make an issue to add a linting rule against this.

Since the zebra on i03 and i04 are setup in the same way, we can instead move the zebra tidy and setup plans into a phase1 folder like with the beamstop

src/mx_bluesky/beamlines/i04/parameters/constants.py
```diff
@@ -0,0 +1,10 @@
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class I04Constants:
+    GRAYLOG_PORT = 12232
```
olliesilvester, line 6: Add comment to what stream this port refers to

olliesilvester (commented):
Added a few comments for myself to make changes. I haven't gone through all of the code yet

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
olliesilvester: Turns out this linting rule does already exist, it just didn't pick it up because this directory didn't have an __init__.py

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
olliesilvester: On a second look I don't think this is needed, the same function is called again before we do anything else

src/mx_bluesky/phase1/__init__.py
```diff
@@ -0,0 +1 @@
+"""For plans used in Hyperion and i04 only"""
```
rtuck99, line 1: Maybe expand just a tad more in order to explain that this is what phase1 means as it's something that if you don't know what it's referring to (i.e. historic terminology), is very confusing.

src/mx_bluesky/phase1/device_setup_plans/cleanup_plans.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def gridscan_generic_tidy(
+    xrc_composite: GridscanTidyDevices, group, wait=True
+) -> MsgGenerator:
+    LOGGER.info("Tidying up Zebra")
+    yield from tidy_up_zebra_after_gridscan(
+        xrc_composite.zebra, xrc_composite.sample_shutter, group=group, wait=wait
+    )
+    LOGGER.info("Tidying up Zocalo")
```
rtuck99, line 35: feels like tidying up the zocalo shouldn't be in this package because the code that sets it up isn't also here

Issue #963, opened 2025-03-31
Get parameters from agamemnon for native collections


DominicOram:
In https://github.com/DiamondLightSource/mx-bluesky/issues/864 we are getting the parameters for the first collection. This will mean that we should always be correct for ligand collections and for the first time we're called in native collections. However, we will be incorrect for the second call in a native collection. Instead when there are multiple collections we should return multiple `LoadCentreCollect` objects, that hyperion will then iterate through.

## Acceptance Criteria
- We correctly return two LoadCentreCollect parameters when hyperion is called with a native collection
- The baton handling code iterates through both of them

DominicOram:
We also want to do a phasing collection:
```
{'collect': {'collection': [{'centring_mode': 'XRAY',
                             'chi': 0.0,
                             'comment': 'phasing_sweep1',
                             'distance': 150.0,
                             'dose': 1.05,
                             'experiment_type': 'SAD',
                             'exposure_time': 0.01,
                             'kappa': 0.0,
                             'number_of_images': 3600,
                             'omega_increment': 0.1,
                             'omega_overlap': 0.0,
                             'omega_start': 0.0,
                             'phi_increment': 0.0,
                             'phi_overlap': 0.0,
                             'phi_start': 0.0,
                             'resolution': 1.248485061741461,
                             'scan_axis': 'omega',
                             'transmission': 0.009952731430978554,
                             'two_theta': 0.0,
                             'wavelength': 0.9785655090765587},
                            {'centring_mode': 'XRAY',
                             'chi': 30,
                             'comment': 'phasing_sweep2',
                             'distance': 150.0,
                             'dose': 1.05,
                             'experiment_type': 'SAD',
                             'exposure_time': 0.01,
                             'kappa': 0.0,
                             'number_of_images': 3600,
                             'omega_increment': 0.1,
                             'omega_overlap': 0.0,
                             'omega_start': 0.0,
                             'phi_increment': 0.0,
                             'phi_overlap': 0.0,
                             'phi_start': 0,
                             'resolution': 1.248485061741461,
                             'scan_axis': 'omega',
                             'transmission': 0.009952731430978554,
                             'two_theta': 0.0,
                             'wavelength': 0.9785655090765587},
                            {'centring_mode': 'XRAY',
                             'chi': 15,
                             'comment': 'phasing_sweep3',
                             'distance': 150.0,
                             'dose': 1.05,
                             'experiment_type': 'SAD',
[...]

PR #1055, opened 2025-05-09
Decode agamemnon native collections

rtuck99:
Fixes
* #963 

1. Unit tests pass

### Checks for reviewer


DominicOram (changes requested):
Thanks. Unfortunately I don't think this is going to work. When we change chi we need to do an XRC as the centre of rotation of the gonio isn't perfect. This is why we have the logic in https://github.com/DiamondLightSource/mx-bluesky/blob/8c39aeb0449ef68f5452f1b71844bcb93dc1e9ff/src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py#L132 to check whether chi has changed from the last collection and redo the XRC.

If we specify multiple sweeps for one collection then Hyperion only runs that code once and instead does all the rotations one after each other. Sorry, I should have been clearer on the initial issue. I think there are two fixes:
* We return multiple `LoadCentreCollect` plans from agamemnon. This is what I suggested in the issue and is probably the quickest path to getting something working
* We review how we do `load_centre_collect`, e.g. maybe it checks if chi has changed then redoes the centring if needed. This is going to be a bigger bit of work so I suggest a new issue for it.

tests/test_data/agamemnon/example_native.json
```diff
@@ -0,0 +1,84 @@
+{
```
DominicOram, line 1: Should: This is functionally the same as `example_collect.json`, could we just use one of them for all the tests? I don't mind which

DominicOram (approved):
Great, thank you!

PR #1059, opened 2025-05-12
Make GridCommon class concrete

### LINKED ISSUE #1058: Make GridCommon a concrete class
`GridCommon` is currently an abstract class, however it needs to be concrete as it is used as the parameters type for the common and i04 grid detect then xray centre plans. This means it needs to contain `detector_params` which it currently doesn't.

## Acceptance Criteria
- `GridCommon` becomes a concrete class
- Code is not repeated in `GridCommonWithHyperionDetectorParams`

jacob720:
Fixes #1058

Added `detector_params` to `GridCommon` to make it a concrete class, and changed `GridCommonWithHyperionDetectorParams`and `HyperionSpecifiedThreeDGridScan` to use the `detector_params` they now inherit.

1. Confirm this is the right thing to do
2. Check Hyperion classes still function as intended

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/parameters/gridscan.py
+++ b/src/mx_bluesky/common/parameters/gridscan.py
@@ -29,7 +29,7 @@ from mx_bluesky.common.parameters.constants import (
 DETECTOR_SIZE_PER_BEAMLINE = {
     "i02-1": EIGER2_X_9M_SIZE,
     "dev": EIGER2_X_16M_SIZE,
-    "I03": EIGER2_X_16M_SIZE,
+    "i03": EIGER2_X_16M_SIZE,
 }
 
 
--- a/src/mx_bluesky/common/parameters/gridscan.py
+++ b/src/mx_bluesky/common/parameters/gridscan.py
@@ -185,34 +185,3 @@ class SpecifiedThreeDGridScan(
     @property
     def num_images(self) -> int:
         return len(self.scan_points["sam_x"])
-
-    @property
-    def detector_params(self):
-        self.det_dist_to_beam_converter_path = (
-            self.det_dist_to_beam_converter_path
-            or DetectorParamConstants.BEAM_XY_LUT_PATH
-        )
-        optional_args = {}
-        if self.run_number:
-            optional_args["run_number"] = self.run_number
-        assert self.detector_distance_mm is not None, (
-            "Detector distance must be filled before generating DetectorParams"
-        )
-        return DetectorParams(
-            detector_size_constants=DETECTOR_SIZE_PER_BEAMLINE[
-                get_beamline_name("dev")
-            ],
-            expected_energy_ev=self.demand_energy_ev,
-            exposure_time_s=self.exposure_time_s,
-            directory=self.storage_directory,
-            prefix=self.file_name,
-            detector_distance=self.detector_distance_mm,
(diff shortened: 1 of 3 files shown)


src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -26,7 +26,11 @@
     HardwareConstants,
 )
 
-DETECTOR_SIZE_PER_BEAMLINE = {"i02-1": EIGER2_X_9M_SIZE, "dev": EIGER2_X_16M_SIZE}
+DETECTOR_SIZE_PER_BEAMLINE = {
+    "i02-1": EIGER2_X_9M_SIZE,
+    "dev": EIGER2_X_16M_SIZE,
+    "I03": EIGER2_X_16M_SIZE,
```
DominicOram, line 32: Must: The `BEAMLINE` env for i03 is `i03` not `I03`

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -26,7 +26,11 @@
     HardwareConstants,
 )
 
-DETECTOR_SIZE_PER_BEAMLINE = {"i02-1": EIGER2_X_9M_SIZE, "dev": EIGER2_X_16M_SIZE}
+DETECTOR_SIZE_PER_BEAMLINE = {
```
DominicOram, line 29: Could: This is messy to have here, it would be better to have some function like `get_beamline_constants` and that would give you an object like `I03Constants` for each beamline

DominicOram: But this may make sense to spin into a new issue, feel free to write one up

DominicOram (changes requested):
Looks good, one comment. Additionally:
* It's not 100% clear why you need this? Presumably there is a line in some follow on PR where you are constructing a `GridCommon`?
* Presumably we can replace the code in `SpecifiedThreeDGridScan` too?

src/mx_bluesky/common/parameters/gridscan.py
jacob720: I initially had it lower case, but was thrown off by [this test](https://github.com/DiamondLightSource/mx-bluesky/blob/4a7c282eeba3eb0cc564785aecd395596b4ba5f1/tests/unit_tests/hyperion/external_interaction/nexus/test_write_nexus.py#L346) failing. I'll change the test as it has the `BEAMLINE` env as `I03`.

jacob720:
This is needed as a `GridCommon` needs to be constructed for the `i04_grid_detect_then_xray_centre` plan. Currently this is constructed in a script which is unlikely to be merged, but I assume even once we do away with the script a `GridCommon` will have to be constructed somewhere.

I hadn't spotted the `SpecifiedGridScan`, thanks for letting me know, I'll remove `DetectorParams` from there too.

src/mx_bluesky/common/parameters/gridscan.py
jacob720: That makes sense, I'll create an issue

jacob720: Sorry forgot to actually post it yesterday, here it is https://github.com/DiamondLightSource/mx-bluesky/issues/1068

PR #1061, opened 2025-05-12
Removed references to collection_time

### LINKED ISSUE #1050: I24 SSX: Use frametime instead of collection time
Following https://github.com/DiamondLightSource/dodal/pull/1216 we are now using the time between frames to work out if the collection has stalled instead of the full collection time. We should remove using this in the i24 SSX code.

## Acceptance Criteria
- `calculate_collection_timeout` should instead calculate the expected time between frames and set this on the device
- The `collection_time` soft signal in the device should be removed

adaudon:
Fixes #1050

Link to dodal PR (if required): DiamondLightSource/dodal#1226

1. Confirm tests past

### Checks for reviewer


DominicOram (approved):
This looks good, thank you. The original issue talks about using a frame time instead of a collection time but this just removes collection time without replacing. I know we discussed this offline and decided that the flat 30s between frames was good for most collections and so the complexity of calculating things wasn't worth the effort. In the future, when requirements change like this it's good to put a note either on the original issue or on the PR about why they changed, who was in the discussion etc.

Ignoring code cov as this entirely removes code.

PR #1062, opened 2025-05-12
Create setup beamstop plan

### LINKED ISSUE #1060: Create plan to move beamstop into position
> Notes from 9th May testing
> - Need to put beamstop into correct position at the start of the plan. We were doing this manually

During testing on i04, we used GDA to move the beamstop into position. I think on i03 the beamstop stays permanently in position so this wasn't needed for Hyperion, but on i04 it moves sometimes. We should create a plan to move the beamstop into position for i04 to use.

## Acceptance Criteria
- Plan that moves beamstop into position is created
- There are tests confirming this plan works

jacob720:
Fixes #1060 

Link to dodal PR (if required): #[1227](https://github.com/DiamondLightSource/dodal/pull/1227)

Adds a plan to move the beamstop into position so that it's blocking the beam. We found during testing on i04 that some GDA plans left the beamstop out of position, meaning the grid detect then xray centre plan would crash at the point the `check_beamstop` plan was run.

1. Confirm the plan moves the beamstop to the correct position
2. Confirm the tests are thorough enough

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.10.0a2",
     "bluesky >= 1.13.1",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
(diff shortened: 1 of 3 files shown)


jacob720:
No longer needed as it's being implemented on the dodal side

Issue #1039, opened 2025-05-02
Move debug logs to `/dls/tmp`


DominicOram:
Despite rolling over, debug logs can get quite big. To help manage disk space we should put them in `/dls/tmp`

## Acceptance Criteria
- Debug logs go in a suitable folder in `/dls/tmp`
- This is documented

Issue #751, opened 2025-01-16
Running run_hyperion in dev mode can break live deployment


olliesilvester:
When playing around with Hyperion using the `run_hyperion.sh` script in dev mode, we managed to create log files in the Hyperion live deployment area which were owned by our own user, rather than gda2. This meant that when Hyperion was running on i03, gda2 didn't have access to write to these logs and failed.

This resulted in callbacks failing to start, so data wasn't pushed to ISPyB, and so Zocalo was never triggered. The actual error we saw in Hyperion was a timeout on waiting for Zocalo results.

We need to take a look at the `run_hyperion` script and decide if it's still a useful thing to have, and if so, to make sure that these logs cannot be created in the live location with the incorrect permissions

## Acceptance Criteria
- `run_hyperion` makes it impossible to create files as a non-gda2 user in the deployment area

rtuck99:
`run_hyperion.sh` is used by `hyperion_restart()` so we cannot remove it just yet

rtuck99:
I found the original slack conversation and it seems that at least one of the instances was when I needed to test something and launched hyperion directly with `--dev`, without going via `run_hyeperion.sh`.
`_get_logging_dirs` will then check `BEAMLINE` in the event LOG_DIR is not set and use the production logging dir.

I suspect other instances may have occurred during refactoring to rename `HYPERION_LOG_DIR` to `LOG_DIR` and a mismatch between the launch script and the main hyperion code in terms of the variable naming.

Since LOG_DIR will always be specified when hyperion is launched in production, the changes for #1042 will also not work in production as `debug_logging_path` will always be the same as `logging_path`

DominicOram:
Fixed by https://github.com/DiamondLightSource/mx-bluesky/pull/1063

PR #1063, opened 2025-05-13
751 fix run hyperion in dev mode

rtuck99:
Fixes

* #751 

Note that the issue was likely caused by launching hyperion directly with the --dev parameter, not via `run_hyperion.sh`

Changes to the `run_hyperion.sh` script consist of the following minor tidyups/fixes:
* Better documentation of options, and don't defer to python for generating the first bit of the help text
* Always specify `ZOCALO_CONFIG` and `ISPYB_CONFIG` with `--dev` so that we don't try to talk to production servers
* Require BEAMLINE to be always set in the environment or specified as a parameter

This PR also changes `--dev-mode` so that now it creates mock devices instead of real ones, ensuring that hyperion can be started without access to the beamline network.

This also addresses some issues arising from

* #1039 

* DEBUG_LOG_DIR can now be specified for use by containers
* the default debug logging dir of /dls/tmp/{beamline} is now applied even when LOG_DIR is specified
* if --dev-mode is specified, when launching hyperion the production logging directories will not be applied

1. Debug logging dirs are now applied when hyperion is started from GDA or in a container
2.  if --dev-mode is specified, when launching hyperion the production logging directories will not be applied
3. --dev-mode now works from a desktop machine

### Checks for reviewer

### DIFF
--- a/docs/developer/hyperion/reference/readme.md
+++ b/docs/developer/hyperion/reference/readme.md
@@ -25,14 +25,17 @@ Note that because Hyperion makes heavy use of [Dodal](https://github.com/Diamond
 
 ## Starting the bluesky runner
 
-The bluesky runner is (re)started in production by GDA when it invokes `run_hyperion.sh`. 
+The bluesky runner is (re)started in production by GDA when it invokes `run_hyperion.sh`.
+
+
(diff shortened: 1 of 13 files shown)


DominicOram (approved):
Great, thank you. Couple of minor comments, take them or leave them

Issue #284, opened 2024-02-16
Investigate why Hyperion tries to connect to devices after already being connected


olliesilvester:
Hyperion raised a timeout error when an ophyd_async device failed to connect in time. After looking at https://graylog2.diamond.ac.uk/messages/logs_8652/cd356670-cc46-11ee-a034-1866dafadc6c, it looks like Hyperion was running its setup plans when it had already been running for a while. This part of the code should only be running once, when Hyperion is first started

DominicOram:
Closed by https://github.com/DiamondLightSource/mx-bluesky/pull/1064

Issue #253, opened 2024-05-08
`--skip-startup-connection` is broken


d-perl:
When running `env BEAMLINE=i03 ./run_hyperion.sh --dev --skip-startup-connection` on a fresh install of Hyperion main, I get:

```
╰─>$ cat tmp/dev/start_log.log
    1 INFO - Using nexgen 0.9.1
    2 adding handler <StreamHandler <stderr> (INFO)> to logger <Logger Dodal (DEBUG)>, at level: 20
    3 adding handler <GELFTCPHandler (INFO)> to logger <Logger Dodal (DEBUG)>, at level: 20
    4 Logging to tmp/dev/hyperion.log
    5 adding handler <TimedRotatingFileHandler /scratch/ziq44869/test_hyperion_1357/hyperion/tmp/dev/hyperion.log (INFO)> to logger <Logger Dodal (DEBUG)>, at level: 20
    6 Logging to tmp/dev/hyperion.log
    7 adding handler <CircularMemoryHandler (DEBUG)> to logger <Logger Dodal (DEBUG)>, at level: 10
    8 adding handler <StreamHandler <stderr> (INFO)> to logger <Logger Hyperion ISPyB and Zocalo callbacks (DEBUG)>, at level: 20
    9 adding handler <GELFTCPHandler (INFO)> to logger <Logger Hyperion ISPyB and Zocalo callbacks (DEBUG)>, at level: 20
   10 Logging to tmp/dev/hyperion_ispyb_callback.log
   11 adding handler <TimedRotatingFileHandler /scratch/ziq44869/test_hyperion_1357/hyperion/tmp/dev/hyperion_ispyb_callback.log (INFO)> to logger <Logger Hyperion ISPyB and Zocalo callbacks (DEBUG)>, at level: 20
   12 Logging to tmp/dev/hyperion_ispyb_callback.log
   13 adding handler <CircularMemoryHandler (DEBUG)> to logger <Logger Hyperion ISPyB and Zocalo callbacks (DEBUG)>, at level: 10
   14 adding handler <StreamHandler <stderr> (INFO)> to logger <Logger Hyperion NeXus callbacks (DEBUG)>, at level: 20
   15 adding handler <GELFTCPHandler (INFO)> to logger <Logger Hyperion NeXus callbacks (DEBUG)>, at level: 20
   16 Logging to tmp/dev/hyperion_nexus_callback.log
[...]

PR #1064, opened 2025-05-13
Remove --skip-startup-connection startup option and associated device warmup

rtuck99:
This removes the `--skip-startup-connection` option, which no longer does anything, since 

* https://github.com/DiamondLightSource/blueapi/pull/751
which implements ADR

https://diamondlightsource.github.io/blueapi/main/explanations/decisions/0005-connect-devices.html

This also removes the preloading of devices that occurs by calling the `create_devices` methods of all the entry points, since standard blueapi beamline loading already connects all the devices.

Also solves the following outdated issues:
* #253 
* #284 

1. Tests pass
2. mx-bluesky still launches

### Checks for reviewer

### DIFF
--- a/docs/developer/hyperion/reference/readme.md
+++ b/docs/developer/hyperion/reference/readme.md
@@ -25,9 +25,14 @@ Note that because Hyperion makes heavy use of [Dodal](https://github.com/Diamond
 
 ## Starting the bluesky runner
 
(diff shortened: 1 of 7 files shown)


DominicOram (approved):
Great, thank you. Happy to merge w/o the codecov passing, it's hard to test this stuff

PR #1067, opened 2025-05-13
Convert numpy types for ispyb

### LINKED ISSUE #1066: Convert numpy floats to Python floats before ispyb
> Notes from 9th May testing

> - Had to change `_upsert_data_collection` in [ispyb_store.py](https://github.com/DiamondLightSource/mx-bluesky/blob/main/src/mx_bluesky/common/external_interaction/ispyb/ispyb_store.py) to convert `np.float64` types to Python `float` type to fix an SQL `ProgrammingError` error.

Ispyb throws an error when you try to input a numpy float. In testing, I found that xbeam and ybeam values in a `DataCollectionInfo` object were causing this error as they were `np.float64` types. To fix this, we could convert any numpy data types to native Python types before they are entered into ispyb.

## Acceptance Criteria
- An error is no longer thrown when an instance of DataCollectionInfo contains numpy data types

jacob720:
Fixes #1066 

Converts numpy data types to native python types before sending to ispyb. This allows data collection info to contain numpy numbers without causing an error.

1. Send data collection params containing floats to ispyb.
2. Confirm this works with no error

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/external_interaction/ispyb/ispyb_store.py
+++ b/src/mx_bluesky/common/external_interaction/ispyb/ispyb_store.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING
 
 import ispyb
 import ispyb.sqlalchemy
+import numpy as np
 from ispyb.connector.mysqlsp.main import ISPyBMySQLSPConnector as Connector
 from ispyb.sp.mxacquisition import MXAcquisition
 from ispyb.strictordereddict import StrictOrderedDict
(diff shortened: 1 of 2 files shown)


rtuck99 (approved):
These changes look fine. I think I probably introduced this issue when I made changes a while back to the lookup table stuff.

Whilst I think that technically the most type-correct thing to do is fix the problem at source when the offending numpy type is returned from `get_beam_position_mm` there is a certain element of whack-a-mole with numpy type pollution.
As far as I know the ispyb API is the only thing that actually cares about the numpy scalar vs native type issue, pretty much everything else doesn't seem to care. 

numpy arrays on the other hand do exhibit strange behaviour when using operations like multiply etc. because they implement them differently, so they are not drop-in replacements.

PR #1069, opened 2025-05-14
Create setup beamstop plan

### LINKED ISSUE #1060: Create plan to move beamstop into position

jacob720:
Fixes #1060

Link to dodal PR (if required): [#1227](https://github.com/DiamondLightSource/dodal/pull/1227)

Uses the changes to beamstop in the dodal PR above to move the beamstop into position in every plan. If the beamstop is already in position, nothing should happen and this should add no time. The `check_beamstop` plan is still in place after this so if the beamstop is not in place at that point, this will raise an exception.

1. Move beamstop out of position
2. Confirm beamstop moves back into position before xray centring

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/hyperion/device_setup_plans/utils.py
+++ b/src/mx_bluesky/hyperion/device_setup_plans/utils.py
@@ -8,8 +8,8 @@ from dodal.devices.detector import (
 )
 from dodal.devices.detector.detector_motion import DetectorMotion, ShutterState
 from dodal.devices.eiger import EigerDetector
-from dodal.devices.i03 import Beamstop
 from dodal.devices.i03.dcm import DCM
+from dodal.devices.mx_phase1.beamstop import Beamstop, BeamstopPositions
 
 from mx_bluesky.common.device_setup_plans.check_beamstop import check_beamstop
 from mx_bluesky.common.device_setup_plans.position_detector import (
@@ -52,6 +52,9 @@ def start_preparing_data_collection_then_do_plan(
         yield from set_shutter(detector_motion, ShutterState.OPEN, group)
         yield from plan_to_run
 
+    yield from bps.abs_set(
+        beamstop.selected_pos, BeamstopPositions.DATA_COLLECTION, wait=True
+    )
     yield from check_beamstop(beamstop)
     yield from bpp.contingency_wrapper(
         wrapped_plan(),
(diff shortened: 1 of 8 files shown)


DominicOram (changes requested):
Thanks! Two comments:
* I think we should maybe remove `check_beamstop` now. If we're moving it in, assuming it doesn't fail to move (which would be caught by an exception anyway) the check will always pass
* The move could take some time so I suggest we put it in the `wrapped_plan` after we start arming the Eiger

jacob720:
That makes sense, I've done that now. It looks like the `check_beamstop` plan is never actually used now. We could delete it? But maybe it will come in handy in the future

DominicOram:
Additionally, could you change the deleted tests to instead confirm that those things are moving the beamstop in before collection?

src/mx_bluesky/hyperion/device_setup_plans/utils.py
```diff
@@ -45,14 +44,16 @@ def start_preparing_data_collection_then_do_plan(
 
     def wrapped_plan():
         yield from bps.abs_set(eiger.do_arm, 1, group=group)  # type: ignore # Fix types in ophyd-async (https://github.com/DiamondLightSource/mx-bluesky/issues/855)
+        yield from bps.abs_set(
+            beamstop.selected_pos, BeamstopPositions.DATA_COLLECTION, wait=True
```
DominicOram, line 48: Should: Sorry, bit pedantic but can you put this in the same `group` as the other bits? we don't actually need it to be in until just before collection

DominicOram (changes requested):
Thanks, one comment, sorry I missed it earlier

src/mx_bluesky/hyperion/device_setup_plans/utils.py
jacob720: That makes sense, have done

PR #1079, opened 2025-05-20
Fix system test failure caused by beamstop changes

rtuck99:
Fixes failure of the system tests to execute because the `check_beamstop` plan is no longer present
Also removes the pyright flag `reportMissingImports = false` which was causing the broken import to not be reported by `tox -e type-checking`

1. Tests pass
2. System tests pass

### Checks for reviewer


jacob720 (approved):
Looks good, thanks!

PR #1084, opened 2025-05-21
i23 Laser Shaping robot: Check and set beamline_safe

### LINKED ISSUE #939: Set beamline safe via bluesky/blueapi
By getting the RBVs of a bunch of PVs https://github.com/DiamondLightSource/aithre/blob/e6bcfa3915a6df59bf609681c6a3258c3a4a07c1/bin/guiv4_prod.py#L616 we can determine if the goniometer is in the correct position for the robot to mount or unmount a sample. This is currently done by forcing a pv on with caput. This should be converted to bluesky.

JamesDoingStuff:
Fixes #939

Link to dodal PR: [#1131](https://github.com/DiamondLightSource/dodal/pull/1131)

1. Confirm new tests pass

### Checks for reviewer


src/mx_bluesky/beamlines/aithre_lasershaping/beamline_safe.py
```diff
@@ -0,0 +1,29 @@
+import bluesky.plan_stubs as bps
+from bluesky.utils import MsgGenerator
+from dodal.common import inject
+from dodal.devices.aithre_lasershaping.goniometer import Goniometer
+from dodal.devices.aithre_lasershaping.laser_robot import ForceBit, LaserRobot
+
+
+def check_beamline_safe(
```
DominicOram, line 8: Should: I think the plan name should make it clear that this is also forcing it on the robot. Maybe `set_beamline_safe_on_robot` is better?

DominicOram, line 8: Should: This is also weird to do from DAQ, it's usually a controls job. Can we put the reasoning behind it as a docstring on the plan?

DominicOram (changes requested):
Thanks, couple of suggestions in the code

Issue #986, opened 2025-04-11
Write scientist facing docs on how to support hyperion


DominicOram:
To allow scientists to do basic edits or front line support we should write a guide on how to support Hyperion.

Things to include:
* Where code is deployed
* How to see what's running
* How to switch versions
* Where logs live
* Link to the code map
* * Where common parameters live and how they're read (mention the config server)

## Acceptance Criteria
- There is a guide that @ndg63276 and @npat113 are happy with

DominicOram:
@aragaod would also be interested in this. Send to him too.

DominicOram:
I think this is covered by https://github.com/DiamondLightSource/mx-bluesky/pull/1229, of course we will need to iterate on the docs but the we now have a good starting point

PR #1086, opened 2025-05-23
Add ability to email support staff

### LINKED ISSUE #394: Ability to alert support staff
Currently we rely on GDA to inform support staff if there is an issue. At some point we will need to do this directly from Artemis

rtuck99:
Fixes #394 

Link to gitlab merge request for helmcharts for alertmanager:
~https://gitlab.diamond.ac.uk/yqf46943/mx-bluesky-alertmanager-sandbox~
https://gitlab.diamond.ac.uk/controls/containers/beamline/i03-services/-/merge_requests/19

This PR provides the ability to send alert notifications via a backend which talks to a remote service.

There are two backends provided:

* A log based backed which can be used to provide notifications using graylog when suitable Event and Notification types are configured. This essentially creates a log message with specific metadata that is picked up by Graylog as fields in the log message, that satisfy a filter in the graylog Event 
* An alertmanager based backend which uses the AlertManager API to raise an alert. Note that the alertmanager API requires alerts to be re-raised while they are active; this backend doesn't do this so an alert notification will be sent once and then the alert will be discarded.
* Initially I intended to have nicely formatted alertmanager alert emails, but have decided against for the time being and therefore reverted to the default alertmanager alerts (the metadata, links etc are still visible and the subject line is still specified), on the grounds that nice formatting for mobile devices etc would be additional effort, however I've got as far as setting things up in the helm charts so that it is possible

The `AlertService` allows a plain-text summary and content to be specified which can be used to generate the email subject and alert message. Formatting of the notification is up to configuration within the particular service.

There are some things which I would like an opinion on:

1. Initialisation of the alert service - I have tentatively put this in the `baton_handler` as ideally we would only need to set the  initialise this once, but at the moment we don't have any dedicated place for global initialisation. Also hyperion as it is currently run from __main__ doesn't initialise the service - but perhaps we want this for initial testing?
2. Computed metadata in the backend or at the alerting callsite: As different backends seem to have different p~references for how much metadata they like (prometheus particularly seems to not like high cardinality labels), I've decided to minimise the number of fields the caller inputs, thus dependent values such as URLs can be computed in hyperion or if the backend service doesn't like this, generated in the notification template on the remote service.
3. Configuration for authentication - there is provision for specifying this, but I haven't written any code to actually obtain the credentials. Potentially no authentication will be needed if we only run in a container - initially I anticipated the alertmanager service would be authenticated but this currently does not seem to be well supported with the configReloader or the standard helmcharts.

1. Do thing x
4. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/common/external_interaction/callbacks/sample_handling/sample_handling_callback.py
```diff
@@ -51,10 +62,24 @@ def activity_gated_stop(self, doc: RunStop) -> RunStop:
...
         return doc
 
-    def _record_exception(self, exception_type: str, expeye: ExpeyeInteraction):
+    def _record_exception(
+        self, exception_type: str, expeye: ExpeyeInteraction, reason: str
+    ):
         assert self._sample_id, "Unable to record exception due to no sample ID"
         sample_status = self._decode_sample_status(exception_type)
         expeye.update_sample_status(self._sample_id, sample_status)
+        if sample_status == BLSampleStatus.ERROR_BEAMLINE:
```
DominicOram, line 71: Should: This feels very tagged on to the sample handling stuff. Whilst it does add a bit more boilerplate it might be better to have it as a separate calback. i.e. we could want alerting without writing sample status or vice versa

tests/system_tests/hyperion/external_interaction/alerting/test_alertmanager_alerts.py
```diff
@@ -0,0 +1,27 @@
+import pytest
+
+from mx_bluesky.common.external_interaction.alerting.alert_manager import (
+    AlertManagerAlertService,
+)
+
+ENDPOINT = "https://yqf46943-alertmanager.diamond.ac.uk"
```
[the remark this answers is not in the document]
rtuck99: So this is just the name that I was using when testing on my personal namespace on argus - if we want this to be generally runnable as a system test we would need to add alertmanager into the system test stack. The hostname is as suggested in the dev portal but for my namespace
https://dev-portal.diamond.ac.uk/guide/kubernetes/tutorials/monitoring/

src/mx_bluesky/common/external_interaction/alerting/alert_manager.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _session(self) -> requests.Session:
+        session = requests.Session()
+        if self._username and self._password:
+            session.auth = HTTPBasicAuth(self._username, self._password)
+        session.headers["Accept"] = "application/json"
+        return session
+
+    def _ispybUrl(self, sample_id: str):
+        return f"https://ispyb.diamond.ac.uk/samples/sid/{quote(sample_id)}"
```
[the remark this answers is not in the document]
rtuck99: So I did think about this, but I can envisage that there might be a problem that happens before we've created a data collection (granted this is unlikely as it does get created quite early but it is still a possibility), so I decided to base it on information that we are guaranteed to always have in the request. You should be able to find the data collection if there is one as it will be the most recent associated with the sample.

The other problem with the data collection is that the DCID is held in the ispyb callbacks, whereas the logical place to raise the alert is from the `SampleHandlingCallback` so we would have to have some way of fetching that information.

src/mx_bluesky/common/external_interaction/callbacks/sample_handling/sample_handling_callback.py
rtuck99: I can move it to a separate callback if needed. I think there is some connection though, as the BLSampleStatus is going to feed back into agamemnon, and so therefore beamline staff will want every event that causes agamemnon to stop (i.e. ERROR_BEAMLINE) to raise an event.

Also, the intention is that ultimately this will not be the only place that calls the alerting service - the way I intend it to be used is that you can call the alerting service from any part of the code, we will have different alert categories and potentially different types of metadata / messages.

I certainly don't think it would be appropriate to pipe every call to the alerting service through this callback.

rtuck99:
>     * For instantiation, what about if we make it a proper singleton and instantiate it on first call to `get_graylog_configuration`, if it has not yet been instantiated. It feels like the kind of thing where it's ok to be a global singleton
> 
Unfortunately `get_graylog_configuration` is in `dodal` so we can't put the call there.

rtuck99:
>     * The graylog method seems so much simpler, as evidenced by the fact I quite easily managed to test it. However, the emails I was getting with it did not fill in the `alert_summary` or `alert_content`, any idea why?
> 

So the reason that the alert messages didn't contain the summary or content was because the example ones I had weren't quite set up correctly. In the Events configuration tab, you need to enter each of the fields that will appear in the notification:

image

Also in the notification configuration, the event field then needs to be referenced correctly, e.g. `${event.fields.alert_summary}`

src/mx_bluesky/hyperion/external_interaction/config_server.py
```diff
@@ -19,6 +19,7 @@ class HyperionFeatureFlags(FeatureFlags):
             multi-centre)
         omega_flip:                     If True then invert the smargon omega motor rotation commands with respect to
          the hyperion request. See "Hyperion Coordinate Systems" in the documentation.
+        alert_manager_url:              Base URL for alertmanager, if using for alerting.
```
DominicOram, line 22: Nit: Remove as no longer used

src/mx_bluesky/hyperion/external_interaction/config_server.py
```diff
@@ -32,4 +33,5 @@ def get_config_server() -> ConfigServer:
     use_gpu_results: bool = CONST.I03.USE_GPU_RESULTS
     set_stub_offsets: bool = CONST.I03.SET_STUB_OFFSETS
     omega_flip: bool = CONST.I03.OMEGA_FLIP
+    alert_manager_url: str = CONST.I03.ALERT_MANAGER_URL
```
DominicOram, line 36: Nit: Remove as no longer used

src/mx_bluesky/hyperion/parameters/constants.py
```diff
@@ -27,6 +27,7 @@ class I03Constants:
     USE_PANDA_FOR_GRIDSCAN = False
     SET_STUB_OFFSETS = False
     OMEGA_FLIP = True
+    ALERT_MANAGER_URL = "https://CONFIGURE_ME"
```
DominicOram, line 30: Nit: Remove as no longer used

tests/system_tests/hyperion/external_interaction/alerting/test_log_based_alerts.py
```diff
@@ -0,0 +1,17 @@
...
+from dodal.log import get_graylog_configuration, set_up_graylog_handler
+
+from mx_bluesky.common.external_interaction.alerting.log_based_service import (
+    LoggingAlertService,
+)
+
+
+@pytest.mark.system_test
+def test_alert_to_graylog():
+    logger = logging.getLogger("Dodal")
```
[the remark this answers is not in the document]
DominicOram: Sorry, what I meant is the stream the event is pointing to in graylog is wrong. It should be pointing to the Hyperion stream and this test should therefore be using the graylog port as defined in `HyperionConstants.GRAYLOG_PORT`.

DominicOram (changes requested):
Thanks! Minor nit-picks. Otherwise, could we put a page in the docs on how the alerting works e.g. links on how I would:
* Change the address list
* Change the email format
If we can write something to the point where beamline scientists could edit those things that would be great. Finally, I think we can probably rename all the events/notifications to drop the test part as it's ready to use in prod

docs/user/hyperion/logging.rst
```diff
@@ -0,0 +1,42 @@
...
+
+The log file location is controlled by the ``LOG_DIR`` environment value. Typically this can be found in 
+``/dls_sw/<beamline>/logs/bluesky``
+
+
+Graylog
+~~~~~~~
+
+Graylog is used for more centralised logging which is also more easily searched and archived. Log messages are sent 
+to the graylog endpoint - this is hardcoded to use port 12232 which corresponds to the Hyperion graylog stream.
```
DominicOram, line 31: Nit: Add the URL to the logs here, more important to an end user than the port. Also probably push this to the top as the "recommended" way of viewing the logs

DominicOram (approved):
Brilliant, thank you! Thanks for adding a bunch of docs too, means we're half way to doing https://github.com/DiamondLightSource/mx-bluesky/issues/986. I've put one minor comment in, I will address it myself

PR #1090, opened 2025-05-27
Update i24 blueapi configuration

noemifrisina:
Updates the blueapi configuration for i24 serial to work with the latest changes in blueapi.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


noemifrisina:
I usually just start up a blueapi server with `blueapi -c blueapi_config.yaml` and check that the stomp configuration works - which is the only configuration change from newer versions blueapi. (Warning that doing this on a non-beamline machine will show every device as not connected)

For an automated test, I honestly am not sure. There's tests for the stomp configuration in blueapi already and most of what I can think of would just duplicate those...

PR #1093, opened 2025-05-27
I24 serial: fix checker pattern value

### LINKED ISSUE #1092: I24 serial: checker pattern always seems to be on
When running a fixed target pump probe collection, the "checker pattern" seems to always be set to True, even when left unchecked on the edm.
Additionally, if checker_pattern=True, the corresponding P-variable is set twice: first to 0 and then to 1. It should only be set once.

## Acceptance Criteria
- In the parameter model checker pattern is set correctly and the corresponding P variable is only set once

noemifrisina:
Fixes #1092 

1. Run tests

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
+++ b/src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
@@ -172,7 +172,7 @@ def read_parameters(
         "chip": chip_params.model_dump(),
         "map_type": map_type,
         "pump_repeat": pump_repeat,
-        "checker_pattern": bool(caget(pv.me14e_gp111)),
+        "checker_pattern": bool(int(caget(pv.me14e_gp111))),
         "chip_map": chip_map,
         "laser_dwell_s": float(caget(pv.me14e_gp103)) if pump_repeat != 0 else 0.0,
         "laser_delay_s": float(caget(pv.me14e_gp110)) if pump_repeat != 0 else 0.0,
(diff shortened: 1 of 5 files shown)


noemifrisina:
The logic in `load_motion_program_data` has slightly changed in the sense that before we were setting the checker pattern P variable twice if checker pattern was selected - the first to 0 like for every other experiment and then again to 1 - and now we always only set it once. The test has been changed to reflect this.

Yes, the value is now cast to an `int`when reading because the problem was exactly what you described... It was doing `bool("False") = True`. I'll slightly refactor the `read_parameters` function to add the test because it currently has 12 calls to `caget` and we'd need to know exactly which one it is - which doesn't make for a great test.

DominicOram:
>The logic in `load_motion_program_data` has slightly changed in the sense that before we were setting the checker pattern P variable twice if checker pattern was selected - the first to 0 like for every other experiment and then again to 1 - and now we always only set it once. The test has been changed to reflect this.

Sorry, I meant more that that isn't where the actual fix for this lies.

> I'll slightly refactor the read_parameters function to add the test because it currently has 12 calls to caget and we'd need to know exactly which one it is

Thanks!

src/mx_bluesky/beamlines/i24/serial/fixed_target/i24ssx_Chip_Manager_py3v1.py
```diff
@@ -172,7 +178,7 @@ def read_parameters(
         "chip": chip_params.model_dump(),
         "map_type": map_type,
         "pump_repeat": pump_repeat,
-        "checker_pattern": bool(caget(pv.me14e_gp111)),
+        "checker_pattern": _is_checker_pattern(),  # bool(int(caget(pv.me14e_gp111))),
```
DominicOram, line 181: Should: Remove commented out code

DominicOram (approved):
Great, thanks. Just one comment, after that feel free to merge

Issue #1097, opened 2025-05-30
Migrate to new daq-config-server


olliesilvester:
For `daq-config-server` > 0.1.1 has a very different interface to previous versions. We should unpin it when ready to use it in Hyperion

We need to take into account that some beamlines (eg i03) will not want to rely on the central argus cluster initially. At some point, probably in a future ticket, we should deploy an instance of the config server onto i03's cluster, so Hyperion can use this one.

## Acceptance Criteria
- daq-config-server unpinned
- Our bluesky plans are updated so that they can optionally use the central config server, else attempt to read the file system directly

PR #1098, opened 2025-05-30
Pin daq config server

olliesilvester:
I will soon be making a release for the daq-config-server, which will break `mx-bluesky` in quite a few places, this PR is pre-emptively pinning it to the compatible version. See https://github.com/DiamondLightSource/mx-bluesky/issues/1097 for unpinning.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


rtuck99 (approved):
Approved

PR #1103, opened 2025-06-03
#1025 Fix system tests interfering with unit tests when run locally

### LINKED ISSUE #1025: Running system tests locally can cause unit tests to fail locally
Running the new system tests locally creates nexus files in `/tmp/`. A few unit tests fail if we have existing files in this directory, since `get_run_number` will see these files and change the number from what we have asserted.

We should make sure that any file creation is cleaned up in system tests, probably by using the `tmp_path` fixture

## Acceptance Criteria
- File creation is always cleaned up after tests
- (Bonus) Tests never write files to anywhere except the `tmp_path` fixture

rtuck99:
Fixes #1025 

This introduces a `tmp_path` parameter in `raw_params_from_file()` that must be supplied to expand `{tmp_data}` references in any supplied json file, to the pytest tmp_path folder.

This should prevent temporary files being created in the `/tmp` root directory that interfere across tests

The `TestData` class event attributes were moved into a new class `TestEventData` which is obtainable as a pytest fixture with directories substituted accordingly.

The test utility method `replace_all_tmp_paths()` can be called in tests to perform the required path substitutions on expected output data dicts.

The unused constant `BASE_DATA_DIR` was removed.

1. Unit tests can be run locally and system tests locally without interference.

### Checks for reviewer


DominicOram (approved):
Great, thanks! Lots of lines changed for something quite small in the end!

PR #1108, opened 2025-06-06
Add monitoring functions for goniometer and robot RBV, and implement …

co2e14:
…go_to_zero function

PR #1111, opened 2025-06-10
Implement go_to_zero function to reset goniometer stages for robot lo…

co2e14:
…ad/unload

Fixes first issue with previous PR #1108 

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/beamlines/aithre_lasershaping/beamline_safe.py
```diff
@@ -34,3 +34,19 @@ def set_beamline_safe_on_robot(
         else ForceBit.NO.value
     )
     yield from bps.abs_set(robot.set_beamline_safe, set_value, wait=True)
+
```
rtuck99, line 37: The logic is fine, however you may want to take a look at `home_and_reset_wrapper()` (and the corresponding decorator) if you plan to use this as part of an overall robot load plan.
This is how we are doing robot load for i03 - the wrapper can wrap the actual robot load plan, then returns the smargon to the original position, look inside `robot_load_and_change_energy` to see how we use it.

co2e14: I think it will take me a while to get my head around this so I will open a separate PR for this

rtuck99 (approved):
Approved - I modified the test in order to add coverage for the wait==True case and make `codecov` happy.
Other than that, all fine

PR #1131, opened 2025-06-24
Use latest BlueAPI version number in dockerfile

olliesilvester:
Our Dockerfile for `mx-bluesky-blueapi` had a misleading `BlueAPI` version. 

In our `pyproject.toml`, we had BlueAPI pinned to a version greater than that specified in the Dockerfile. When the image did  `pip install`, it uninstalled the version specified at the top of the Dockerfile and used the pyproject's version instead.

This is confusing when debugging, so we should make sure that the version in the Dockerfile agrees with the pyproject.toml's pinnings

### Checks for reviewer


olliesilvester:
Also done in https://github.com/DiamondLightSource/mx-bluesky/pull/1133

Issue #1118, opened 2025-06-16
Hyperion testing v1.5.0 16/6/25


olliesilvester:
- When creating new venv on deployment, need to do some in RH7 and some in RH - check if this is documented
- Error on starting Hyperion:
```
Traceback (most recent call last):
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/_core/__init__.py", line 22, in <module>
    from . import multiarray
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/_core/multiarray.py", line 11, in <module>
    from . import _multiarray_umath, overrides
ImportError: /lib64/libm.so.6: version `GLIBC_2.27' not found (required by /dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/_core/_multiarray_umath.cpython-311-x86_64-linux-gnu.so)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/__init__.py", line 112, in <module>
    from numpy.__config__ import show_config
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/__config__.py", line 4, in <module>
    from numpy._core._multiarray_umath import (
  File "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/lib/python3.11/site-packages/numpy/_core/__init__.py", line 48, in <module>
    raise ImportError(msg) from exc
ImportError: 

IMPORTANT: PLEASE READ THIS FOR ADVICE ON HOW TO SOLVE THIS ISSUE!

Importing the numpy C-extensions failed. This error can happen for
many reasons, often due to issues with your setup or how NumPy was
installed.

We have compiled some common reasons and troubleshooting tips at:

    https://numpy.org/devdocs/user/troubleshooting-importerror.html

Please note and check the following:

  * The Python version is: Python3.11 from "/dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/bin/python"
[...]

PR #1134, opened 2025-06-25
Beamline fixes from v1.5.0 testing

rtuck99:
Fixes various issues arising from

* #1118
   * Wasn't using GPU results as we need to set `use_gpu` on the Zocalo device
   * Grid snapshots worked but still moved the backlight in
   * We should set `use_grid_snapshots` for multipin by default
   * Snapshot folder needs to be created

* DiamondLightSource/dodal#1311

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/hyperion/external_interaction/agamemnon.py
+++ b/src/mx_bluesky/hyperion/external_interaction/agamemnon.py
@@ -232,6 +232,7 @@ def update_params_from_agamemnon(parameters: T) -> T:
                 # Before we do https://github.com/DiamondLightSource/mx-bluesky/issues/226
                 # this will give no snapshots but that's preferable
                 parameters.multi_rotation_scan.snapshot_omegas_deg = []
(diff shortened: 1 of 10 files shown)


src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -232,6 +232,7 @@ def update_params_from_agamemnon(parameters: T) -> T:
                 # Before we do https://github.com/DiamondLightSource/mx-bluesky/issues/226
                 # this will give no snapshots but that's preferable
                 parameters.multi_rotation_scan.snapshot_omegas_deg = []
+                parameters.multi_rotation_scan.use_grid_snapshots = True
```
DominicOram, line 235: Should: this comment is now out of date, right?

DominicOram (changes requested):
Thanks! One comment, otherwise good

PR #1143, opened 2025-07-01
Fix long exception messages causing ispyb callback handler to throw and not log run status.

rtuck99:
Fixes an issue discovered during system testing for another issue whereby if an exception is raised in the course of a data collection and the exception message is sufficiently long (for example, it references a large composite or lengthy request parameters) then the length may exceed the available comment column length when appended to the existing comments.

This results in the exception not being recorded in ispyb, and also the datacollection status not being updated.

1. Tests pass

### Checks for reviewer


DominicOram (approved):
Great, thank you!

PR #1150, opened 2025-07-07
1009 start hyperion in independent mode

### LINKED ISSUE #1009: Have ability to start Hyperion in an "independent mode"
Following https://github.com/DiamondLightSource/mx-bluesky/issues/358 we need to be able to put start hyperion in a mode that has it just listen for the baton to then do the queue.

This will effectively make hyperion independent from GDA so could be done in a separate way but given that we still need to be running as gda2 for some permissions and we may want to turn this mode on and off it will be easier to reuse the `hyperion_restart` functionality.

## Notes
Prototyped on the beamline by doing:
```python
def new_main():

    from mx_bluesky.hyperion.experiment_plans.load_centre_collect_full_plan import create_devices
    from dodal.devices.baton import Baton
    do_default_logging_setup(
        CONST.LOG_FILE_NAME, CONST.GRAYLOG_PORT
    )
    LOGGER.info("Started hyperion with baton")
    context = setup_context()
    RE= RunEngine()

    RE.subscribe(LogUidTaggingCallback())

    LOGGER.info("Connecting to external callback ZMQ proxy...")
    publisher = Publisher(f"localhost:{CONST.CALLBACK_0MQ_PROXY_PORTS[0]}")
    RE.subscribe(publisher)

    composite = create_devices(context)
    baton = find_device_in_context(context, "baton", Baton)
    LOGGER.info("Starting plan")
    RE(run_udc_when_requested(baton, composite))
    LOGGER.info("Plan done")
```

## Acceptance Criteria
- There is a GDA property that specifies starting Hyperion in an "independent mode"
- When activated and `hyperion_restart` is called then hyperion starts up waiting for the baton

rtuck99:
Fixes #1009

This PR introduces a new command line option `--mode` which can be specified as either `gda` (the default) or `udc`

GDA mode is exposes the existing behaviour
UDC mode causes hyperion to start in a mode that polls Agamemnon for instructions.

* On start Hyperion waits for the baton to be requested for Hyperion
* If the baton is requested for Hyperion, then Hyperion sets the current baton holder to be Hyperion
* Hyperion then fetches instructions from Agamemnon and processes them until no more instructions are available
* If a beamline error is encountered, then processing stops
* While in UDC mode, Hyperion doesn't accept requests via the REST endpoint to run experiment plans, but does respond to status requests

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


tests/unit_tests/hyperion/test_baton_handler.py
```diff
@@ -303,33 +384,226 @@ def test_initialise_udc_reloads_all_devices(
...
+        ],
+    ],
+)
+@pytest.mark.timeout(1)
+async def test_wait_on_queue_resumes_collection_when_normal_completion_and_baton_returned(
+    mock_create_parameters_from_agamemnon: MagicMock,
+    udc_runner: UDCRunner,
+    executor: Executor,
+):
+    async def take_baton_away_then_wait_for_release_then_re_request():
```
DominicOram, line 523: Should: This doesn't take the baton away, it waits on the queue to finish?

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -87,45 +139,85 @@ def initialise_udc(context: BlueskyContext, dev_mode: bool = False):
...
+                    )
+                case Wait():
+                    runner.start(_runner_sleep, parameters)
+                case _:
+                    raise AssertionError(
+                        f"Unsupported instruction decoded from agamemnon {type(parameters)}"
+                    )
+            yield from runner.execute_next_command()
+    else:
+        yield from _safely_release_baton(baton)
```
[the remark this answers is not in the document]
DominicOram: Actually, this even more complicated than I thought, we change `requested_user` here but then the whole thing is wrapped in a `release_baton`, which then changes `current_user`. I guess this works but why are we doing it twice?

DominicOram (changes requested):
Sorry, only half a review really, it took me a while to get my head around the logic and I ran out of time. The tests are timing out in CI, and on my machine, I think it would be good to replace them with tests that use `asyncio.Event`s to control flow rather than relying on the timing of the `Wait` command

DominicOram:
>Happy to discuss again next week. Apologies that we have already had the discussion but hopefully now I've written this I can articulate my point better.

Might be good to bring @olliesilvester in too as a tie-break opinion

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -87,45 +139,85 @@ def initialise_udc(context: BlueskyContext, dev_mode: bool = False):
...
+
+
+def _hyperion_loop_exception_handler(runner: UDCRunner, exception: Exception):
+    if runner.RE.state == "aborting":
+        baton = find_device_in_context(runner.context, "baton", Baton)
+        yield from _safely_release_baton(baton)
+        if command := runner.try_fetch_next_command():
+            if command.action == Actions.SHUTDOWN:
+                LOGGER.info("Shut down command received, shutting down Hyperion")
+                return
```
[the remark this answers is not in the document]
rtuck99: I remembered why! RunEngine.stop() raises `RequestAbort` which is caught inside the plan. However, regardless of what happens inside, whatever was outside the plan that called RE(my_plan) will get a `RunEngineInterrupted` exception, and this is why we don't need to check a flag in the UDCRunner implementation

src/mx_bluesky/hyperion/baton_handler.py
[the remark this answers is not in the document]
rtuck99: This safely_release_baton terminates the outer loop, which is predicated on the baton being requested for Hyperion. That check is so that if a user from outside wants to interrupt Hyperion in an orderly fashion, they can do so by requesting the baton.

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -135,33 +131,29 @@ def _fetch_and_process_agamemnon_instruction(
...
     if parameter_list:
         for parameters in parameter_list:
+            LOGGER.info(
+                f"Executing plan with parameters: {parameters.model_dump_json(indent=2)}"
+            )
             match parameters:
                 case LoadCentreCollect():
+                    devices: Any = PLAN_REGISTRY["load_centre_collect_full"]["setup"](
+                        runner.context
+                    )
```
DominicOram, line 141: Could: I still don't understand why we're using the `PLAN_REGISTRY` at all, why not just call `load_centre_collect_full_plan.create_devices`?

tests/unit_tests/conftest.py
```diff
@@ -75,6 +78,75 @@ async def RE():
...
+
+        loop.set_debug(True)
+
+        request.addfinalizer(
+            lambda: _error_and_kill_pending_tasks(
+                loop, request.node.name, request.session.testsfailed == fail_count
+            )
+        )
+    # Once https://github.com/bluesky/ophyd-async/issues/683
+    # is finished we can remove this try, except.
```
DominicOram, line 146: Should: It's not clear what you mean by finished? The issue is closed, the associated PR is merged, it looks finished to me.

tests/unit_tests/conftest.py
```diff
@@ -75,6 +78,75 @@ async def RE():
...
+        raise RuntimeError(
+            f"Not all tasks closed during test {test_name}:\n"
+            f"{pprint.pformat(unfinished_tasks, width=88)}"
+        )
+
+    return unfinished_tasks
+
+
+@pytest.fixture(autouse=True, scope="function")
+async def fail_test_on_unclosed_tasks(request: FixtureRequest):
```
DominicOram, line 128: The `dodal` approach is that we log tasks and wait on them to finish (https://github.com/DiamondLightSource/dodal/blob/2a35a94c15b7efbc02b6f30cad5e8f0455f53308/tests/common/beamlines/test_beamline_utils.py#L28). I guess it's fine to have different approaches in each module but seems potentially confusing?

tests/unit_tests/t01.py
```diff
@@ -0,0 +1,14 @@
+from dodal.common.beamlines.beamline_utils import device_factory
+from dodal.devices.baton import Baton
+from dodal.utils import BeamlinePrefix, get_beamline_name
+
```
DominicOram, line 4: Nit: Could be worth a comment here too as to why we need it.

tests/unit_tests/hyperion/test_baton_handler.py
```diff
@@ -25,58 +30,137 @@
...
+# PYTEST_TEST_TIMEOUT_S > TEST_SCRIPT_TIMEOUT in order that an exception on the script
+# is bubbled up via the future and not lost.
+PYTEST_TEST_TIMEOUT_S = 10
+
+AGAMEMNON_WAIT_INSTRUCTION = Wait.model_validate(
+    {
+        "duration_s": AGAMEMNON_WAIT_FOR_TEST_STEP_S,
+        "parameter_model_version": PARAMETER_VERSION,
+    }
+)
```
DominicOram, line 64: As I said in https://github.com/DiamondLightSource/mx-bluesky/pull/1150#pullrequestreview-3054717581, I still think we can improve some of the flakiness by using Events etc. but I've spun into https://github.com/DiamondLightSource/mx-bluesky/issues/1185

DominicOram (approved):
Great, thank you, some minor comments, take them or leave them.

PR #1153, opened 2025-07-07
Implement ConfigServer for MX and Hyperion

### LINKED ISSUE #1097: Migrate to new daq-config-server

olliesilvester:
This PR also changes how Hyperion toggles its features. Previously, GDA would pass these feature flags through the 'features' dictionary in the parameter model. This dictionary is now removed, and the source of truth becomes the values within `domain.properties`. If this file cannot be read, Hyperion will use its defaults in `FeatureFlags` instead.

Will require changes to GDA for constructing Hyperion params

Fixes #1097

1. Check if it's okay to have `omega_flip` as a Hyperion constant, or should we also add it as a feature flag within domain.properties? Same question for `ALTERNATE_ROTATION_DIRECTION`

### Checks for reviewer


src/mx_bluesky/hyperion/parameters/gridscan.py
```diff
@@ -9,22 +9,24 @@
...
     @property
     def detector_params(self):
         params = super().detector_params
-        params.enable_dev_shm = self.features.use_gpu_results
+        params.enable_dev_shm = self.config_server.get_feature_flags().USE_GPU_RESULTS
         return params
 
 
-class HyperionSpecifiedThreeDGridScan(WithHyperionUDCFeatures, SpecifiedThreeDGridScan):
+class HyperionSpecifiedThreeDGridScan(  # type: ignore
```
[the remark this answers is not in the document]
olliesilvester: I think this was because MyPy was checking that all the baseclasses had the same signature, and since `WithHyperionUDCFeatures` wasn't a BaseClass, it was complaining. Fixed by making it a BaseModel too, even though it looks weird

src/mx_bluesky/hyperion/parameters/constants.py
```diff
@@ -24,13 +26,26 @@ class I03Constants:
...
+    PANDA_RUNUP_DISTANCE_MM = "gda.mx.hyperion.panda_runup_distance_mm"
+
+
+# Use these defaults if we can't read from the config server
+@dataclass
+class HyperionFeatureFlags(FeatureFlags):
+    USE_GPU_RESULTS: bool = True
+    USE_PANDA_FOR_GRIDSCAN: bool = False
+    SET_STUB_OFFSETS: bool = False
+    PANDA_RUNUP_DISTANCE_MM: float = 0.16
```
olliesilvester, line 48: Maybe flags is a bad name given that they don't have to be boolean?

src/mx_bluesky/hyperion/external_interaction/callbacks/rotation/nexus_callback.py
```diff
@@ -103,6 +103,6 @@ def activity_gated_start(self, doc: RunStart):
                 full_num_of_images=self.full_num_of_images,
                 meta_data_run_number=self.meta_data_run_number,
                 axis_direction=AxisDirection.NEGATIVE
-                if parameters.features.omega_flip
+                if I03Constants.OMEGA_FLIP
```
rtuck99, line 106: ditto

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -1,58 +1,140 @@
...
+        super().__init__(url)
+
+    def _verify_feature_parameters(self):
+        sources_keys = {feature.name for feature in self.feature_sources}
+        feature_dc_keys = {key.name for key in fields(self.feature_dc)}
+        assert sources_keys == feature_dc_keys, (
+            f"MXConfig server feature_sources names do not match feature_dc keys: {sources_keys} != {feature_dc_keys}"
+        )
+
+    def get_oav_config(self, reset_cached_result=False) -> dict[str, Any]:
```
rtuck99, line 57: There isn't anything that currently uses this - I assume that's going to be in another PR? How do you plan to map this to what the device currently wants? The OAV configuration is currently split across 3 different files, there's the zoom parameters, which is XML, the OAV centring config file, which is json, and there's the display configuration, which is `<name> = <value>`

rtuck99: 2. Do we even want the caller to know about what file format the data is in? Ideally that should be an implementation detail that we aim to get rid of later. So perhaps it should just be `get_config(self, resource_name) -> dict[str, Any]`. The mapping to the file type perhaps should be internal, we just want something that for a given named resource, returns a nested dict of structured data.

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -1,58 +1,140 @@
...
+                or time() - self._time_of_last_feature_get > FEATURE_FLAG_CACHE_LENGTH_S
+            ):
+                self._cached_features = None
+                # Construct self.feature_dc by reading the domain.properties file
+                enum_dict = (
+                    self.feature_sources._value2member_map_
+                )  # As of python 3.12, can do checks using "in" for enums instead
+                feature_dict = {}
+                domain_properties = self.get_file_contents(
+                    GDA_DOMAIN_PROPERTIES_PATH, reset_cached_result=reset_cached_result
```
rtuck99, line 113: I have concerns about using existing filename constants (this, and OAV_CONFIG_JSON) in this class. Long term, we want to get away from the linkage between file paths and the configuration resource so that we are independent of where the files are located and how they are organised.

These constants should be thought of more like resource keys. So I think these should be replaced with a set of well known resource name constants which are just simple strings. 

Internally we can map them for now to file locations since that's what the config server currently uses, but later on we can remove that.

When you look at other configuration APIs, you tend to find that they follow some kind of similar pattern - you have a resource name, which then gets you either a flat dict or a nested dict. Often the dict is comprised by a series of overlays of multiple dicts, for example to allow user settings to override system settings to override site-wide settings. The consumer of the API doesn't care about how the settings were resolved.

src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
```diff
@@ -133,7 +133,7 @@ def calculate_motion_profile(
     direction = params.rotation_direction
     start_scan_deg = params.omega_start_deg
 
-    if params.features.omega_flip:
+    if I03Constants.OMEGA_FLIP:
```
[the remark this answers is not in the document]
olliesilvester: True, but everything in the `Hyperion` folder is tied to i03 right now

src/mx_bluesky/hyperion/parameters/components.py
```diff
@@ -1,14 +1,20 @@
...
+from mx_bluesky.hyperion.parameters.constants import HyperionFeatureFlags
 
 
-class WithHyperionUDCFeatures(WithPandaGridScan):
-    features: HyperionFeatureFlags = Field(default=HyperionFeatureFlags())
+class WithHyperionConfigServer(BaseModel):
+    # MyPy checks for base classes signature compatibility, so make this a BaseModel to stop mypy errors
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+    config_server: MXConfigServer[HyperionFeatureFlags] = Field(
```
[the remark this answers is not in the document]
olliesilvester: Need some kind of way to choose between different config server clients. Right now there's only the Hyperion one, but there will be others. Any suggestions if not the parameter model?

docs/developer/hyperion/reference/coordinate-systems.rst
```diff
@@ -20,7 +20,7 @@ defined by ``rotation_direction`` which can be ``POSITIVE`` or ``NEGATIVE``
 
 The coordinate systems of the underlying hardware may differ from the request. At Diamond traditionally the omega 
 axis is implemented to be in the opposite direction, therefore the commands sent to the hardware are transformed and 
-the ``omega_flip`` feature flag controls this behaviour, which by default is set to ``True`` as this is 
+the ``omega_flip`` flag controls this behaviour, which by default is set to ``True`` as this is
```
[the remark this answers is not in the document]
olliesilvester: We discussed on slack last week - this option was hardcoded in GDA prior to this PR, I've just moved the hardcoding to `mx-bluesky`. I can adjust the docs to say that the 'flag' is just a constant in the code though

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -1,58 +1,140 @@
...
+    OavConstants,
+)
+from mx_bluesky.common.utils.log import LOGGER
+
+FEATURE_FLAG_CACHE_LENGTH_S = 60 * 5
+
+T = TypeVar("T", bound=FeatureFlags)
+
+
+class MXConfigServer(ConfigServer, Generic[T]):
```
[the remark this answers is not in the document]
olliesilvester: I tried calling it `MXConfigClient` but changed my mind. Maybe that is better though

src/mx_bluesky/hyperion/experiment_plans/rotation_scan_plan.py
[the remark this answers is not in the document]
olliesilvester: Could make a new ticket to move omega flip to domain.properties?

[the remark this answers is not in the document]
olliesilvester: And `ALTERNATE_ROTATION_DIRECTION`

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -54,31 +55,44 @@ def _verify_feature_parameters(self):
...
         if reset_cached_result:
-            self._cached_oav_config = None
-        if not self._cached_oav_config:
-            config_path = OavConstants.OAV_CONFIG_JSON
+            self._cached_json_config = {}
+        str_to_json = str(path_to_json)
+        if str(str_to_json) not in self._cached_json_config:
             try:
-                self._cached_oav_config = TypeAdapter(dict[str, Any]).validate_python(
+                self._cached_json_config[str(str_to_json)] = TypeAdapter(
```
rtuck99, line 70: `str_to_json` is already a str?

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -54,31 +55,44 @@ def _verify_feature_parameters(self):
...
+                ).validate_python(
                     self.get_file_contents(
-                        config_path, dict, reset_cached_result=reset_cached_result
+                        path_to_json, dict, reset_cached_result=reset_cached_result
                     )
                 )
             except Exception as e:
                 LOGGER.warning(
-                    f"Failed to get oav config from config server: {e} \nReading the file directory..."
+                    f"Failed to get json config from config server: {e} \nReading the file directory..."
```
rtuck99, line 79: Do we want the behaviour of the config reader to be the same if there is a syntax error in the json file vs if the file is not there?

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -42,7 +43,7 @@ def __init__(
         self.feature_sources = feature_sources
         self.feature_dc: type[T] = feature_dc
         self._cached_features: T | None = None
-        self._cached_oav_config: dict[str, Any] | None = None
+        self._cached_json_config: dict[str, dict[str, Any]] = {}
```
rtuck99, line 46: `MXConfigServer` has `_cached_json_config` but parent class `ConfigServer` also has `_cache`. This is too much caching!

src/mx_bluesky/common/external_interaction/config_server.py
```diff
@@ -54,31 +55,44 @@ def _verify_feature_parameters(self):
...
-        if not self._cached_oav_config:
-            config_path = OavConstants.OAV_CONFIG_JSON
+            self._cached_json_config = {}
+        str_to_json = str(path_to_json)
+        if str(str_to_json) not in self._cached_json_config:
             try:
-                self._cached_oav_config = TypeAdapter(dict[str, Any]).validate_python(
+                self._cached_json_config[str(str_to_json)] = TypeAdapter(
+                    dict[str, Any]
+                ).validate_python(
```
rtuck99, line 72: Why do we call `validate_python()` here? Isn't it the job of `get_file_contents()` to return the correct type?

src/mx_bluesky/common/external_interaction/config_server.py
olliesilvester: Ah I misread your original comment. I think it's safest to try and read the file manually if there were any exceptions from the config server. We will be able to see what the real problem is regardless

rtuck99 (approved):
Approved

PR #1162, opened 2025-07-15
Fix system tests trying to access /dls/tmp on test collection

rtuck99:
When local system tests are run on a non-diamond machine test collection fails because import of  `test_agamemnon.py` executes a pydantic model validation that tries to access /dls/tmp folder. Even though the tests in this file are not part of the system tests since they aren't decorated as such.

I believe this does not cause issues with the nightly build because inside the system test execution container, everything executes as root so creation of this folder succeeds. However running locally outside of a container, runs into issues.

1. tox -e `localsystemtests` passes

### Checks for reviewer

### DIFF
--- a/tests/system_tests/hyperion/external_interaction/test_agamemnon.py
+++ b/tests/system_tests/hyperion/external_interaction/test_agamemnon.py
@@ -53,27 +53,23 @@ EXPECTED_ROTATION_PARAMS = {
     ],
 }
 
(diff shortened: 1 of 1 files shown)


DominicOram (commented):
The PR description doesn't seem to match the change, or am I missing something?

DominicOram (approved):
Sorry, you're right, this is obvious. Must have been tired when I reviewed it before

Issue #1180, opened 2025-07-29
Do alerts for things other than errors


DominicOram:
As beamline staff I would like to be alerted of the following:
* UDC start
* Baton request - The scientist is happy with this being on the release of the baton by hyperion
* UDC finish  - This is covered by GDA still, see https://github.com/DiamondLightSource/mx-bluesky/issues/1102#issuecomment-3163746956
* ~Visit change~
* Puck change - every new visit is a new puck anyway so we can just alert on this rather than visit

DominicOram:
Worth baring in mind that we probably don't want emails for puck changes when running outside UDC

DominicOram:
I think all of this is now done

PR #1173, opened 2025-07-28
Provide detailed metadata when alerting for beamline errors

rtuck99:
Fixes #1071 

Link to dodal PR (if required): N/A

(remember to update `pyproject.toml` with the dodal commit tag if you need it for tests to pass!)

Graylog events and notification configuration for the Hyperion stream have been updated with additional event fields.

Beamline errors should now generate alerts with
* Sample ID
* Visit
* Proposal
* Container ID
* ISPyB link (to the corresponding sample)
* Link to graylog 

### Instructions to reviewer on how to test:

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

PR #1181, opened 2025-07-29
Alert when the container changes

DominicOram:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/1180

Based on https://github.com/DiamondLightSource/mx-bluesky/pull/1173, review/merge that first.

Additionally renamed these callbacks (including the sample handling one) to UDC callbacks as I can't see why we would want them running outside UDC

1. Confirm tests pass

### Checks for reviewer


docs/user/hyperion/alerts.rst
```diff
@@ -1,7 +1,7 @@
 Alerts
 ------
 
-Hyperion can be configured to raise alerts in order to notify staff of issues with data collections or with Hyperion 
+Hyperion can be configured to raise  in order to notify staff of issues with data collections or with Hyperion
```
rtuck99, line 4: raise... ? something, if not alerts then what?

src/mx_bluesky/common/external_interaction/callbacks/udc/alert_on_container_change.py
```diff
@@ -0,0 +1,40 @@
...
+    get_alerting_service,
+)
+from mx_bluesky.common.external_interaction.callbacks.common.plan_reactive_callback import (
+    PlanReactiveCallback,
+)
+from mx_bluesky.common.utils.log import ISPYB_ZOCALO_CALLBACK_LOGGER
+
+
+class AlertOnContainerChange(PlanReactiveCallback):
+    """Sends an alert to beamline staff when a pin from a new puck has been loaded.
```
rtuck99, line 15: The fact that our callbacks are not stateless makes me rethink my previous discussion on callbacks re. handling failure of the callback process.

This would mean that the document store / messages in the rabbitmq message queue couldn't be guaranteed to be a complete description of everything needed to (re)generate output artefacts. In the event that the callback process fails, playing back the events won't be sufficient unless you ensure that no prior context outside the replayed messages is required, and also that the messages are played back in an unbroken ordered sequence.

It does mean that we would have to rethink some aspects of our communication with our callbacks though, because for example in the ispyb callbacks we would need to have some way of returning the IDs back to the experiment plan, so that subsequent events include sufficient context.

In this case here, it is not so important since it's unlikely that anyone would care too much about replay of lost alert messages.

src/mx_bluesky/common/external_interaction/callbacks/udc/alert_on_container_change.py
```diff
@@ -0,0 +1,40 @@
...
+            visit = metadata.get("visit")
+
+            if new_container != self._last_container:
+                beamline = get_beamline_name("")
+                get_alerting_service().raise_alert(
+                    f"UDC moved on to puck {new_container} on {beamline}",
+                    f"Hyperion finished container {self._last_container} and moved on to {new_container}",
+                    {
+                        Metadata.SAMPLE_ID: str(sample_id),
+                        Metadata.VISIT: visit or "",
```
rtuck99, line 35: Are there occasions where we don't have a visit?

rtuck99 (approved):
Approved

src/mx_bluesky/common/experiment_plans/inner_plans/write_sample_status.py
```diff
@@ -3,7 +3,7 @@
 import bluesky.plan_stubs as bps
 import bluesky.preprocessors as bpp
 
-from mx_bluesky.common.external_interaction.callbacks.sample_handling.sample_handling_callback import (
+from mx_bluesky.common.external_interaction.callbacks.udc.sample_handling_callback import (
```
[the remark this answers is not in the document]
DominicOram: I think for the sample handling stuff this is complicated by wanting to use them for https://jira.diamond.ac.uk/browse/MXGDA-4148. I agree with you about the alerting though, I will move that one

src/mx_bluesky/common/external_interaction/callbacks/udc/alert_on_container_change.py
DominicOram: There shouldn't be but I don't think we need to assume it

rtuck99 (approved):
Approved with minor comments

PR #1186, opened 2025-08-04
Remove unused function for i04 grid scans

DominicOram:
This function is used in other Hyperion plans to instantiate devices but in this case device instantiation is done for us by BlueAPI so we do not need it.

1. Confirm tests pass

### Checks for reviewer


rtuck99 (approved):
Approved

PR #1165, opened 2025-07-21
Add plan to enter UDC default state

DominicOram:
Fixes #396 

Requires https://github.com/DiamondLightSource/dodal/pull/1364

### Instructions to reviewer on how to test:

1. Confirm plan matches the description on the issue
2. Confirm tests pass

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

PR #1193, opened 2025-08-06
188 hyperion restart for UDC mode

### LINKED ISSUE #188: Mechanism for hyperion_restart() when hyperion is running in a container
When hyperion is running in a container, the existing `run_hyperion.sh` will no longer be able to restart hyperion because it will be invoked from outside the container.

As an interim workaround, this can be achieved by scaling down the pods to 0 and then back up to 1 in the kubernetes dashboard.

Longer term we will need a solution, it could be implemented as one of the following:

* Implement the currently unimplemented `Actions.SHUTDOWN` command to shutdown the server - we already implement `STOP` but this only stops the current plan. Once the existing service instance exits, k8s should then restart it naturally.
* `hyperion_restart()` could invoke a script that manages the kubernetes deployment directly, however it would need to be running with the appropriate privileges so this would be dependent on how the account is set up.

I think option 1 is probably simplest

## Acceptance Criteria

* Non-containerised deployments of hyperion are still able to restart without requiring any changes
* `hyperion_restart()` on a containerised deployment restarts hyperion

rtuck99:
Fixes #188

Also requires corresponding GDA changes

This introduces one new REST endpoint to Hyperion when running in UDC mode:

* GET `/status` - returns `application/json` 
```
{
 "status": <status>
}
```
Where <status> is one of `WARN|FAILED|SUCCESS|BUSY|ABORTING|IDLE`

In addition, in UDC mode we install a `SIGTERM` signal handler that calls the `shutdown()` method of the plan runner. This sends an abort() to the `RunEngine` instance such that the app will exit.

See also https://gitlab.diamond.ac.uk/MX-GDA/hyperion-system-testing/-/merge_requests/4
which provides a dummy agamemnon server for one of the system tests

Some relevant documentation changes were also made as part of
* #1192

It was decided not to implement `hyperion_restart`, since this reduces the attack surface of Hyperion, leaving no write methods, which eliminates the need for access control (effectively deferring #731 until such time as we migrate to blueapi and implement supervisor process) since restart would then be limited to k8s users.

1. Run the system test infrastructure with `podman compose -f system-test-compose.yml up -d`
2. `test_main_udc.test_udc_mode` should run shutdown cleanly with SIGTERM

### Checks for reviewer


DominicOram (changes requested):
Couple of comments but otherwise good, thanks. Some tests are failing (including the system test) now that https://github.com/DiamondLightSource/mx-bluesky/pull/1165 has been merged.

DominicOram (approved):
Thanks. Code looks good but the original issue talks about replacing `hyperion_restart`, which this doesn't do. I think I remember discussion about how we effectively will just do it through k8s, which is fine but can you document the decision here and reasoning (I assume your memory is better than mine on it?)

PR #1196, opened 2025-08-07
Ignore omega order in grid detect callback

### LINKED ISSUE #1194: GridDetectionCallback should take into account different rotation orders
Following https://github.com/DiamondLightSource/mx-bluesky/pull/1179 we now do grid detection in a different omega order. However, `GridDetectionCallback` makes assumption about the order that the are coming in with and so this does not work.

## Acceptance Criteria
- `GridDetectionCallback` gives correct grid scan co-ordinates regardless of order of events

DominicOram:
Fixes #1194

1. Confirm new test passes and tests the issue
 
### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/external_interaction/callbacks/common/grid_detection_callback.py
+++ b/src/mx_bluesky/common/external_interaction/callbacks/common/grid_detection_callback.py
@@ -1,4 +1,4 @@
-from typing import TypedDict
+from typing import Generic, TypedDict, TypeVar
 
 import numpy as np
 from bluesky.callbacks import CallbackBase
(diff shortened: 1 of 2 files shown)


rtuck99 (approved):
Approved.

I've updated the system tests, which should now be fixed. :crossed_fingers:   However I have been running into issues when running them in gitlab CI due to a setuptools version warning. This seems to have crept in due to updated dependency of epicscorelibs.

I've added an ignore, I think probably an issue should be raised against an upstream project, however I'm not sure which, since the offending commit in epicscorelibs has been there for about 4 years, this is only surfaced due to later version of pvxslibs having very stringent version requirements, so any epicscorelibs change woud probably require an update to that lib as well.

Issue #1200, opened 2025-08-08
Add basic plans + docs for Jungfrau commissioning


olliesilvester:
Using the [ophyd-async jungfrau device](https://github.com/bluesky/ophyd-async/pull/999), we should start adding some plans from the old commissioning [branch](https://github.com/DiamondLightSource/mx-bluesky/compare/main...updated-jungfrau-branc)

Some of the most useful ones are:
- do_darks
- do_pedestal_darks
- do_manual_acquisition

Future plans that also may be needed in future PR but need discussion with scientists:
- rotation_scan_plan
- do_burst_mode

This will also require adding some devices to i24 on dodal

## Acceptance Criteria
- Useful plans added
- Add a "setup ipython" script so that scientists can easily trigger plans
- ~Add scientist-facing docs for using mx-bluesky to trigger jungfrau data collections~ moved to https://github.com/DiamondLightSource/mx-bluesky/issues/1325

PR #1202, opened 2025-08-13
Add external triggering jungfrau plan

olliesilvester:
CI will fail as the jungfrau isn't merged into ophyd-async yet.

Begins #1200

Requires:
- https://github.com/bluesky/ophyd-async/pull/999
- https://github.com/DiamondLightSource/dodal/pull/1451

1. Use the branches in the above PR's for ophyd-async and dodal
2. Check new plan
3. Check new tests

### Checks for reviewer


tests/unit_tests/beamlines/i24/jungfrau_commissioning/test_do_manual_acquisition.py
```diff
@@ -0,0 +1,80 @@
...
+from mx_bluesky.beamlines.i24.jungfrau_commissioning.do_external_acquisition import (
+    do_external_acquisition,
+)
+
+
+async def _do_sleep():
+    await asyncio.sleep(0)
+
+
+def test_full_do_external_acquisition(jungfrau: Jungfrau, RE: RunEngine, caplog):
```
DominicOram, line 21: Should: Why is this file name referencing manual acquisition when you're doing an external triggered one?

tests/unit_tests/beamlines/i24/jungfrau_commissioning/test_do_manual_acquisition.py
```diff
@@ -0,0 +1,80 @@
...
+    def test_plan():
+        status = yield from do_external_acquisition(0.001, 0.002, 5, 5, jungfrau)
+        assert not status.done
+        val = 0
+        while not status.done:
+            val += 1
+            set_mock_value(jungfrau._writer._drv.num_captured, val)
+
+            # Let status update
+            yield from bps.wait_for([_do_sleep])
```
DominicOram, line 32: Nit: could use a partial here
```suggestion
            yield from bps.wait_for([partial(asyncio.sleep,0)])
```

src/mx_bluesky/beamlines/i24/jungfrau_commissioning/do_external_acquisition.py
```diff
@@ -0,0 +1,71 @@
...
+)
+from ophyd_async.fastcs.jungfrau import (
+    Jungfrau,
+    create_jungfrau_external_triggering_info,
+)
+
+from mx_bluesky.common.utils.log import LOGGER
+
+
+def do_external_acquisition(
```
DominicOram, line 20: Should: I'm not sure how useful this plan is in isolation because to get the external triggering working we will need to set the zebra up etc. The issue also says that the first one to start with is `do_manual_acquisition`, which is an internal acquisition, not an external one, like the rotation

DominicOram (changes requested):
Thanks, couple of comments in code. Do we also need to add the jungfrau to `dodal.beamlines.i24`?

olliesilvester:
> Do we also need to add the jungfrau to `dodal.beamlines.i24`?

I just added a draft PR [here](https://github.com/DiamondLightSource/dodal/pull/1473). Not sure the prefix is / the number of odin node / the default path + filename to use yet though

DominicOram (approved):
Looks good, thanks. I think we need to get the ophyd-async one merged and released first then we can do this one

olliesilvester:
Once this CI passes I will merge this PR. The plan here is outdated compared to the one tested on i24, but that is addressed in https://github.com/DiamondLightSource/mx-bluesky/pull/1216/files#diff-6c9e203e0e62b4dd5bf23a5423b224801bcbede08dc0865d9ec76efcfc08ac3f

PR #1210, opened 2025-08-20
Make new parameter component for required sample id

### LINKED ISSUE #1209: Sample ID shouldn't be required unless doing robot load
In our parameter model, sample id is currently a required integer. We only currently use the sample id for:
- Robot load - we set a PV in the robot to the sample id value
- Ispyb deposits for data collection + gridscans, but ispyb doesn't require this.

We should allow for sample id to be None for plans which don't use robot load. I04 have requested this since GDA sometimes doesn't know the sample_id, but they don't want this to prevent grid scans from running

## Acceptance Criteria
- Sample id not required except for robot load plan

olliesilvester:
Fixes #1209

1. Confirm nothing is broken by this and new component is correct solution

### Checks for reviewer


DominicOram (approved):
Thanks, I think this will work for now but we're going to hit the issue again when we need to load the warm pins as the robot load code still requires a sample id

PR #1216, opened 2025-08-29
Add jungfrau internal acquisition plan

olliesilvester:
Part of  #1200 

Requires https://github.com/DiamondLightSource/dodal/pull/1548

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/beamlines/i24/jungfrau_commissioning/do_internal_acquisition.py
```diff
@@ -0,0 +1,46 @@
...
+        total_frames: Number of frames taken after being internally triggered.
+        period_between_frames_s: Time between each detector frame, including deadtime. Not needed if frames_per_triggers is 1.
+        jungfrau: Jungfrau device
+        path_of_output_file: Absolute path of the detector file output, including file name. If None, then use the PathProvider
+            set during jungfrau device instantiation
+        wait: Optionally block until data collection is complete.
+    """
+
+    if path_of_output_file:
+        override_file_name(jungfrau, path_of_output_file)
```
DominicOram, line 42: Should: The name and docs of `path_of_output_file` implies that it's a full path but `override_file_name` actually juts wants a file name. Either update `override_file_name` to take a full path or change `do_internal_acquisition` to only take a new path filename

olliesilvester:
Tests are passing locally, but will wait for CI to get fixed before re-review

src/mx_bluesky/beamlines/i24/jungfrau_commissioning/do_internal_acquisition.py
DominicOram: So the `AutoIncrementingPathProvider` gives us a slightly different format of `/dls/i24/data/2025/cm40647-4/jungfrau/2025-09-15/zeo/00000/data_file` then `/dls/i24/data/2025/cm40647-4/jungfrau/2025-09-15/zeo/00001/data_file`. I think this is fine but we should make the beamline aware

Issue #1149, opened 2025-07-04
MSP: Collect data starting from the tip


DominicOram:
When we collected using different sample IDs on a MSP (see collection at 2025-07-04 14:21:37 on https://ispyb.diamond.ac.uk/dc/visit/lb18145-275) we did so starting from the base of the pin. It's more intuitive to so from the tip of the pin first and doesn't really affect the speed of collection.

## Note
We already seem to sort based on count but this is per sample, we need to sort the whole set of results on their `x` position

## Acceptance Criteria
- MSP collections are done from the tip first

PR #1222, opened 2025-09-02
1149 collect samples in order starting from the tip of the pin

rtuck99:
Fixes

* #1149

When multiple centres are selected, collections should now be collected in order of ascending x-coordinate

1. MSPs are collected in order starting from the end of the tip. Sample IDs are still assigned correctly.
2. Tests pass

### Checks for reviewer


DominicOram (approved):
Great, looks much cleaner too. Thank you!

PR #1229, opened 2025-09-03
986 Hyperion support documentation for beamline staff

rtuck99:
Fixes 

* #986

1. Beamline staff to review the documentation and decide if they are happy

### Checks for reviewer


docs/user/hyperion/index.rst
```diff
@@ -18,6 +18,23 @@ The software supports two modes of operation:
 
 The mode of operation is determined by configuration using the ``gda.mx.udc.hyperion.enable`` parameter
 
+Once Hyperion has received a request, either from GDA or directly from Agamemnon, it will do the following tasks:
+
+- Robot Load (if necessary)
+- Pin tip detection using the OAV
```
DominicOram, line 24: Nit: 
```suggestion
- Pin tip centring using the OAV
```

docs/user/hyperion/troubleshooting.rst
```diff
@@ -0,0 +1,95 @@
...
+~~~~~~~
+
+Graylog is the recommended way to view logs. It is used for more centralised logging which is also more easily 
+searched and archived. Log messages are sent to the `Hyperion graylog stream <https://graylog.diamond.ac.uk/streams/66264f5519ccca6d1c9e4e03/search>`_.
+
+
+Startup Log
+~~~~~~~~~~~
+
+When ``hyperion_restart()`` is called by GDA, it will log the initial console output to a log file. This log file
```
DominicOram, line 23: Could: Cross link to https://diamondlightsource.github.io/mx-bluesky/main/user/hyperion/running.html from here?

docs/user/hyperion/troubleshooting.rst
```diff
@@ -0,0 +1,95 @@
...
+    1181422 ?        Sl     1:54 /dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/bin/python /dls_sw/i03/software/bluesky/mx-bluesky_v1.5.0/mx-bluesky/.venv/bin/hyperion-callbacks
+
+Kubernetes Install
+------------------
+
+If Hyperion is deployed on Kubernetes then it can be managed from the beamline kubernetes dashboard, e.g. 
+https://k8s-i03-dashboard.diamond.ac.uk
+
+In the beamline namespace there will be a deployment ``hyperion-deployment``, a ``hyperion-svc`` service and associated 
+pods, ingress etc. through which the current state may be observed / managed.
```
DominicOram, line 95: Should: I think most users will not know/care if Hyperion is running in Kubernetes. So I would make it very clear here that this is not the regular mode of operation and then move any mention of kubernetes above in terms of logs etc. into this section.

docs/user/hyperion/configuration.rst
```diff
@@ -23,3 +26,27 @@ Configuration of several properties that control Hyperion execution are availabl
     "gda.mx.hyperion.xrc.use_gpu_results", "boolean", "If true, then zocalo gridscan processing uses the GPU results"
     "gda.mx.hyperion.xrc.use_roi_mode", "boolean", "If true then ROI mode is used."
     "gda.mx.udc.hyperion.enable", "boolean",  "Enables Hyperion UDC mode."
```
DominicOram, line 28: Nit: What is `Hyperion UDC mode` to a user?

DominicOram (changes requested):
Thanks, some comments. I wonder if another useful section would be "commonly seen issues and how to fix them". Where we could have:
* A copy of the odin error we see on file system problems
* What you see in GDA if Hyperion isn't running

@npat113 - you also see some smargon issues quite regularly right? Got a link to an instance and a recommendation on what to do?

docs/user/hyperion/configuration.rst
rtuck99: Yes, I was a bit uncertain how much to mention "UDC mode" and also k8s as a) they are somewhat in flux at the moment and all this complication will ultimately go away and also b) as you say kubernetes isn't terribly visible. Apart from restarting it, I don't know what else there is likely to be, and if the user doesn't have cluster namespace admin rights, then there isn't anything they can really do anyway.

PR #1230, opened 2025-09-03
Use murko results during thawing

### LINKED ISSUE #136: I04: Get centring info from murko after thawing
TBD following https://github.com/DiamondLightSource/mx_bluesky/issues/135

DominicOram:
Fixes #136

This actually only moves to the centre after the first rotation, we should do it again after the second rotation but spinning that into a new issue.

1. Confirm tests pass and docstrings make sense

### Checks for reviewer


rtuck99 (approved):
Nice. Approved.

My only somewhat minor comment is it would be nice to have the public plans at the top and the implementations underneath in `thawing_plan.py`

PR #1234, opened 2025-09-03
Add Jungfrau rotation scan plan

olliesilvester:
Part of #1200 
Prerequisites for CI to pass: https://github.com/DiamondLightSource/dodal/pull/1721

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/beamlines/i24/jungfrau_commissioning/plan_stubs/test_plan_utils.py
+++ b/tests/unit_tests/beamlines/i24/jungfrau_commissioning/plan_stubs/test_plan_utils.py
@@ -65,7 +65,7 @@ async def test_fly_jungfrau_does_read_plan_after_prepare(
     read_hardware = MagicMock()
 
     filename = "test"
-    jungfrau.prepare = MagicMock(side_effect=lambda: completed_status())
+    jungfrau.prepare = MagicMock(side_effect=lambda _: completed_status())
 
     parent_mock = MagicMock()
     parent_mock.attach_mock(jungfrau.prepare, "jungfrau_prepare")


src/mx_bluesky/beamlines/i24/parameters/rotation.py
```diff
@@ -0,0 +1,20 @@
...
+from sqlalchemy import Float
+
+from mx_bluesky.common.parameters.rotation import SingleRotationScan
+
+
+class MultiRotationScanByTransmissions(SingleRotationScan):
+    transmission_fractions: list[Float]
+    transmission_frac: float = -1
+
+    @field_validator("transmission_frac")
```
olliesilvester, line 13: I don't really like this, but I couldn't think of a quick+nice way to have a parameter model which allowed multiple transmission fractions, since the base DiffractionExperiment doesn't allow it to be a list. I made a note [here](https://github.com/DiamondLightSource/mx-bluesky/issues/1308)

src/mx_bluesky/beamlines/i24/jungfrau_commissioning/experiment_plans/rotation_scan_plan.py
```diff
@@ -0,0 +1,274 @@
...
+            @bpp.run_decorator(
+                md={
+                    "subplan_name": I24PlanNameConstants.ROTATION_META_READ,
+                    "scan_points": [params.scan_points],
+                    "rotation_scan_params": params.model_dump_json(),
+                }
+            )
+            # Write metadata json file
+            def _do_read():
+                yield from read_devices_for_metadata(composite)
```
olliesilvester: Must: The hotfix on the beamline used a new `PathProvider` which worked out its correct path during `prepare`, so the `_do_read` either needs to happen after the writer is prepared, or we rethink how the logic will work. Again, this will be easier if we do #1336 first

src/mx_bluesky/common/parameters/components.py
```diff
@@ -152,7 +155,9 @@ class WithVisit(BaseModel):
     det_dist_to_beam_converter_path: str = Field(
         default=DetectorParamConstants.BEAM_XY_LUT_PATH
     )
-    insertion_prefix: str = "SR03S" if TEST_MODE else "SR03I"
+    insertion_prefix: str = (
```
rtuck99: This leads to a lot of ugliness. We shouldn't be initialising the beamline name in here on module load. The logic setting this also seems highly suspect.

What do we even use `insertion_prefix` for? I can't find anything in the code. Can we get rid of it?

src/mx_bluesky/common/parameters/device_composites.py
```diff
@@ -63,3 +67,10 @@ class GridDetectThenXRayCentreComposite(FlyScanEssentialDevices):
     zebra: Zebra
     robot: BartRobot
     sample_shutter: ZebraShutter
+
+
+class GonioWithXYZOmega(Readable, Protocol):
```
rtuck99: Where is this used?

src/mx_bluesky/common/parameters/constants.py
```diff
@@ -135,6 +137,7 @@ class PlanGroupCheckpointConstants:
     MOVE_GONIO_TO_START = "move_gonio_to_start"
     READY_FOR_OAV = "ready_for_oav"
     PREPARE_APERTURE = "prepare_aperture"
+    SETUP_ZEBRA_FOR_ROTATION = "setup_zebra_for_rotation"
```
rtuck99: This string is the default in `setup_zebra.py` but it doesn't reference the constant. We also specify it explicitly as the group name but then don't wait on it outside the function in which it is used.

src/mx_bluesky/common/parameters/device_composites.py
olliesilvester: Will add to the rotation device composite. In general I'd like to start typing all of plan "required devices" as protocols rather than the devices themselves.

olliesilvester: Actually I'm not going to worry about this for now. Will remove this

src/mx_bluesky/beamlines/i24/parameters/rotation.py
rtuck99: This does seem like yet another case of our parameter model getting in the way.

The validation seems sensible, I can't think of any way of doing things that doesn't involve wholesale refactoring of our current parameter model.

src/mx_bluesky/beamlines/i24/jungfrau_commissioning/experiment_plans/rotation_scan_plan.py
```diff
@@ -194,40 +228,27 @@ def _rotation_scan_plan(
...
-
-            @bpp.subs_decorator([metadata_writer])
-            @bpp.set_run_key_decorator(I24PlanNameConstants.ROTATION_META_READ)
-            @bpp.run_decorator(
-                md={
-                    "subplan_name": I24PlanNameConstants.ROTATION_META_READ,
-                    "scan_points": [params.scan_points],
-                    "rotation_scan_params": params.model_dump_json(),
-                }
+            # Should check topup gate here, but not yet implemented,
```
rtuck99: Are we actually currently using the external callbacks for JF?

rtuck99:
I think this last batch of changes could probably have done with being another PR instead of making an already large PR larger.

src/mx_bluesky/beamlines/i24/jungfrau_commissioning/experiment_plans/rotation_scan_plan.py
olliesilvester: Nope, internal only for JF right now

PR #1267, opened 2025-09-15
Use in-built ophyd-async code for checking smargon limits

### LINKED ISSUE #1263: Remove smargon specific limit checking
With https://github.com/DiamondLightSource/dodal/pull/1511 merged we can now remove specific checking on the smargon:

* The code in https://github.com/DiamondLightSource/mx-bluesky/blob/main/src/mx_bluesky/hyperion/device_setup_plans/smargon.py can now just try and do the move, catch the `MotorLimitsException` if it occurs and re-raise it as a `SampleException`.
* Confirm that there is nowhere else in `dodal` or `mx-bluesky` now using `get_xyz_limits` on the smargon
* Remove `get_xyz_limits` and any code that exclusively uses e.g. `XYZLimits` and `AxisLimit`

## Acceptance Criteria
- The above is done and tests are updated accordingly

teoching0705:
Fixes #1263

1. Check if anything crashes

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
@@ -290,7 +290,7 @@ def test_given_moving_out_of_range_when_move_with_warn_called_then_warning_excep
 
 @patch(
     "mx_bluesky.hyperion.device_setup_plans.smargon.bps.mv",
-    side_effect=FailedStatus(RuntimeError("RuntimeError")),
+    new=MagicMock(side_effect=FailedStatus(RuntimeError("RuntimeError"))),
 )
 def test_re_raise_failed_status_that_is_not_MotorLimitsException(
     RE: RunEngine, smargon: Smargon
@@ -304,8 +304,8 @@ def test_re_raise_failed_status_that_is_not_MotorLimitsException(
 
 
 @patch(
-    "mx_bluesky.hyperion.device_setup_plans.smargon.move_smargon_warn_on_out_of_range",
-    side_effect=RuntimeError("RuntimeError"),
+    "mx_bluesky.hyperion.device_setup_plans.smargon.bps.mv",
+    new=MagicMock(side_effect=RuntimeError("RuntimeError")),
 )
 def test_does_not_catch_exception_that_is_not_MotorLimitsException(
     RE: RunEngine, smargon: Smargon


src/mx_bluesky/hyperion/device_setup_plans/smargon.py
```diff
@@ -2,19 +2,11 @@
...
 ):
-    """Throws a SampleException if the specified position is out of range for the
-    smargon. Otherwise moves to that position."""
-    limits = yield from smargon.get_xyz_limits()
-    if not limits.position_valid(position):
-        raise SampleException(
-            "Pin tip centring failed - pin too long/short/bent and out of range"
-        )
+    """Moves Smargon to the specific position. Motor limits are handled by ophyd_async"""
     yield from bps.mv(
```
DominicOram, line 10: Must: Unfortunately we need to keep the `SampleException` here for two reasons:
* `SampleException` is a special `WarningException` that doesn't stop the program but instead makes it skip on to the next sample
* The error message in this exception is what gets given to the end user. `pin too long/short/bent and out of range` makes sense to them but a generic "motor out of range" won't.

The solution is to wrap the move in a try/except and convert the MotorLimit exception into a `SampleException` and re-raise

tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
```diff
@@ -282,7 +283,7 @@ def test_given_moving_out_of_range_when_move_with_warn_called_then_warning_excep
 ):
     set_mock_value(smargon.x.high_limit_travel, 10)
 
-    with pytest.raises(WarningException):
+    with pytest.raises(FailedStatus):
```
DominicOram, line 286: Must: See comment above, this test needs to pass without being modified as it is making sure we skip to the next sample rather than kill the process. It's not clear how important that is though so could you add a comment to the top of this test about it.

DominicOram (changes requested):
Thanks, some comments in code

DominicOram (changes requested):
Thanks. See comment in coed, happy to discuss in person if it doesn't make sense

tests/unit_tests/hyperion/experiment_plans/test_pin_tip_centring.py
```diff
@@ -286,6 +286,17 @@ def test_given_moving_out_of_range_when_move_with_warn_called_then_warning_excep
         RE(move_smargon_warn_on_out_of_range(smargon, (100, 0, 0)))
 
 
+@patch(
+    "mx_bluesky.hyperion.device_setup_plans.smargon.move_smargon_warn_on_out_of_range",
+    side_effect=RuntimeError("RuntimeError"),
+)
+def test_catch_exception_that_is_not_MotorLimitsException(
```
DominicOram, line 293: Nit: 
```suggestion
def test_does_not_catch_exception_that_is_not_MotorLimitsException(
```

src/mx_bluesky/hyperion/device_setup_plans/smargon.py
```diff
@@ -9,12 +11,13 @@ def move_smargon_warn_on_out_of_range(
...
-            "Pin tip centring failed - pin too long/short/bent and out of range"
+    smargon. Otherwise moves to that position. The check is from ophyd-async"""
+    try:
+        yield from bps.mv(
+            smargon, CombinedMove(x=position[0], y=position[1], z=position[2])
         )
-    yield from bps.mv(
-        smargon, CombinedMove(x=position[0], y=position[1], z=position[2])
-    )
+    except FailedStatus as fs:
```
DominicOram, line 19: Should: Sorry, I think this will silently throw away any `FailedStatus` Exceptions that are not due to `MotorLimitsException`can we make sure to re-raise those too (with an associated test)

DominicOram (approved):
Great, thank you!

PR #1270, opened 2025-09-15
Refactor common gridscan

olliesilvester:
mx-bluesky side for https://github.com/DiamondLightSource/dodal/pull/1479

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -87,36 +95,80 @@ def detector_params(self):
...
+    def do_set_stub_offsets(self, value: bool):
+        self._set_stub_offsets = value
+
+    @property
+    def grid_1_spec(self):
+        x_end = self.x_start_um + self.x_step_size_um * (self.x_steps - 1)
+        y1_end = self.y_start_um + self.y_step_size_um * (self.y_steps - 1)
+        grid_1_x = Line("sam_x", self.x_start_um, x_end, self.x_steps)
+        grid_1_y = Line("sam_y", self.y_start_um, y1_end, self.y_steps)
+        grid_1_z = Static("sam_z", self.z_start_um)
```
rtuck99, line 124: z_start_um doesn't exist in this class.

Q: Do we/should we still need to specify z (which is presumably constant) in 2d gridscan? Or can we assume this by the current smargon location? It feels like we should

rtuck99 (approved):
Approved although see additional comment for changes that are perhaps outside the immediate scope of this PR, but we should consider doing

PR #1282, opened 2025-09-17
Add docs on logging level

DominicOram:
Adds some more docs as requested by scientist

1. Confirm docs are correct

### Checks for reviewer


rtuck99 (approved):
Approved. Made a couple of commits to fix the formatting and also because Info level is 6 not 5

PR #1288, opened 2025-09-18
Aithre pincentring

### LINKED ISSUE #1269: Aithre: Pin tip centring
Aithre needs to be able to pin tip centre on a sample. We have this ability already in `mx-bluesky` and should reuse it. 

## Notes
* @co2e14 started this at https://github.com/DiamondLightSource/mx-bluesky/tree/aithre-pincentring, which uses https://github.com/DiamondLightSource/dodal/compare/main...aithre-get_move_required_so_that_beam_is_at_pixel_non_smargon
* However, this is done by copying the existing plan to a new one. Instead we should be able to modify the plan so that can be used by both sets of hardware by doing the following:
    * We can just remove the backlight from both
    * We can get the plan to take CONST.HARDWARE.OAV_REFRESH_DELAY as an argument, or move it someone else maybe
    * The plan should take an XYZOmegaStage instead of a Goniometer or Smargon
    * Get `move_smargon_warn_on_out_of_range` to work for `Goniometer` too now we have done https://github.com/bluesky/ophyd-async/issues/933 - There is a gotcha here that Smargon does combine moves whereas Goniometer doesn't let you do that. I think we might have to make Goniometer match the same interface but just do the moves sequentially in the device
* We should then add the plan to somewhere in https://github.com/DiamondLightSource/mx-bluesky/tree/main/src/mx_bluesky/beamlines/aithre_lasershaping so `aithre` can call is

## Acceptance Criteria
- There is a `pin_tip_centre` plan that we think will work for aithre hardware
- The behaviour for other beamlines is completely unchanged

teoching0705:
Fixes #1269 and part of https://github.com/DiamondLightSource/mx-bluesky/issues/1326

Requires https://github.com/DiamondLightSource/dodal/pull/1533

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


teoching0705:
Should be fine for the test IF dodal is on branch aithre-ge_move_required_so_that_beam_is_at_pixel_non_smargon at commit 44c1ce63c9a8fbac128ee5c8e5d4610ae77e7e56

src/mx_bluesky/common/device_setup_plans/xyzomegastage.py
```diff
@@ -0,0 +1,44 @@
...
+            smargon, CombinedMove(x=position[0], y=position[1], z=position[2])
+        )
+    except FailedStatus as fs:
+        if isinstance(fs.__cause__, MotorLimitsException):
+            raise SampleException(
+                "Pin tip centring failed - pin too long/short/bent and out of range"
+            ) from fs.__cause__
+        else:
+            raise fs
+"""
```
DominicOram, line 26: Must: We shouldn't have commented out code in the repository. It makes things harder to read.

src/mx_bluesky/common/device_setup_plans/xyzomegastage.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def move_xyzomegastage_warn_on_out_of_range(
+    xyzomegastage: XYZOmegaStage,
+    position: np.ndarray | list[float] | tuple[float, float, float],
+):
+    """General version of move_smargon_warn_on_out_of_range"""
```
DominicOram, line 33: Should: I wouldn't reference the old one here, just describe the code as it exists currently. After this PR the old code wont exist and everyone will quickly forget what it did

src/mx_bluesky/common/experiment_plans/oav_snapshot_plan.py
```diff
@@ -24,19 +23,16 @@ class OavSnapshotComposite(Protocol):
     smargon: Smargon
     oav: OAV
     aperture_scatterguard: ApertureScatterguard
-    backlight: Backlight
 
 
 def setup_beamline_for_OAV(
     smargon: Smargon,
-    backlight: Backlight,
```
DominicOram, line 32: Should: Why did we have to remove the backlight from here? I think we just need to be able to use the `pin_tip_centre_plan` without a backlight but as far as I can see that plan doesn't call `setup_beamline_for_OAV`?

src/mx_bluesky/common/parameters/constants.py
```diff
@@ -40,6 +40,8 @@ def _get_oav_config_json_path():
         return "tests/test_data/test_OAVCentring.json"
     elif BEAMLINE == "i03":
         return f"/dls_sw/{BEAMLINE}/software/daq_configuration/json/OAVCentring_hyperion.json"
+    elif BEAMLINE == "AITHRE":
+        return "/dls/science/groups/i23/aithre/daq_configuration/json/OAVCentring_aithre.json"
```
DominicOram, line 44: Should: There is an issue here where we won't pick up the beamline as we get it from the environment variable, which isn't filled in on aithre. I think the answer is to put `AITHRE` in the environment variable but we should chat to @co2e14

tests/unit_tests/common/experiment_plans/test_pin_tip_centring.py
```diff
@@ -283,35 +282,41 @@ def test_given_moving_out_of_range_when_move_with_warn_called_then_warning_excep
...
+    set_mock_value(smargon.y.low_limit_travel, -10)
+    set_mock_value(smargon.z.high_limit_travel, 10)
 
     with pytest.raises(WarningException):
-        RE(move_smargon_warn_on_out_of_range(smargon, (100, 0, 0)))
+        RE(move_xyzomegastage_warn_on_out_of_range(smargon, (100, 0, 0)))
+    with pytest.raises(WarningException):
+        RE(move_xyzomegastage_warn_on_out_of_range(smargon, (0, -100, 0)))
+    with pytest.raises(WarningException):
+        RE(move_xyzomegastage_warn_on_out_of_range(smargon, (0, 0, 100)))
```
DominicOram, line 293: Nit: these are good tests but it's usually better to test one thing per test. Take a look at `pytest.mark.paramterize`, might help you here

src/mx_bluesky/common/experiment_plans/pin_tip_centring_plan.py
```diff
@@ -14,27 +14,26 @@
...
 
 
 @pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
 class PinTipCentringComposite:
     """All devices which are directly or indirectly required by this plan"""
 
-    backlight: Backlight
     oav: OAV
-    smargon: Smargon
+    xyzomegastage: XYZOmegaStage
```
DominicOram, line 36: Should: A better name here is `gonio`. I think the class name of `XYZOmegaStage` is fine but the device itself makes more sense to a scientist as a `gonio`. The same is true throughout, e.g. call it `move_gonio_warn_on_out_of_range`

tests/unit_tests/common/experiment_plans/test_pin_tip_centring.py
```diff
@@ -353,11 +352,18 @@ async def test_when_pin_tip_centre_plan_called_then_expected_plans_called(
     test_config_files: dict[str, str],
     RE: RunEngine,
 ):
+    def mock_get_move_plan(gonio, pixel, oav):
+        """Mock bluesky plan"""
+        if False:
+            yield
```
DominicOram, line 358: Should: we would normally do this by doing `yield from bps.null()`.

DominicOram (changes requested):
Great, thanks. Some comments in code. We also need to add an entry point so that we can call it with `BlueAPI`. We can talk through how to do that but basically we want to take a look at the code in https://github.com/DiamondLightSource/mx-bluesky/tree/main/src/mx_bluesky/beamlines/aithre_lasershaping and add a similar thing that calls `pin_tip_centre_plan` with the `aithre` hardware

src/mx_bluesky/common/device_setup_plans/gonio.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+    Throws a SampleException if the specified position is out of range for the
+    gonio. Otherwise moves to that position. The check is from ophyd-async
+    """
+    group = "move_warn_out_of_range"
+    try:
+        yield from bps.abs_set(gonio.x, position[0], group=group, wait=True)
+        yield from bps.abs_set(gonio.y, position[1], group=group, wait=True)
+        yield from bps.abs_set(gonio.z, position[2], group=group, wait=True)
+        yield from bps.wait(group=group)
```
DominicOram: Should: The `wait=True` on each of these means that we're waiting for each move to finish before doing the next one. This means that the `bps.wait` at the end isn't doing anything. If you remove all the `wait=True` it will move them all at the same time. Alternatively you can do:
```suggestion
        yield from bps.mv(gonio.x, position[0], gonio.y, position[1], gonio.z, position[2])
```
To have the same behavior in fewer lines

DominicOram (changes requested):
Thanks, couple of small comments but generally looks good.

PR #1306, opened 2025-09-26
Set beamsize to 20 and aperture to small for xray centring on i04

### LINKED ISSUE #1225: i04 XRC: Always set beamsize to 20um using transfocator
Unlike i03, which uses the apertures, i04 mostly controls beamsize through their transfocator (also known as an f-swicth). For XRC on i04, for now, we always want to use the same small beamsize e.g. 20um. We will probably want to optimise this later as switching it all the time is slow but it's a good start. For subsequent collections after the XRC we will want to change back to the beamsize that was previously set on the f-switch

## Acceptance Criteria
- For XRC on i04 the beamsize is changed to 20um using the transfocator and the large aperture is used throughout the scan
- i03 XRC is unaffected

srishtysajeev:
Fixes #1225

1.Confirm unit tests work

### Checks for reviewer

### DIFF
--- a/.github/workflows/_helm.yml
+++ b/.github/workflows/_helm.yml
@@ -6,7 +6,6 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-
       - name: checkout repo
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


olliesilvester:
We think the failing CI is due to a bug in Bluesky, see https://github.com/bluesky/bluesky/pull/1955#issuecomment-3376801890

Otherwise, logic looks good but see my above comment about making a new decorator

DominicOram (changes requested):
Must: I think this is still going to select the small aperture and we want the big one. This is because `selected_aperture` in `GridCommon` defaults to `SMALL` and we don't override it in GDA.

Should: Ideally I think we should change the interface to the XRC code to instead take a beamsize and then find a way for the plans to delegate the responsibility for this across the beamlines but this is gonna be a bigger job so probably another issue

PR #1320, opened 2025-10-01
Move scintillator in and take an OAV image

### LINKED ISSUE #1305: Create a plan for moving scintillator in and taking an image of it
The first step for https://confluence.diamond.ac.uk/x/xoAJF is to be able to move the scintillator in and take an OAV image. The order of operations for this is:

1. Check with the robot that there is no pin mounted. If there is raise an exception with a nice error message.
2. Move the beamstop to data collection position
3. Move the scintillator in (will need https://github.com/DiamondLightSource/dodal/issues/1567)
4. Wait for the above to finish
5. Set transmission to 100%
6. Open the fast shutter
7. Take an OAV image

## Acceptance Criteria
- There is a plan that does the above

srishtysajeev:
Fixes #1305 

Requires https://github.com/DiamondLightSource/dodal/pull/1583

1. Check the plans in oav_imaging.py
2. Check the tests 
3. Make sure the tests pass 

### Checks for reviewer


src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -0,0 +1,83 @@
...
+    # check pin is mounted
+    pin_mounted = yield from bps.rd(robot.gonio_pin_sensor)
+    if pin_mounted == PinMounted.NO_PIN_MOUNTED:
+        raise ValueError("Pin should not be mounted!")
+
+    # feedback check
+    yield from bps.trigger(feedback, wait=True)
+
+    # move beamstop to data collection position
+    beamstop_pos = beamstop.selected_pos
```
olliesilvester, line 52: Should: I would create a separate plan called `_prepare_beamline_for_scintillator_images` which handles all the device moving.

We also probably want the option to revert beamline to previous state, so we have something like

```
revert_params = yield from _get_current_beamline_state(...)

@finalise_decorator(_revert_beamline_to_previous_state(revert_params, revert=revert))
def inner_plan(...):
   yield from _prepare_beamline_for_sicntillator_images(...)
   yield from take_oav_image(...)

yield from inner_plan(...)
```

Does that make sense? Can discuss in person if you'd like

src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -0,0 +1,83 @@
...
+    # take image
+    take_OAV_image(file_path=image_path, file_name=image_name, oav=oav)
+
+
+def take_OAV_image(
+    file_path: str,
+    file_name: str,
+    oav: OAV,
+) -> MsgGenerator:
+    group = "path setting"
```
olliesilvester, line 79: Should: Better to move group names out to global constants. If they're defined within a function then your tests and other files can't use the same variable.

tests/conftest.py
```diff
@@ -567,6 +564,27 @@ async def fake_load(val: SampleLocation):
...
+    scintillator = i03.scintillator(connect_immediately=True, mock=True)
+
+    return scintillator
+
+
+@pytest.fixture
+def shutter(RE: RunEngine):
+    shutter = i03.sample_shutter(connect_immediately=True, mock=True)
+
+    return shutter
```
olliesilvester, line 585: Could: In my opinion we should try to use `init_devices` instead of a beamline's dodal module when creating device fixtures

tests/unit_tests/beamlines/i04/test_oav_imaging.py
```diff
@@ -0,0 +1,155 @@
...
+    take_image,
+    take_OAV_image,
+)
+
+"""
+Check if pin is mounted exception is raised
+Check code continues if pin isn't mounted
+Check you wait on beamstop to be in before opening the shutter
+Check each stub plan is called once and in correct order
+Check that a mock image is created in the correct place
```
olliesilvester, line 26: Should: Remove these comments from here, although it is often useful to have short comments in the test if it's not obvious what a specific part is testing

tests/unit_tests/beamlines/i04/test_oav_imaging.py
```diff
@@ -0,0 +1,155 @@
...
+    messages = assert_message_and_return_remaining(
+        messages,
+        lambda msg: msg.command == "set"
+        and msg.obj.name == "sample_shutter"
+        and msg.args[0] == ZebraShutterState.OPEN,
+        # and msg.kwargs["wait"] == True
+    )
+
+
+def test_oav_image(
```
olliesilvester, line 124: Should: These tests are good, but ideally we should also have a test which runs the plans through the real run engine, since the run engine can pick up different errors.

src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # feedback check
+    yield from bps.trigger(feedback, wait=True)
+
+    # move beamstop to data collection position
+    beamstop_pos = beamstop.selected_pos
+    yield from bps.abs_set(
+        beamstop_pos, BeamstopPositions.DATA_COLLECTION, group=initial_wait
+    )
+
```
srishtysajeev: # should move backlight out

tests/conftest.py
```diff
@@ -581,6 +581,52 @@ async def fake_load(val: SampleLocation):
...
+# from ophyd_async.core import Reference
+
+# @pytest.fixture
+# def scintillator(RE: RunEngine):
+#     scintillator = Scintillator(
+#         "", aperture_scatterguard=Reference[ApertureScatterguard], beamline_parameters=GDABeamlineParameters, name="scintillator"
+#     )
+
+#     return scintillator
+
```
srishtysajeev: Need to get rid of these comments and ask  what to put down for the device params

olliesilvester (dismissed):
Thanks, some more comments. Also, my previous comment https://github.com/DiamondLightSource/mx-bluesky/pull/1320/files#r2413638178 still stands

PR #1355, opened 2025-10-11
Added waits to stage/unstage

### LINKED ISSUE #1304: Confirm we wait on stage/unstage when appropriate
https://github.com/DiamondLightSource/mx-bluesky/pull/1216/commits/0bcf53606cf20c08b9bc56db3397ae54a66877e9 Fixed a bug where we were unstaging without waiting. I think that generally it might be assumed `stage`/`unstage` does wait by default. We should double check all the instances we use it and confirm if they should be `wait=True` or not.

## Acceptance Criteria
- Confirmed that all uses of `stage`/`unstage` appropriately wait

DominicOram:
Fixes #1304

Also removed a bunch of `type: ignore`s we no longer needed

1. Confirm tests pass

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/beamlines/i24/jungfrau_commissioning/test_plan_utils.py
+++ b/tests/unit_tests/beamlines/i24/jungfrau_commissioning/test_plan_utils.py
@@ -4,8 +4,9 @@ from unittest.mock import AsyncMock
 
 import bluesky.plan_stubs as bps
 import pytest
+from bluesky import FailedStatus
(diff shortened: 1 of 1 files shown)


DominicOram:
> I think it would be nice if there were more unit tests.

I'll add some.

> because with the old-style devices, bps.unstage() does this

This explains why we're only just starting to see this more now. I wonder if it's too late to get the default to `wait=True`...

PR #1383, opened 2025-10-20
Murko: use correct oav for each centring

### LINKED ISSUE #1312: Murko: Use the correct OAV for each centring
When centring in murko we have two different ways of reading data from the OAV:
* The `oav_to_redis_forwarder` directly forwards the raw data to redis
* We also read the beam centre and microns per pixel from the OAV device to put into redis

We have two `OAV` devices, one for the full screen mode and one for the ROI mode. After thawing in one direction we switch the forwarder to start using images from the ROI mode OAV but we do not update which one we're using for the `beam_centre` and `microns_per_pixel` metadata.

We can probably naively fix this by the plan also requiring a ROI OAV device but then this requires 3 different OAV-ish devices, which seems overly complex 

## Acceptance Criteria
- When we switch to ROI mode on the forwarder during murko thawing we also switch which metadata we're transmitting

jacob720:
Fixes #1312, https://github.com/DiamondLightSource/dodal/issues/1619 is also relevant.

Link to dodal PR: https://github.com/DiamondLightSource/dodal/pull/1644

1. Confirm the correct OAV is used for each centring

### Checks for reviewer


src/mx_bluesky/beamlines/i04/thawing_plan.py
```diff
@@ -196,58 +192,98 @@ def _thaw_and_stream_to_redis(
...
-            oav_to_redis_forwarder.sample_id,
-            sample_id,
-            oav_to_redis_forwarder.selected_source,
-            Source.FULL_SCREEN.value,
+        inital_velocity = yield from bps.rd(smargon.omega.velocity)
+        new_velocity = abs(rotation / time_to_thaw) * 2.0
+
+        def cleanup():
+            yield from bps.abs_set(smargon.omega.velocity, inital_velocity, wait=True)
+            yield from bps.abs_set(thawer.control, OnOff.OFF, wait=True)
```
DominicOram: Should: I think we can just combine the cleanups

tests/unit_tests/beamlines/i04/callbacks/test_murko_callback.py
```diff
@@ -13,28 +13,31 @@
...
 
-test_oav_event = event_template("oav_to_redis_forwarder-uuid", test_oav_uuid)
-test_smargon_event = event_template("smargon-omega", test_smargon_data)
+test_oav_event = event_template(
+    {
+        "oav_to_redis_forwarder-uuid": test_oav_uuid,
+        "oav_full_screen-microns_per_pixel_x": 1.2,
+        "oav_full_screen-microns_per_pixel_y": 2.5,
+        "oav_full_screen-beam_centre_i": 158,
+        "oav_full_screen-beam_centre_j": 452,
```
DominicOram: Should: I think it would be good to have a test for if we do a full screen event then roi event all the metadata is updated as we expect (and vice versa)

src/mx_bluesky/beamlines/i04/thawing_plan.py
```diff
@@ -110,146 +117,172 @@ def thaw_and_murko_centre(
...
-    @run_decorator(
-        md={
-            "microns_per_x_pixel": microns_per_pixel_x,
-            "microns_per_y_pixel": microns_per_pixel_y,
-            "beam_centre_i": beam_centre_i,
-            "beam_centre_j": beam_centre_j,
-            "zoom_percentage": zoom_percentage,
-            "sample_id": sample_id,
-        }
+    source: Source,
```
DominicOram: Nit: `source` is a bit of a generic name.

src/mx_bluesky/beamlines/i04/thawing_plan.py
```diff
@@ -110,146 +117,172 @@ def thaw_and_murko_centre(
...
+
         yield from bps.rel_set(smargon.x, x_predict)
         yield from bps.rel_set(smargon.y, y_predict)
         yield from bps.rel_set(smargon.z, z_predict)
 
-        yield from bps.kickoff(oav_to_redis_forwarder, wait=True)
-
-    sample_id = yield from bps.rd(robot.sample_id)
-    yield from bps.mv(murko_results.sample_id, str(sample_id))
+    @subs_decorator(murko_callback)
```
DominicOram: Could: You could use the `contingency_decorator` here instead of the wrapper below (`@bpp.contingency_decorator(final_plan=cleanup)`). For every Bluesky wrapper there is also a decorator, which one you use is basically up to which looks cleanest in that situation.

PR #1406, opened 2025-10-27
Tidy release notes

### LINKED ISSUE #878: Tidy up release notes
With `mx_bluesky` expanding to be more than just hyperion it would be good to categorise the release notes better. i.e. something like https://github.com/DiamondLightSource/mx-bluesky/releases/tag/v1.4.6

It looks like some automated categorisation is possible with https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

We should at least automatically produce the categories, even if we don't fill them all in.

## Acceptance Criteria
- Auto generated release notes have more structure
- If there is a workflow change needed for this (e.g. adding labels) it is documented and agreed upon

jacob720:
Fixes #878 

Release notes should now have sections based on GH labels attached to PRs. There is also now an action that fails if a PR doesn't have a label attached, to encourage developers to add labels.

An example of generated release notes can be seen here: https://github.com/jacob720/mx-bluesky-fork/releases/tag/test_release.

Labels don't correspond 1 to 1 to categories. For example, the `dev experience` and `refactor` labels are put into one category in generated release notes. Not all labels have a category. Changes under these labels, or no label, will go into the `Other Changes` category. 

I can make the action smarter so that it fails if a PR does have a label, but not one with an associated category, but I don't think this is a good idea as there may be changes which don't fit into any category. It's easy to create a GH label so the CI passes, but annoying to have to go into `release.yaml` to add a category in this instance, and it's probably fine for this type of change to go into `Other Changes`.

1. Confirm action passes when the PR has a label attached, and fails when it does not
2. Check there's enough categories and labels

### Checks for reviewer

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -17,6 +17,9 @@ changelog:
     - title: I04
       labels:
         - i04
+    - title: I04-1
+      labels:
+        - i04-1
     - title: I23
       labels:
         - i23


.github/release.yml
```diff
@@ -0,0 +1,41 @@
...
+        - jungfrau
+    - title: Murko
+      labels:
+        - murko
+    - title: I03
+      labels:
+        - i03
+    - title: I04
+      labels:
+        - i04
```
DominicOram: Should: Missed out i04-1
```suggestion
    - title: I04
      labels:
        - i04
    - title: I04-1
      labels:
        - i04-1
```

PR #1411, opened 2025-10-28
User documentation update as per feedback

### LINKED ISSUE #1344: Docs feedback
* Asynchronous operation, need to be clearer what is happening at the same time so that if you get an error in one you know it might affect others. Might be useful to do a chart a bit like the code map that shows which operations are happening at the same time
* Smargon set values need to be mentioned in known issues
* We should add an advanced section with, where we move:
    * The alerts
    * Server install
    * Special PVs
* Separate troubleshooting into a different page from logs
* Add a note on the config page to say some will be removed

## Acceptance Criteria
* The above is fixed
* Once this is merged send around to #mx-beamlines

rtuck99:
Fixes #1344 

* There is now an Advanced section to the User Guide, which now contains relocated sections:
   * Special PVs
   * Server Install
   * Alerts
additionally
   * Reference sequence diagrams showing the order and concurrency of operations during data collection for diagnostic purposes

* Logging and Troubleshooting are now separate sections
* The config page now notes that some options will be removed in a future release

* Beamline scientists to re-review the user documentation

### Checks for reviewer

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,3 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
-# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
-
 version: 2
 updates:
   - package-ecosystem: "github-actions"
@@ -26,3 +21,6 @@ updates:
           - "*"
     commit-message:
       prefix: "chore"
+    ignore:
+      - dependency-name: "pyright"
+        versions: [ ">=1.1.407" ]  # Until https://github.com/microsoft/pyright/issues/11060 is fixed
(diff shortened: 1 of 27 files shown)


docs/user/hyperion/advanced/operations.rst
```diff
@@ -0,0 +1,46 @@
...
+Note that not all of the operations are shown, in order to reduce the complexity of the diagrams, and some of the 
+steps have been slightly simplified.
+
+In conjunction with the `Code Map`_ this should provide a good overview of operations that may assist in interpreting
+logging output should any issues occur.
+
+
+.. _`Code Map`: ../../../developer/code-map/index.rst
+
+Robot Load and X-ray Centring
```
DominicOram: These are great diagrams, thank you!

DominicOram (changes requested):
Great, thank you! Just minor things really:
* In the index page where we're giving an overview of what Hyperion does it would be good to link to the overview of operations page, maybe instead of the code map.
* We need to add the smargon issue to the known issues

DominicOram (approved):
Great, thank you!

PR #1429, opened 2025-11-05
Check beamstop using ipin diode on udc_default_state

### LINKED ISSUE #351: Checks after hutch opening
We need to check that the beamstop is in the beam after every time the hutch is open, this is done using the diode on the front of the detector.
* Move sample out of the way
* Move detector in the correct location
* Check beam
* Check beamstop

rtuck99:
Fixes #351

* DiamondLightSource/dodal#1695

* On acquiring the baton, the default state check should now include a beamstop check which measures the ipin current 
   * When the beamstop is out, this current must be above a threshold defined in the beamlineParameters file `ipin_threshold`
   * When the beamstop is in, this current must be below the same threshold

This check is performed if the feature flag setting `gda.mx.hyperion.enable_beamstop_diode_check` is true in `domain.properties`
Otherwise the previous behaviour where the beamstop is simply moved into position is retained.

If any of the explicit checks in UDC default state plan or the beamstop check plan fail, this should no longer cause Hyperion to exit, and instead it should just release the baton and wait for it to be granted.

### Checks for reviewer

1. Review the concurrency of the various udc default state actions - ~I have left the existing `udc_default_state` actions largely untouched in their own group which overlaps the beamstop check - possibly these should be regrouped, and whether it would be safer to wait for their completion to avoid collisions - e.g. the collimation table motion currently carries on through the beamstop check~. Also the pre- and post- beamstop checks perhaps could be shuffled slightly - they are currently in two groups but it is somewhat arbitrary.
2. Whether some of the things that are currently side-effects of beamstop check should actually just be part of udc default state (this distinction is somewhat artificial since the beamstop check is unconditional)
3. Detector z-axis limits are in `domain.properties` but I think they ought to live in `beamlineParameters` - not sure how I feel about the behaviour of config client falling back to hard-coded defaults in the event of not being able to read the file, seems like this should be a hard fail.
4. 
5. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,265 @@
...
+
+def _verify_no_sample_present(robot: BartRobot):
+    sample_id = yield from bps.rd(robot.sample_id)
+    pin_mounted = yield from bps.rd(robot.gonio_pin_sensor)
+
+    if sample_id or pin_mounted != PinMounted.NO_PIN_MOUNTED:
+        # Cannot unload this sample because we do not know the correct visit for it
+        raise UnexpectedSampleError(
+            "An unexpected sample was found, please unload the sample manually."
+        )
```
DominicOram: Must: I think we need a solution for this to happen automatically. Currently we are able to hand the baton to UDC with a loaded sample so the scientists will consider this a regression. I think the answer is probably not to record this unload in ispyb. If we just did the load of the next sample we wouldn't record the unload of this one anyway

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,265 @@
...
+    yield from bps.abs_set(devices.collimation_table.upstream_y, 0, group="udc_default")
+    yield from bps.abs_set(devices.collimation_table.upstream_x, 0, group="udc_default")
+    yield from bps.abs_set(
+        devices.collimation_table.downstream_x, 0, group="udc_default"
+    )
+
+    beamline_parameters = get_beamline_parameters()
+    # XXX Are the above movements compatible with the beamstop check or should they
+    # be completed first instead of executing in parallel?
+    yield from move_beamstop_in_and_verify_using_diode(devices, beamline_parameters)
```
DominicOram: Could: I would consider not actually calling this here until we have tested the `move_beamstop_in_and_verify_using_diode` well in isolation. If it goes wrong we fry the detector and it's not a 100% required thing anyway

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,265 @@
...
+    if sample_current_uA < minimum_threshold_current_uA:
+        raise SampleCurrentBelowThresholdError(
+            f"Unable to perform beamstop check - "
+            f"Sample current {sample_current_uA}uA below threshold "
+            f"{minimum_threshold_current_uA}uA."
+        )
+
+    yield from bps.abs_set(
+        devices.backlight, InOut.OUT, group=_GROUP_PRE_BACKGROUND_CHECK
+    )
```
DominicOram: Should: Surely we need to do this before we check there is a current? Otherwise the backlight will be blocking the beam

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,265 @@
...
+    config_client = get_hyperion_config_client()
+    detector_current_z = yield from bps.rd(devices.detector_motion.z)
+    features_settings: HyperionFeatureSetting = config_client.get_feature_flags()
+    detector_min_z = features_settings.DETECTOR_DISTANCE_LIMIT_MIN_MM
+    detector_max_z = features_settings.DETECTOR_DISTANCE_LIMIT_MAX_MM
+    target_z = max(min(detector_current_z, detector_max_z), detector_min_z)
+    if detector_current_z != target_z:
+        LOGGER.info(
+            f"Detector distance {detector_current_z}mm outside acceptable range {detector_min_z} <= z <="
+            f" {detector_max_z}, moving it."
```
DominicOram: Should: As above, if these are ranges for the diode they should mention the word diode

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,157 @@
...
+)
+from mx_bluesky.hyperion.external_interaction.config_server import (
+    get_hyperion_config_client,
+)
+from mx_bluesky.hyperion.parameters.constants import HyperionFeatureSetting
+
+_GROUP_PRE_BEAMSTOP_CHECK = "pre_beamstop_check"
+_GROUP_POST_BEAMSTOP_CHECK = "post_beamstop_check"
+
+_PARAM_DATA_COLLECTION_MIN_SAMPLE_CURRENT = "dataCollectionMinSampleCurrent"
```
DominicOram: Should: This isn't used?

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,157 @@
...
+    )
+    yield from bps.abs_set(
+        devices.collimation_table.upstream_x, 0, group=_GROUP_PRE_BEAMSTOP_CHECK
+    )
+    yield from bps.abs_set(
+        devices.collimation_table.downstream_x, 0, group=_GROUP_PRE_BEAMSTOP_CHECK
+    )
+
+    # Wait for all of the above to complete
+    yield from bps.wait(group=_GROUP_PRE_BEAMSTOP_CHECK, timeout=0.1)
```
DominicOram: Could: I'm not sure why we need to wait for these? I don't see how there could be a collision?

src/mx_bluesky/hyperion/experiment_plans/udc_default_state.py
```diff
@@ -0,0 +1,157 @@
...
+        ApertureValue.SMALL,
+        group=_GROUP_POST_BEAMSTOP_CHECK,
+    )
+
+    yield from bps.abs_set(
+        devices.cryostream.course, CryoInOut.IN, group=_GROUP_POST_BEAMSTOP_CHECK
+    )
+    yield from bps.abs_set(
+        devices.cryostream.fine, CryoInOut.IN, group=_GROUP_POST_BEAMSTOP_CHECK
+    )
```
DominicOram: Could: I don't think these movements have any chance of affecting the beamstop check, might as well do them before like everything else?

src/mx_bluesky/common/device_setup_plans/xbpm_feedback.py
```diff
@@ -15,8 +17,13 @@ def unpause_xbpm_feedback_and_set_transmission_to_1(
         xbpm_feedback (XBPMFeedback): The XBPM device that is responsible for keeping
                                       the beam in position
         attenuator (BinaryFilterAttenuator): The attenuator used to set transmission
+        timeout_for_stable: If specified and non-zero, specifies the time in seconds to wait for
+            feedback to stabilise, otherwise we do not wait.
     """
     yield from bps.mv(xbpm_feedback.pause_feedback, Pause.RUN, attenuator, 1.0)
+    if timeout_for_stable:
```
DominicOram: Could: I'm not sure I like adding this in to this plan. It feels like a separate thing.

src/mx_bluesky/hyperion/experiment_plans/beamstop_check.py
```diff
@@ -0,0 +1,231 @@
...
+    if ipin_in_beam_uA > beamstop_threshold_uA:
+        raise BeamstopNotInPositionError(
+            f"Ipin is too high at {ipin_in_beam_uA} - check that beamstop is "
+            f"in the correct position."
+        )
+
+
+def _check_ipin(devices: BeamstopCheckDevices):
+    yield from bps.abs_set(devices.sample_shutter, ZebraShutterState.OPEN, wait=True)
+    yield from bps.sleep(1)  # wait for reading to settle
```
DominicOram: Could: We could instead wait for the ipin to be above/below the expected threshold with a timeout of 1s. Will speed the plan up a bit.

tests/unit_tests/hyperion/experiment_plans/conftest.py
```diff
@@ -370,3 +377,62 @@ def beamline_specific(
...
+):
+    async def noop(_):
+        await asyncio.sleep(0)
+
+    run_engine.register_command("sleep", noop)
+    try:
+        async with init_devices(mock=True):
+            beamstop = Beamstop("", MagicMock())
+
+        with patch_all_motors(beamstop):
```
DominicOram: Nit: it's a bit odd that all of the below is done inside this context when I think only the yield needs to be?

tests/unit_tests/hyperion/experiment_plans/test_beamstop_check.py
```diff
@@ -0,0 +1,457 @@
...
+)
+@patch(
+    "mx_bluesky.hyperion.experiment_plans.beamstop_check._post_beamstop_out_check_actions"
+)
+def test_beamstop_check_checks_beamstop_out_diode_above_threshold_before_second_check(
+    mock_post_beamstop_out_actions,
+    beamstop_check_devices: BeamstopCheckDevices,
+    run_engine: RunEngine,
+    beamline_parameters: GDABeamlineParameters,
+    ipin_reading: float,
```
DominicOram: Nit: 
```suggestion
    ipin_reading_with_beamstop_out: float,
```

DominicOram (changes requested):
Thanks, broadly correct but a few comments

src/mx_bluesky/common/device_setup_plans/xbpm_feedback.py
rtuck99: Does it? To me it feels like they aren't separate, and I want to know what the postconditions of this method are when I call it.

src/mx_bluesky/hyperion/baton_handler.py
```diff
@@ -121,8 +122,21 @@ def release_baton() -> MsgGenerator:
         yield from bps.abs_set(baton.current_user, NO_USER, wait=True)
         _raise_baton_released_alert(get_alerting_service(), previous_requested_user)
 
+    def trap_default_state_exception(e: Exception):
+        if isinstance(e, BeamlineCheckFailureError):
+            LOGGER.warning("Caught default state check failure:", exc_info=e)
+            yield from bps.null()
+        else:
+            LOGGER.warning("Caught unexpected exception", exc_info=e)
```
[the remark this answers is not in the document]
rtuck99: Given that normal experiment plans don't cause it to exit at the moment, I've made a change so that it re-raises udc default state exceptions by wrapping in `PlanError` instead, so that an error in the default state plan will recover, but an error when handing over the baton won't.

DominicOram (changes requested):
Thanks. Sorry, still a couple of points but basically there now.

src/mx_bluesky/hyperion/baton_handler.py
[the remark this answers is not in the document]
rtuck99: We will effectively get that for free anyway when we move to blueapi

PR #1436, opened 2025-11-11
Optionally use fastcs eiger for XRC

### LINKED ISSUE #1076: Migrate to the ophyd-async Eiger
Following https://github.com/DiamondLightSource/dodal/pull/1187 we have a basic example of how to use the `ophyd-async` Eiger. We should add the ability to use this in Hyperion. 

We will need some time for soak testing, particularly as the `ophyd-async` Eiger uses a different IOC, which is running on K8s. So we will need to be able to control which Eiger we use from a GDA property (e.g. a parameter in the model sent to Hyperion)

## Acceptance Criteria
- There is a parameter I can send to Hyperion which makes us use the `ophyd-async` Eiger

jacob-williamson:
Fixes #1076 

Requires https://github.com/bluesky/ophyd-async/pull/1127 and https://github.com/DiamondLightSource/dodal/pull/1706/

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/parameters/device_composites.py
+++ b/src/mx_bluesky/common/parameters/device_composites.py
@@ -24,11 +24,12 @@ from dodal.devices.xbpm_feedback import XBPMFeedback
 from dodal.devices.zebra.zebra import Zebra
 from dodal.devices.zebra.zebra_controlled_shutter import ZebraShutter
 from dodal.devices.zocalo import ZocaloResults
+from ophyd_async.fastcs.eiger import EigerDetector as FastCSEiger
 
 
 @pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
 class FlyScanEssentialDevices:
-    eiger: EigerDetector
+    eiger: EigerDetector | FastCSEiger
     synchrotron: Synchrotron
     zocalo: ZocaloResults
     smargon: Smargon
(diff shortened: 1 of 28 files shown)


src/mx_bluesky/common/parameters/device_composites.py
```diff
@@ -23,11 +23,12 @@
 from dodal.devices.zebra.zebra import Zebra
 from dodal.devices.zebra.zebra_controlled_shutter import ZebraShutter
 from dodal.devices.zocalo import ZocaloResults
+from ophyd_async.fastcs.eiger import EigerDetector as FastCSEiger
 
 
 @pydantic.dataclasses.dataclass(config={"arbitrary_types_allowed": True})
 class FlyScanEssentialDevices:
-    eiger: EigerDetector
+    eiger: EigerDetector | FastCSEiger
```
DominicOram: Must: Annoyingly I don't think this is going to work. The code in https://github.com/DiamondLightSource/mx-bluesky/blob/8b6a2fc94bcc552a3b0e08e242eda20877a365ed/src/mx_bluesky/common/utils/context.py#L45 will populate the composite based on the parameter name being the same as the device name and the type being the same. I think you will instead need to have a new device in this composite called `fastcs_eiger: FastEiger`

DominicOram (changes requested):
Thanks, some comments in code. When you test it I suspect you might get issues with not calling `complete` and instead calling `unstage` (at https://github.com/DiamondLightSource/mx-bluesky/blob/76a7e489c41d38f56fa12ff5293aa3f49b623477/src/mx_bluesky/common/experiment_plans/inner_plans/do_fgs.py#L106). I think we just try that and see though

olliesilvester (commented):
Noticed a couple of things after briefly skimming

rtuck99:
Closing because replaced by other PRs as noted above

PR #1437, opened 2025-11-11
Murko: continue thaw if redis is down

### LINKED ISSUE #810: Murko: Continue thawing if valkey is down
Currently, if valkey is down the `mx-bluesky` plan will crash, the thawing will stop and then GDA will stop.

We should instead print a warning in GDA and carry on with the thawing and whatever GDA is doing next

## Acceptance Criteria
- If valkey is down thawing still occurs
- If valkey is down GDA will print a warning but continue

jacob720:
Fixes #810 

Requires https://github.com/DiamondLightSource/dodal/pull/1750

1. Check tests pass and thaw happens when redis connection fails.

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/beamlines/i04/callbacks/test_murko_callback.py
+++ b/tests/unit_tests/beamlines/i04/callbacks/test_murko_callback.py
@@ -291,7 +291,7 @@ def test_if_redis_connection_fails_then_there_is_no_error(
     callback.stop(doc)
 
 
-def test_rwarning_is_logged_if_redis_connection_fails(caplog):
+def test_warning_is_logged_if_redis_connection_fails(caplog):
     callback = MurkoCallback("", "")
     doc = {}
     callback.start(doc)


DominicOram (changes requested):
Thanks. This works but I think it might be cleaner if it was handled inside the `MurkoCallback` and the `MurkoResultsDevice`. They can both just do nothing and warn, then the `MurkoResultsDevice` can keep the positions all at 0 and the plan would never have to know there's a redis issue

DominicOram:
I should add my reasoning here. Basically, what if we wanted to make use of the `MurkoCallback` again? We would have to put a similar check everywhere we wanted to use it, whereas if we make the callback itself handle it then we don't have to worry about it. Also, we would have less places we're trying to make redis connections, which is nice

DominicOram (approved):
Thanks, couple of minor things, take them or leave them

PR #1443, opened 2025-11-11
Remove VerbosePlanExecutionLoggingCallback

### LINKED ISSUE #1442: Remove VerbosePlanExecutionLoggingCallback
`VerbosePlanExecutionLoggingCallback` is only used in tests and even then we don't actually test anything against it. It can be useful for debugging but I don't think it's really been used for a year or so and it's trivial to write something similar if we need to

## Acceptance Criteria
- `VerbosePlanExecutionLoggingCallback` is removed

DominicOram:
Fixes #1442 

1. Confirm tests still pass

### Checks for reviewer


rtuck99 (approved):
Approved

PR #1445, opened 2025-11-12
Use beamsize device to get beamsize

### LINKED ISSUE #1424: I04 XRC: Put beamsize in ispyb correctly
Currently, when running a bluesky XRC beamsize is being put in ispyb as 80x20 e.g. 

Image

This is incorrect, the real beamsize can be read from the transfocator, specifically `BL04I-MO-FSWT-01:VER` and `BL04I-MO-FSWT-01:HOR`. We should read these during collection and put the data into ispyb. I03 does not have a transfocator and so they should use the beamsize as currently calculated.

## Acceptance Criteria
- Beamsize is read from the above PVs on i04
- This is what is put into ispyb
- Behaviour on i03 is unchanged

jacob720:
Fixes #1424 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/1730

Uses the beamsize devices created in the dodal PR above, to correctly enter beamsize information into ispyb during the ispyb callback.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.13.7",
     "bluesky >= 1.14.6",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@fix_beamsize_devices",
+    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
 ]
 
 


src/mx_bluesky/common/experiment_plans/inner_plans/read_hardware.py
```diff
@@ -65,13 +66,16 @@ def standard_read_hardware_during_collection(
...
     detector: EigerDetector,
+    beamsize: BeamsizeBase,
 ):
     signals_to_read_during_collection = [
         aperture_scatterguard,
         attenuator.actual_transmission,
         flux.flux_reading,
         dcm.energy_in_keV,
         detector.bit_depth,
+        beamsize.x_um,
```
DominicOram: Could: You can just read the whole `beamsize` device, rather than individual signals

PR #1473, opened 2025-11-21
1318 optimise oav transmission new

### LINKED ISSUE #1318: Create plan to optimise transmission using OAV images
Plan should do something like this:
- Move beamline into position for scintillator (+ move scint in)
- Take image of OAV
- Open image and get average pixel saturation fraction
- If saturation is above target value, then lower transmission, if higher than raise transmission
- Do this with a binary search - i.e if saturation is too high then move transmission to (current trans + lowest trans / 2)
- Repeat this until we're sufficiently close to the target saturation value (~50%?)

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

srishtysajeev:
Fixes #1318 

Need to add tests. 

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.14.0",
     "bluesky >= 1.14.6",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@i03-device-manager",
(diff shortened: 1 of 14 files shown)


olliesilvester:
An updated PR for the same changes is [here](https://github.com/DiamondLightSource/mx-bluesky/pull/1496) - a lot of the comments here are still relevant though

PR #1496, opened 2025-12-04
Add plan for automatically finding the beam centre on the OAV

### LINKED ISSUE #1489: Automated OAV Centring: Create a plan that sets the beam centres for different zooms
To do the full OAV centring we should run it for each zoom level and put the data into the appropriate PVs. To test this we should make a plan that:

* Print all the values from `BL04I-EA-OAV-01:FZOOM:PBCX:VALX` e.g. `BL04I-EA-OAV-01:FZOOM:PBCX:VALA`, `BL04I-EA-OAV-01:FZOOM:PBCX:VALB` etc. (so that we can revert them if we need to)
* Print all the values from `BL04I-EA-OAV-01:FZOOM:PBCY:VALX` (so that we can revert them if we need to)
* Move to `BL04I-EA-OAV-01:FZOOM:MP:SELECT.ZRST`
* Find the beam centre with the beam centre device
* Write it to  `BL04I-EA-OAV-01:FZOOM:PBCX:VALA`/ `BL04I-EA-OAV-01:FZOOM:PBCY:VALA`
* Repeat for `BL04I-EA-OAV-01:FZOOM:MP:SELECT.ONST` and `BL04I-EA-OAV-01:FZOOM:PBCX:VALB`/ `BL04I-EA-OAV-01:FZOOM:PBCY:VALB`
* Repeat for all values filled in on `BL04I-EA-OAV-01:FZOOM:MP:SELECT`

The synoptic can be found at:
Image
Image

Note: We should do this for just one zoom level to start and confirm it is getting good data. Then we can do it for all of them.

## Acceptance Criteria
- The above plan exists

srishtysajeev:
Fixes #1489

Requires https://github.com/DiamondLightSource/dodal/pull/1757

To test:
* Confirm tests pass

### DIFF
--- a/src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
+++ b/src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
@@ -268,6 +268,10 @@ def find_beam_centres(
     Finds beam centres at the zoom levels given by zoom_levels_to_centre, first
     optimising transmission if the zoom level is in zoom_levels_to_optimise_transmission.
 
+    Note that the previous beam centre values are used to draw an ROI box around the
+    OAV image when finding updated beam centre. If the previous values are very wrong,
+    this plan may fail or give inaccurate results.
+
     zoom_levels_to_centre: The levels to do centring at, by default runs at all known
                            zoom levels.
     zoom_levels_to_optimise_transmission: The levels to optimise transmission at,


src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -109,7 +134,178 @@ def take_and_save_oav_image(
...
+        ],
+        "5.0x": [
+            oav.zoom_controller.x_placeholder_zoom_6,
+            oav.zoom_controller.y_placeholder_zoom_6,
+        ],
+        "10.0x": [
+            oav.zoom_controller.x_placeholder_zoom_8,
+            oav.zoom_controller.y_placeholder_zoom_8,
+        ],
+    }
```
DominicOram: This should be stored on the device

DominicOram (commented):
Mostly reminders to myself that I will fix but would like an answer on the question around when we optimise transmission

src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -109,7 +130,174 @@ def take_and_save_oav_image(
...
+    xbpm_feedback: XBPMFeedback,
+    transmission: float,
+):
+    yield from bps.trigger(xbpm_feedback, wait=True)
+    yield from bps.mv(attenuator, transmission)
+    yield from bps.trigger(max_pixel, wait=True)
+    return (yield from bps.rd(max_pixel.max_pixel_val))
+
+
+def optimise_transmission_with_oav(
```
olliesilvester: Notes from testing:
- Occasionally this would hit the `StopIteration` for various reasons: We are being overly precise; there is an issue with i04's xbpm PV which is causing us to not wait for long enough,;the error margin on the transmission device is much higher than the PV suggests.
We should keep the binary search, but either increase the tolerence, or call it "done" as soon as the next iteration would involve moving the transmission by <x%

src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
```diff
@@ -109,7 +130,174 @@ def take_and_save_oav_image(
...
         raise FileExistsError("OAV image file path already exists")
+
+
+def _max_pixel_at_transmission(
+    max_pixel: MaxPixel,
+    attenuator: BinaryFilterAttenuator,
+    xbpm_feedback: XBPMFeedback,
+    transmission: float,
+):
+    yield from bps.trigger(xbpm_feedback, wait=True)
```
olliesilvester: During testing we ended up adding a sleep after triggering xbpm feedback. This real fix is to address the issues with the PV though

olliesilvester (changes requested):
This worked as-is on the beamline, but the main things to address are:
- Better validation if you typo the parameters
- Less strict binary search (but we can put that into a new issue)
- Address dodgy xbpm feedback. Could add sleeps in for now, or chase up with controls

src/mx_bluesky/beamlines/i04/oav_centering_plans/oav_imaging.py
DominicOram: I don't think we've properly captured what the issues are with the xbpm feedback. Can we leave as is and write up in more detail at the next round of testing?

olliesilvester:
I will update this PR and put the fix in

PR #1521, opened 2025-12-11
Use config server for oav config

### LINKED ISSUE #1448: Migrate OAV configuration to use config server
The OAV gets configuration from the following files:
- `jCameraManZoomLevels.xml`
- `display.configuration`
- `OAVCentring_hyperion.json` - this is supported by the config server but we don't seem to call it yet

## Acceptance Criteria
- Hyperion / mx-bluesky no longer read OAV configuration directly from the filesystem

jacob720:
Fixes #1448 

Link to dodal PR (if required): https://github.com/DiamondLightSource/dodal/pull/1748

This PR also removes all OAV test config from the repo. Instead it is read from a mocked config server, and all tests in mx-bluesky and dodal are now using the same config. Tests have been changed to reflect this.

1. Confirm config is read through the `daq-config-server`

### Checks for reviewer


rtuck99 (changes requested):
I would rather retain the images and original test values, for instance, when I created the snapshot tests I started off with the existing i03 config files to validate that the images reproduced existing snapshot behaviour, by comparing the output from log files to what the snapshot code generated, then used the same images in the system tests to prove that it works end-to-end. So that is why the config is maintained separately, because those configs have to work for the full load-centre-collect plan and not just for individual unit tests, plus they also have to generate sensible looking results, in combination with all the other config.

rtuck99 (approved):
Approved.

PR #1389, opened 2025-10-21
Ability to launch blueapi and run Hyperion plans

rtuck99:
Fixes
* #1364 

Requires

* DiamondLightSource/blueapi#1242
* ~DiamondLightSource/blueapi#1245~ DiamondLightSource/blueapi#1267
* DiamondLightSource/dodal#1649
* DiamondLightSource/dodal#1760

(remember to update `pyproject.toml` with the dodal commit tag if you need it for tests to pass!)

Note:

This PR requires updating when
* #1359
is merged

### Instructions to reviewer on how to test:

It should now be possible to
1. Launch blueapi using the `run_hyperion.sh` script with the `--blueapi` flag, and connected to mock devices with the `--dev` flag
2. Stop blueapi from the above script
3. Run the `load_centre_collect`,`robot_unload` and `move_to_udc_default_state` plans with blueapi
4. Fetch the plan schemas from the `/plans` endpoint
5. Tests pass

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes
- [ ] Acceptance criteria for parent issue are met

Issue #1365, opened 2025-10-14
Hyperion blueapi supervisor service


rtuck99:
## Acceptance Criteria
- The Hyperion supervisor service can be deployed as a standalone service
- The supervisor service monitors the Hyperion baton PVs and agamemnon as it does currently
- Instead of executing the experiment plans in-process, it makes a REST call execute the hyperion-blueapi plan
- Documentation is updated with deployment instructions
- The supervisor process doesn't have a dependency on any of the internals of `hyperion-blueapi`

PR #1530, opened 2025-12-18
Hyperion supervisor implementation

rtuck99:
Fixes:
* #1365

Requires:
* #1531
* #1389 
* DiamondLightSource/dodal#1804

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/hyperion/supervisor/_supervisor.py
```diff
@@ -0,0 +1,116 @@
...
+                        )
+                        self._run_task_remotely(task_request)
+                    case UDCCleanup():
+                        task_request = TaskRequest(
+                            name="clean_up_udc",
+                            params={"visit": current_visit},
+                            instrument_session=instrument_session,
+                        )
+                        self._run_task_remotely(task_request)
+                    case _:
```
jacob720: It strikes me that currently if we add a plan that agamemnon can request, we'd have to make changes in 3 places:
- `SupvervisorRunner.decode_and_execute`
- `InProcessRunner.decode_and_execute`
- `create_parameters_from_agamemnon`

Could we push some of this logic into new classes that extend or contain the parameter models? 

Something like 
```python
class WithTaskRequest:
    @abstractmethod
    def get_task_request(self, instrument_session) -> TaskRequest: ...


class WithExecute:
    @abstractmethod
    def execute_plan(self, context: BlueskyContext) -> MsgGenerator: ...


class TaskRequestAndExecuteFromLoadCentreCollect(WithTaskRequest, WithExecute):
    def __init__(self, parameters: LoadCentreCollect):
        self.parameters = parameters

    def get_task_request(self, instrument_session):
        return TaskRequest(
            name="load_centre_collect",
            params={"parameters": self.parameters},
            instrument_session=instrument_session,
        )

    def execute_plan(self, context: BlueskyContext):
        devices: Any = create_devices(context)
        yield from self.execute_plan(
            partial(load_centre_collect_full, devices, self.parameters)
        )
        return self.parameters.visit


class TaskRequestFromUDCDefaultState(WithTaskRequest, WithExecute):
    def get_task_request(self, instrument_session):
        return TaskRequest(
            name="load_centre_collect",
            params={},
            instrument_session=instrument_session,
        )

    def execute_plan(self, context: BlueskyContext):
        udc_default_devices: UDCDefaultDevices = device_composite_from_context(
            context, UDCDefaultDevices
        )
        yield from move_to_udc_default_state(udc_default_devices)
```
etc.
Then we could get rid of the switch statements in `SupvervisorRunner.decode_and_execute` and  `InProcessRunner.decode_and_execute` and instead have
```python
def decode_and_execute(
    self, current_visit: str | None, parameter_list: Sequence[WithTaskRequest]
) -> MsgGenerator:
    for parameters in parameter_list:
        current_visit = self._run_task_remotely(parameters.get_task_request(instrument_session))
    return current_visit
```
or
```python
def decode_and_execute(
    self, current_visit: str | None, parameter_list: Sequence[WithExecute]
  ) -> MsgGenerator:
    for parameters in parameter_list:
        current_visit = parameters.execute_plan(self.context)
    return current_visit
```
There may be good reason to not do this, but thought it could be worth discussing.

jacob720 (approved):
Looks good, a couple of comments but happy to merge if you disagree

src/mx_bluesky/hyperion/supervisor/_supervisor.py
rtuck99: `InProcessRunner` will ultimately go away, once we are running happily in blueapi, we can delete it. The abstraction has been created so that we can temporarily switch between both setups. It is annoying that we have to have code that is very similar in both runner implementations, unfortunately there isn't a nice way around it with one method being a generator function and the other not.

It's probably unlikely we will add more plans for agamemnon to run. What is more likely is commissioning plans but these wouldn't be called by `hyperion-supervisor`.

jacob720: That makes sense, thanks for explaining

PR #1535, opened 2026-01-08
External callbacks via stomp

rtuck99:
This addresses the hyperion portion of 

* DiamondLightSource/bluesky-stomp#14

This requires:

* https://github.com/DiamondLightSource/mx-bluesky/pull/1530

* Stomp support is now enabled in the callbacks when running hyperion-blueapi mode when the external callbacks are passed the `--stomp-config` parameter specifying a blueapi config file which contains the stomp configuration

* System tests require the rabbitmq-stomp container from the latest `hyperion-system-test`

### Checks for reviewer


DominicOram (changes requested):
Thanks, I still think this needs the docs I suggested though

Issue #1525, opened 2025-12-12
Hyperion: Stop hyperion if no pin was loaded


DominicOram:
If there is no pin in the puck the robot will report an error (with the error code 25) after loading. Currently hyperion ignores this and will skip the sample during optical centring. However, we would rather Hyperion stop completely as a missing pin could also be an indication of actual robot errors, so we would like to trigger a callout.

Annoyingly the robot gives the error after the smargon is re-enabled:
```
BL03I-MO-SGON-01:DISABLED      2025-12-12 15:07:31.580359 1  
BL03I-MO-SGON-01:DISABLED      2025-12-12 15:07:40.157888 0  
BL03I-MO-ROBOT-01:PRG_ERR_CODE 2025-12-12 15:07:41.112465 25  
```

and the current logic is to try and start centring the minute the smargon is enabled. Instead of checking the robot error we could double check that there is a pin on the sensor after the smargon has become enabled again and raise if there isn't (though this is sort of re-creating logic in the robot).

## Acceptance Criteria
- Hyperion stops if there is no pin loaded

ndg63276:
The robot could raise error 25 before re-enabling the smargon, I didn't do it that way to start with as the pin sensor was a little flaky, and sometimes the pin wouldn't be sensed until it had warmed up a little. So currently it only raises an error if no pin is detected before it gets back to (roughly) the barcode scanner. I can play with the order, would Hyperion just raise an error then?

Issue #1478, opened 2025-11-24
Aithre: Combine robot load


DominicOram:
The aithre robot load plans and the `mx-bluesky` should be able to share quite a bit more code. See https://github.com/DiamondLightSource/mx-bluesky/pull/1435#pullrequestreview-3489142638 for a list of some of the differences and ideas to merge them.

## Acceptance Criteria
- `aithre` robot load and `mx-bluesky` robot load use the same code where possible
- Both still work as intended

DominicOram:
The proper solution to the energy change bit is https://github.com/DiamondLightSource/mx-bluesky/issues/224

PR #1537, opened 2026-01-12
Remove smargon enabled check

DominicOram:
Part of #1478 and related to https://github.com/DiamondLightSource/mx-bluesky/issues/1525

Now that we are waiting for the smargon to be enabled in the device (https://github.com/DiamondLightSource/dodal/pull/1791) we no longer need to do it in the plan 

1. Confirm tests still pass

### Checks for reviewer


jacob720 (commented):
Looks good but I think some of the tests need fixing

PR #1541, opened 2026-01-14
Change i04 XRC XBPM behaviour

olliesilvester:
Part of https://github.com/DiamondLightSource/mx-bluesky/issues/1533

i04 has requested the following behaviour for XBPM feedack:
- Immediately before starting the actual grid scan motion, change transmission to desired value and then trigger XBPM feedback
- Don't pause XBPM feedback during collection, don't revert transmission after collection

### Instructions to reviewer on how to test:

1. Confirm that this PR implements the above behaviour

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

PR #1545, opened 2026-01-15
Add xbpm and transmission decorator to pin_centre_then_flyscan_plan

### LINKED ISSUE #1522: Hyperion: XRC should set the transmission and running whilst the xbpm feedback is off
Currently, when running hyperion in UDC we do not obey the transmission specified for the XRC at https://github.com/DiamondLightSource/mx-bluesky/blob/bba467527fe90c87eafadc42166dd28995c36ed7/src/mx_bluesky/hyperion/external_interaction/agamemnon.py#L270, instead we always run with 100% regardless of what this says. This is because we enter the grid detect straight from the pin detect and at no point do we call `transmission_and_xbpm_feedback_for_collection_decorator` for the XRC.

## Acceptance Criteria
- Specified transmission is obeyed for XRCs when Hyperion runs in UDC mode
- XBPM is paused when XRC is run in UDC mode

jacob720:
Fixes #1522

The decorator could also in one of the common plans such as `detect_grid_and_do_gridscan`. Currently we have one in a few places (`i04_default_grid_detect_and_xray_centre`, `hyperion_grid_detect_then_xray_centre` and now `pin_centre_then_flyscan_plan`)  but the logic only gets run when the XRC run starts.

If we will always want to change transmission and pause xbpm feedback we should put it in the common plan.

1. Check test passes
2. Check this is the best place to put the decorator

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/hyperion/experiment_plans/test_pin_centre_then_xray_centre_plan.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_pin_centre_then_xray_centre_plan.py
@@ -422,7 +422,7 @@ def test_pin_tip_centre_then_xray_centre_moves_beamstop_into_place(
 @patch(
     "mx_bluesky.common.experiment_plans.common_flyscan_xray_centre_plan._fetch_xrc_results_from_zocalo"
 )
-def test_pin_tip_centre_then_xray_centre_sets_transmission_fraction_and_xbpm_is_paused(
+def test_pin_tip_centre_then_xray_centre_sets_transmission_fraction_and_xbpm_is_paused_and_both_reverted(
     mock_fetch_zocalo_results: MagicMock,
     mock_run_gridscan: MagicMock,
     mock_grid_detection_plan: MagicMock,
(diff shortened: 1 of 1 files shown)


tests/unit_tests/hyperion/experiment_plans/test_pin_centre_then_xray_centre_plan.py
```diff
@@ -391,3 +398,69 @@ def test_pin_tip_centre_then_xray_centre_moves_beamstop_into_place(
...
+            test_pin_centre_then_xray_centre_params,
+        )
+    )
+    msgs = assert_message_and_return_remaining(
+        msgs,
+        predicate=lambda msg: msg.command == "set"
+        and msg.obj.name == "xbpm_feedback-pause_feedback"
+        and msg.args[0] == Pause.PAUSE,
+    )
+    msgs = assert_message_and_return_remaining(
```
olliesilvester: Could: Also assert the unpause and revert transmission messages

olliesilvester:
> If we will always want to change transmission and pause xbpm feedback we should put it in the common plan.

Discussed in person: Different beamlines will want slightly different behaviour here, eg https://github.com/DiamondLightSource/mx-bluesky/pull/1541

PR #1575, opened 2026-01-22
Fix: hyperion-supervisor unable to serialize load centre collect

### LINKED ISSUE #1564: `hyperion-supervisor` cannot run `load_centre_collect` unable to serialize `parameter_model_version`
See 
https://github.com/DiamondLightSource/mx-bluesky/issues/1556

https://github.com/DiamondLightSource/mx-bluesky/issues/1556#issuecomment-3769487734

rtuck99:
Fixes #1564

`hyperion-supervisor` should now be able to call the load_centre_collect plan on `hyperion-blueapi`.

The `LoadCentreCollectParams` payload replaces the `LoadCentreCollect` payload and no longer includes the paramter model version or duplicates of internal parameters.

1. Tests pass

### Checks for reviewer


src/mx_bluesky/hyperion/external_interaction/agamemnon.py
DominicOram: Must: `select_centres` seems to be missing from `LoadCentreCollectParams`? I think we need to specify at least the `n` (at the moment we only ever use the one selection strategy so I don't think we need to specify that). More worrying is why a test didn't catch this given that we should be forbidding extra fields?

src/mx_bluesky/hyperion/_plan_runner_params.py
```diff
@@ -10,13 +12,13 @@ class Wait(MxBlueskyParameters):
...
-class UDCDefaultState(MxBlueskyParameters):
+class UDCDefaultState(BaseModel):
     """Represents an instruction to execute the UDC default state plan."""
 
     pass
 
 
-class UDCCleanup(MxBlueskyParameters):
+class UDCCleanup(BaseModel):
     """Represents an instruction to perform UDC Cleanup,
```
DominicOram: Should: In `collect()` we're still passing in a `parameter_model_version` for both of these, which seems unnecessary.

Issue #1532, opened 2025-12-18
Remove legacy non-blueapi code


rtuck99:
When hyperion is finally fully migrated to blueapi, we will be able to remove the following code:

* Support for legacy GDA mode aka `GDARunner`, launcher shell scripts support for the same
* Support for the InProcessRunner, launcher code for the same
* hard-coded utility functions for loading i03 beamline and plans, since this is all now done via configuration
* Legacy REST API server code aka `create_app`
* Plan registry
* Any other things we can think of

We can also tidy up the legacy composite factory methods since these are now all invoked by blueapi

## Acceptance Criteria
- All the legacy cruft is *GONE*

Issue #840, opened 2025-02-20
Hyperion: Remove unused endpoints


DominicOram:
The only endpoint used in production now is `load_centre_and_collect`. Given that long term we would like to be calling things via BlueAPI we should remove the endpoints that are not in use.

## Acceptance Criteria
- Unused endpoints are removed
- GDA properties are removed so that we cannot get to these end points
- GDA code to call these end points is removed
- If there are plans that are no longer called we evaluate their future usefulness, potentially removing them

DominicOram:
https://github.com/DiamondLightSource/mx-bluesky/issues/847 is a subset of this

PR #1618, opened 2026-02-06
Remove GDARunner and associated bits

rtuck99:
Fixes
* part of #1532 
* #840

1. Tests pass
2. Documentation is correct and no longer refers to GDA mode

### Checks for reviewer


src/mx_bluesky/hyperion/__main__.py
```diff
@@ -177,30 +44,11 @@ def main():
     """Main application entry point."""
     args = parse_cli_args()
     initialise_globals(args)
-    hyperion_port = HyperionConstants.HYPERION_PORT
```
DominicOram: Nit: This would now make more sense being called `WATCHDOG_PORT` rather than `HYPERION_PORT`

src/mx_bluesky/hyperion/runner.py
```diff
@@ -1,61 +1,15 @@
...
-
-def make_error_status_and_message(exception: Exception):
-    return ErrorStatusAndMessage(
-        status=Status.FAILED.value,
-        message=repr(exception),
-        exception_type=type(exception).__name__,
-    )
-
-
 class BaseRunner:
```
DominicOram: Could: Now that `PlanRunner` is the only implementation of this we can merge the two

PR #1628, opened 2026-02-12
Fix system tests caused by zocalo and smargon changes

olliesilvester:
Should fix the two test failures seen in https://gitlab.diamond.ac.uk/MX-GDA/hyperion-system-testing/-/jobs/319770

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


rtuck99 (approved):
Still get the test failure even with #1627 merged, will approve this and investigate why that is still happening

Issue #1208, opened 2025-08-20
Properly configure debug logging for plans running with BlueAPI


olliesilvester:
The BlueAPI helm chart currently has some [simple configuration](https://github.com/DiamondLightSource/blueapi/blob/08334bc62e4b7ea19219b34e0b109d7bfa66aed9/helm/blueapi/values.yaml#L209) for logging. BlueAPI configures the root logger, which is then automatically used by our dodal+bluesky+ophyd logs.

Our debug logs are too noisy to be sending to graylog, so we need a way to save our debug logs to a file when there's an exception.
Using k8's for logs also isn't suitable since it deletes them after 2 pod restarts.

Proposed short term solution:
- Run dodal's `set_up_DEBUG_memory_handler` in our blueapi plan entry-points
- In ixx-values.yaml, mount the debug logging location to a PVC
This will effectively give us the same behaviour as Hyperion has

Longer-term solution:
See https://github.com/DiamondLightSource/blueapi/issues/1170#issuecomment-3206339533

## Acceptance Criteria
- All of our debug logs are retrievable when running BlueAPI

rtuck99:
#1630 Should address this at least for non-k8s deployment

Issue #1213, opened 2025-08-20
Create configurable logger which is mounted to a PVC


olliesilvester:
It will be useful to have a configurable logger in BlueAPI which can be saved to-file in a PVC.

The use-case for me is for debug logs which I don't want to send to graylog as the debug logs of bluesky/ophyd are too loud. 

MX currently write debug logs to a file which overwrites itself once it gets to a certain size, as well as saves the last hour of output to file if an exception has been raised. If these files could get written to a PVC, we can keep our behaviour the same

## Acceptance Criteria
- Configurable logger added which writes to file in a mounted PVC
- Can optionally write to STDOUT, graylog, and to file at different levels

gilesknap:
I agree with @callumforrester original comment. We should get Graylog to do this rather than making our own custom thing. 

Spend the effort working on how to provide the retention you need in GrayLog. I'm no expert so I can't say for sure that it can do what you want but I'd be surprised if it cannot.

PR #1630, opened 2026-02-13
debug and file log handler support for `hyperion-blueapi`

### LINKED ISSUE #1377: Debug, file logging handlers for hyperion-blueapi
Currently non-blueapi Hyperion supports the following logging handlers:
* Stream handler (stderr)
* Graylog
* File-based handler to the configured logging directory
* Debug logging handler, that captures debug logs continuously and dumps the last 20k lines of logs whenever there is an ERROR severity message

Currently OOTB blueapi only provides the first two logging handlers.

If deployed in kubernetes, we may not need to have file-based logging as k8s will by default use stream-based logging. However we may still want the debug logger.

## Acceptance Criteria
- Decide whether we want debug and/or file logging, and if so that hyperion-blueapi supports them

rtuck99:
Fixes #1377

Adds debug and file log handler support for `hyperion-blueapi`

The behaviour of LOG_DIR and DEBUG_LOG_DIR has changed slightly - previously if not specified then the log file location would be calculated from the `BEAMLINE` environment variable as `/dls_sw/<beamline>logs/bluesky` and `/dls/tmp/<beamline>/logs/bluesky` respectively, now it requires both environment variables to be specified explicitly.

The `run_hyperion.sh` launcher script has been updated accordingly which should result in no change in normal usage, however this should reduce the potential to accidentally write to the wrong file locations in test.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


DominicOram (changes requested):
Thanks, one main comment. Additionally, the original issue says:

> Decide whether we want debug and/or file logging, and if so that hyperion-blueapi supports them

Was the conversation had about dropping some of this support? I'm not sure how much the (non-debug) file is used given it's all in graylog anyway?

Also, does this close https://github.com/DiamondLightSource/mx-bluesky/issues/1208? I guess it also relates to https://github.com/DiamondLightSource/mx-bluesky/issues/1213?

PR #1642, opened 2026-02-25
Fix system tests failing due to cached device references

rtuck99:
Fixes system tests failing due to soak tests detecting a device leak.

This was due to a change introduced in 
* DiamondLightSource/ophyd-async#1188

which caches type hints, however in cases where the device has derived signals, _derived_signal.py calls `cached_get_type_hints` with a function reference that may be bound to a device instance, thus extending its lifetime.

1. System tests pass
https://gitlab.diamond.ac.uk/MX-GDA/hyperion-system-testing/-/jobs/326333
2. 
### Checks for reviewer


coretl:
> What I found was that changing the call to cached_get_type_hints() back to get_type_hints() in _derived_signal.py _get_return_datatype() fixed the soak tests.

Quite happy for this change to be made to ophyd-async. Would you like to make a PR?

Issue #1611, opened 2026-02-05
i04 XRC shouldn't be changing aperture based on crystal size


olliesilvester:
i04's XRC plan currently changes the aperture based on crystal size. David doesn't want it to do this, we should just leave it in the "large" position to which we set at the start of the plan

Issue #561, opened 2024-10-10
 Multipin, GUP does not set aperture for each centre


rtuck99:
After the move-to-centre operation was delegated to the `multi_rotation_scan` plan, nothing in the `load_centre_collect` plan performs the aperture set.

rtuck99:
Behaviour required: 
Single pin - If crystal size > 1 box in x-dimension, then large aperture, otherwise medium
Multipin - always use large aperture.

rtuck99:
See also 
* #760

PR #1644, opened 2026-02-26
Dont change aperture during i04 xrc

jacob720:
Fixes 
- #1611 

1. Confirm aperture no longer moved after XRC result
2. New test passes and makes sense

### Checks for reviewer

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -802,8 +802,8 @@ wheels = [
 
 [[package]]
 name = "dls-dodal"
-version = "2.1.1.dev12+g7913c366f"
-source = { git = "https://github.com/DiamondLightSource/dodal.git?rev=main#7913c366f503780ada9fc189179b5c626932f94d" }
+version = "2.1.1.dev15+gb6927cc54"
+source = { git = "https://github.com/DiamondLightSource/dodal.git?rev=main#b6927cc548cc0474f30328c53c9464fa51193366" }
 dependencies = [
     { name = "aiofiles", marker = "platform_machine == 'x86_64' and sys_platform == 'linux'" },
     { name = "aiohttp", marker = "platform_machine == 'x86_64' and sys_platform == 'linux'" },


src/mx_bluesky/common/experiment_plans/change_aperture_then_move_plan.py
```diff
@@ -17,33 +21,49 @@
...
+    parameters: SpecifiedThreeDGridScan,
+    flyscan_event_handler: XRayCentreEventHandler,
+):
+    flyscan_results = yield from _get_xrc_results(
+        composite.zocalo, parameters, flyscan_event_handler
+    )
+    yield from move_to_xtal(flyscan_results[0], composite.gonio)
+
+
+def get_results_then_change_aperture_and_move_to_xtal(
```
olliesilvester: Now that Rob has removed a bunch of old Hyperion code and i04 don't want this behaviour, we never use this anymore. I think we should remove it along with `_set_aperture_for_bbox_mm`. 

It might be useful at some point, but I think best to remove as it isn't going to happen anytime soon

olliesilvester (changes requested):
Thanks! A few minor comments.

I actually think David wants large aperture to be left in after tidy-up even with udc=false, but I've asked him to clarify

src/mx_bluesky/common/experiment_plans/change_aperture_then_move_plan.py
olliesilvester: Yep, seems like we haven't recorded the reason as to why this isn't being used in Hyperion.

Please could you add a comment on top of the function, something like
`# Currently not being used, but see https://github.com/DiamondLightSource/mx-bluesky/issues/561`

tests/unit_tests/beamlines/i04/test_i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -585,3 +587,69 @@ def test_detect_grid_and_do_gridscan_gives_params_specified_grid(
...
+    mock_get_xrc_results.return_value = fake_get_xrc_results_plan()
+
+    mock_create_flyscan_params.return_value = test_fgs_params
+    run_engine(
+        i04_grid_detect_then_xrc_default_params(
+            udc=False,
+        )
+    )
+
+    assert get_mock_put(aperture_scatterguard.selected_aperture).call_args_list == [
```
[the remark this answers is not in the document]
jacob720: David said he does want to move it out at the end of the plan, as whatever happens next it will need to be moved out first

src/mx_bluesky/Getting started.ipynb
```diff
@@ -163,7 +169,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.11"
```
olliesilvester: Did you mean to change this file?

Issue #1679, opened 2026-03-18
Hyperion 1.6.5 testing


rtuck99:
Items to test for next i03 deployment:
* #1669 
* #1362 
* Latest config server changes
* DiamondLightSource/dodal#1951
* DiamondLightSource/dodal#1972
* #1598 
* #1659 - FIXED
* #1662 - FIXED
* DiamondLightSource/dodal#1958 / #1669
* #1675 - FIXED
* #1498 - FIXED
* #1601

## Acceptance Criteria
- Functionality as detailed in above PRs / GH issues is working

rtuck99:
Possible race condition in AndStatus?

```
Traceback (most recent call last):
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 622, in contingency_wrapper
    ret = yield from plan
          ^^^^^^^^^^^^^^^
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/src/mx_bluesky/common/device_setup_plans/utils.py", line 44, in wrapped_plan
    yield from plan_to_run
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/bluesky/utils/__init__.py", line 1279, in dec_inner
    return (yield from plan)
            ^^^^^^^^^^^^^^^
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/src/mx_bluesky/common/preprocessors/preprocessors.py", line 117, in pause_xbpm_feedback_during_collection_at_desired_transmission_wrapper
    yield from bpp.contingency_wrapper(
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 622, in contingency_wrapper
    ret = yield from plan
          ^^^^^^^^^^^^^^^
  File "/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/bluesky/preprocessors.py", line 188, in plan_mutator
    raise ex
[...]

PR #1687, opened 2026-03-30
Mx bluesky 1679 misc fixes from testing 1 6 5

rtuck99:
Fixes misc issues from #1679

Addresses the following issues uncovered during testing:
* Synchrotron beam dump countdown is not ignored if commissioning mode is on
* If there is no visit when UDC ends (if for example UDC default state failed) then the supervisor exits
* _supervisor.py doesn't return current_visit in decode_and_execute, so final robot unload is recorded with no visit.
* assertion in ispyb_callback_base `isinstance(SynchrotronMode)` fails
* Supervisor failed to raise an alert while waiting for feedback, due to dict value type issue

1. Tests pass
2. Functionality as described

### Checks for reviewer


jacob720 (approved):
Looks good

Issue #1688, opened 2026-03-30
Hyperion supervisor exits if cannot connect to blueapi


rtuck99:
During testing of hyperion, the following exception was raised and caused the supervisor to exit when it could not reach blueapi:

```
  "message": "[31m[E 260319 14:35:13.560 run_engine:1755] Run aborted\u001b[0m
Traceback (most recent call last):
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/connection.py\", line 204, in _new_conn
    sock = connection.create_connection(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/util/connection.py\", line 85, in create_connection
    raise err
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/util/connection.py\", line 73, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py\", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py\", line 493, in _make_request
    conn.request(
  File \"/dls_sw/i03/software/bluesky/mx-bluesky_1.6.5/mx-bluesky/.venv/lib/python3.11/site-packages/urllib3/connection.py\", line 500, in request
    self.endheaders()
  File \"/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/http/client.py\", line 1278, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File \"/dls_sw/apps/python/miniforge/4.10.0-0/envs/python3.11/lib/python3.11/http/client.py\", line 1038, in _send_output
    self.send(msg)
[...]

Issue #1660, opened 2026-03-05
Hyperion should be more tolerant of transient, retryable HTTP errors when calling services


rtuck99:
Hyperion should be more tolerant of errors that it can retry, such as this incident calling agamemnon:

https://diamondlightsource.slack.com/archives/C017UDM8R5W/p1772696004966989

We should go through and look at our calls to expeye, agamemnon etc and try and make them more robust rather than causing collection to fail straight away.

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

PR #1690, opened 2026-04-01
Add retry, error handling for agamemnon and blueapi requests

rtuck99:
Fixes
* #1688 

Also partially addresses 
* #1660 

On encountering a connection error when making a request to `hyperion-blueapi`, `hyperion-supervisor` should now attempt a further 2 more retries before giving up.

Other errors or failure after exhausting retries should cause `hyperion-supervisor` to end UDC and release the baton, raise an error alert but not exit.

Similarly receiving a timeout error, 50x HTTP error or a connection error when requesting the next instruction from agamemnon, should result in up to 2 more retries, any other error or exhaustion of the retries should result in `hyperion-supervisor` releasing the baton with and raise an error alert but remaining running awaiting the baton.

1. Tests pass
2. Behaviour is as described above

### Checks for reviewer


src/mx_bluesky/common/external_interaction/alerting/_service.py
```diff
@@ -47,6 +49,20 @@ def raise_alert(
         """
         pass
 
+    def raise_error_alert(self, content: str, metadata: dict[Metadata, str]):
+        """
+        Raise an alert that will be forwarded to beamline staff and EHC controllers for out-of-hours
```
CoePaul, line 54: This sentence would be shorter in the form 
"Alert beamline staff and EHC controllers ... support."

( We can see from the method name it achieves this by raising something )

src/mx_bluesky/common/external_interaction/alerting/_service.py
```diff
@@ -47,6 +49,20 @@ def raise_alert(
...
+            support.
+        Args:
+            content: Plain text content detailing the nature of the incident.
```
CoePaul, line 57: Plain text details of the incident
would be a shorter explanation

src/mx_bluesky/common/external_interaction/alerting/_service.py
```diff
@@ -47,6 +49,20 @@ def raise_alert(
...
+            metadata: A dict of strings that can be included as metadata in the alert for
```
CoePaul, line 58: Metadata dict (of strings) included with the alert, for the benefit of backends which support its use.

would be shorter

src/mx_bluesky/common/external_interaction/alerting/_service.py
```diff
@@ -47,6 +49,20 @@ def raise_alert(
...
+                those backends that support it. The summary and content will be included
+                by default.
+
+        """
+        beamline = get_beamline_name("")
```
CoePaul: Reading the dodal getter for beamline name - passing it an empty string 
as the default is going to leave us without a beamline name, but suppresses errors 

I assume that is intended - it seems odd that the getter doesn't just define an empty string 
if everyone in practice is just going to step around the potential warning / error anyway

CoePaul: given line 64 it's tempting to use "UNSPECIFIED BEAMLINE" or similar ... ?

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -72,8 +82,39 @@ def _instruction_and_data(agamemnon_instruction: dict) -> tuple[str, Any]:
 
 
 def _get_parameters_from_url(url: str) -> dict:
-    response = requests.get(url, headers={"Accept": "application/json"})
-    response.raise_for_status()
+    tries, delay = MAX_TRIES, RETRY_INITIAL_DELAY_S
```
CoePaul, line 88: really - could these not have been done on separate lines - 
or does ruff ( or similar ) keep putting them back onto one line?
one line - one action is always reader friendlier IMHO

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -72,8 +82,39 @@ def _instruction_and_data(agamemnon_instruction: dict) -> tuple[str, Any]:
...
+        except ConnectionError as e:
+            LOGGER.warning(
+                f"Connection error attempting to connect to agamemnon, retries left {tries}",
+                exc_info=e,
+            )
+        except Timeout:
+            LOGGER.warning(
+                f"Timed out attempting to connect to agamemnon, retries left {tries}"
+            )
+        if tries:
```
CoePaul, line 114: a bit dangerous - if tries ever makes it through zero this will not behave as You hoped

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -72,8 +82,39 @@ def _instruction_and_data(agamemnon_instruction: dict) -> tuple[str, Any]:
...
+            time.sleep(delay)  # noqa
+            delay *= 2
```
CoePaul, line 116: this delay should be capped somewhere - rather than just blindly doubling ( ad infinitum )

delay = min (delay*2,ceiling_value)

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -72,8 +82,39 @@ def _instruction_and_data(agamemnon_instruction: dict) -> tuple[str, Any]:
...
+    tries, delay = MAX_TRIES, RETRY_INITIAL_DELAY_S
+    while tries > 0:
+        tries -= 1
+        try:
+            response = requests.get(url, headers={"Accept": "application/json"})
+            try:
+                response.raise_for_status()
+                break
+            except HTTPError as e:
+                if 500 <= response.status_code < 600:
```
CoePaul: if there were a method named something like 
is_server_error(http_error_code):
    return 500 <= http_error_code < 600

this would be clearer - otherwise every poor future reader has to go look up what the 500s block of codes are for ...

yes I know the word "server" appears on line 96 - but by the time You read that it is too late

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
```diff
@@ -72,8 +82,39 @@ def _instruction_and_data(agamemnon_instruction: dict) -> tuple[str, Any]:
...
+            )
+        except Timeout:
+            LOGGER.warning(
+                f"Timed out attempting to connect to agamemnon, retries left {tries}"
+            )
+        if tries:
+            time.sleep(delay)  # noqa
+            delay *= 2
+    else:
+        msg = f"Unable to fetch instruction from agamemnon after {MAX_TRIES} attempts, ending UDC."
```
CoePaul, line 118: Deeply nit: this msg could be conditionally singular if MAX_TRIES == 1
attempts -> attempt

src/mx_bluesky/hyperion/supervisor/_supervisor.py
```diff
@@ -125,9 +131,38 @@ def shutdown(self):
...
+                        task_status = self.blueapi_client.run_task(
+                            task_request, on_event=task_monitor.on_blueapi_event
+                        )
+                        break
+                    except ServiceUnavailableError as e:
+                        LOGGER.warning(
+                            "Could not connect to blueapi client.", exc_info=e
+                        )
+                        time.sleep(delay)  # noqa
+                        delay += delay
```
CoePaul, line 147: Here You double delay by adding it to itself 
in a different place You used *= 2 

please choose one way to do it and stick to that across the whole update

CoePaul: both ways are fine with me - though in more traditional languages with integers I often go for <<= 1
just for unnecessary speed - but I wouldn't trust python to play nicely with that operator ( even if it does )

src/mx_bluesky/hyperion/supervisor/_supervisor.py
```diff
@@ -125,9 +131,38 @@ def shutdown(self):
...
+                    except BlueskyStreamingError:
+                        raise
+                    except Exception as e:
+                        get_alerting_service().raise_error_alert(
+                            "Unexpected error communicating with hyperion-blueapi", {}
+                        )
+                        raise PlanError(
+                            "Unexpected error communicating with hyperion-blueapi"
```
CoePaul: could this string not be defined on line 150 and a half and then used twice as a local variable ?

src/mx_bluesky/hyperion/supervisor/_supervisor.py
```diff
@@ -125,9 +131,38 @@ def shutdown(self):
...
+                        ) from e
+                else:
+                    LOGGER.error("Max retries reached, ending UDC.")
+                    get_alerting_service().raise_error_alert(
+                        "hyperion-supervisor stopped UDC because unable to connect to hyperion-blueapi.",
```
CoePaul: UDC was stopped because h-s was unable ... is slightly clearer 
- here the sentence needs reading 2 or 3 times to establish subject object (and thence the meaning)

src/mx_bluesky/hyperion/supervisor/_supervisor.py
```diff
@@ -125,9 +131,38 @@ def shutdown(self):
...
+                        {},
+                    )
+                    raise PlanError(
+                        f"Unable to connect to hyperion-blueapi after {MAX_TRIES} attempts, ending UDC"
```
CoePaul, line 161: see deep nit pick comment about plurals above - 
hey you could even make a function or something

tests/unit_tests/common/external_interaction/alerting/alerting/test_alerting.py
```diff
@@ -84,3 +84,25 @@ def test_logging_alerting_service_raises_a_log_message_with_additional_metadata_
             Metadata.VISIT: "cm14451-2",
         },
     )
+
+
+@patch("mx_bluesky.common.external_interaction.alerting.log_based_service.LOGGER")
+def test_raise_error_alert_delegates_to_raise_alert(mock_logger: MagicMock):
+    set_alerting_service(LoggingAlertService(CONST.GRAYLOG_STREAM_ID, WARNING))
+    get_alerting_service().raise_error_alert(
+        "Test message", {Metadata.SAMPLE_ID: "123456", Metadata.VISIT: "cm14451-2"}
```
CoePaul, line 93: don't we usually either 
a) write some parameterised thing and pass in a semi-pointless bunch of sample_ID and visit examples
or 

( and b is strongly preferred ) 

b) pass in a mock string for each payload, give the mocks enough string-ness to survive the journey and prove those mocks appear in the delegating call 

I thought that's mostly what mocks were for ( in tests of delegation )

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+def test_create_parameters_from_agamemnon_retries_on_timeout_error(
+    mock_requests: MagicMock,
+):
+    mock_requests.get.side_effect = Timeout()
+    with pytest.raises(PlanError, match="Unable to fetch instruction from agamemnon"):
+        create_parameters_from_agamemnon()
+    mock_requests.get.assert_has_calls(
+        [
+            call(
+                "http://agamemnon.diamond.ac.uk/getnextcollect/i03",
```
CoePaul, line 456: it's not clear to me from the test 
why i03 is going to be the obvious default ...

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+    mock_requests.get.side_effect = ConnectionError()
+    parent = MagicMock()
+    parent.attach_mock(mock_sleep, "sleep")
+    parent.attach_mock(mock_requests, "requests")
+    with pytest.raises(PlanError, match="Unable to fetch instruction from agamemnon"):
+        create_parameters_from_agamemnon()
+    parent.assert_has_calls(
+        [
+            call.requests.get(
+                "http://agamemnon.diamond.ac.uk/getnextcollect/i03",
```
CoePaul, line 479: see - not clear why i03 comment above and repeat for similar lines below

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+                headers={"Accept": "application/json"},
+            ),
+        ]
+    )
+
+
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.time.sleep")
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.requests.get")
+def test_create_parameters_from_agamemnon_retries_on_500_error(
```
CoePaul, line 498: Nit: Naively I thought it was called error 500 rather than 500 error - shrug

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+    response.status_code = 500
+    response.raise_for_status.side_effect = HTTPError("Test 500 error", response)
+    mock_requests_get.return_value = response
+    parent = MagicMock()
+    parent.attach_mock(mock_sleep, "sleep")
+    parent.requests.attach_mock(mock_requests_get, "get")
+    parent.attach_mock(response, "response")
+    with pytest.raises(PlanError, match="Unable to fetch instruction from agamemnon"):
+        create_parameters_from_agamemnon()
+    parent.assert_has_calls(
```
CoePaul, line 513: I'm surprised that we only see sleeps of 2 and 4 
I don't think I know what "retries" means - if retries = 3 I would expect a grand total of four calls

CoePaul: I'm also not sure we should bake in a brittle test to trip wire the hard wired "3 retries" magic number.

Why should tests fail just because someone in the future sets retries = 5 ?
or is this a subtle joke from the Holy Grail?

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.time.sleep")
+@patch("mx_bluesky.hyperion.external_interaction.agamemnon.requests.get")
+def test_create_parameters_from_agamemnon_fails_on_40x_error_and_ends_udc(
+    mock_requests_get: MagicMock,
+    mock_sleep: MagicMock,
+    mock_alert_service: MagicMock,
+):
+    response = Mock(spec=Response)
+    response.status_code = 400
```
CoePaul, line 547: if we really mean 40x - could we not have a parameterised test that racks up inputs 400 to 409 inclusive ?

CoePaul: it could skip a few but at least maybe cover 3 or 4 of 'em

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
```diff
@@ -417,3 +419,122 @@ def test_create_parameters_from_agamemnon_creates_wait(agamemnon_response):
...
+    parent.attach_mock(mock_sleep, "sleep")
+    parent.requests.attach_mock(mock_requests_get, "get")
+    with pytest.raises(
+        PlanError, match="Agamemnon returned unexpected HTTP response status code 400"
+    ):
+        create_parameters_from_agamemnon()
+    mock_requests_get.assert_called_once()
+    mock_sleep.assert_not_called()
+    mock_alert_service.raise_error_alert.assert_called_once_with(
+        "Agamemnon returned unexpected HTTP response status code 400", {}
```
CoePaul, line 560: then this last string would be a bit brittle / need a tweak

It seems inconsistent to name the test as if it were quite general ( in a ten fold sense )
and then actually only write it to very definitely only cover 1 scenario

we could have just named it 400 instead of 40x and then it would all be internally clear

tests/unit_tests/hyperion/supervisor/test_supervisor.py
```diff
@@ -380,3 +389,68 @@ def test_sample_error_skips_subsequent_instructions(
...
+    parent = MagicMock()
+    parent.attach_mock(mock_sleep, "sleep")
+    parent.attach_mock(mock_blueapi_client.run_task, "run_task")
+    with pytest.raises(PlanError, match="Unable to connect to hyperion-blueapi"):
+        runner.context.run_engine(
+            runner.decode_and_execute(TEST_VISIT, [external_load_centre_collect_params])
+        )
+    parent.assert_has_calls(
+        [
+            call.run_task(ANY, on_event=ANY),
```
CoePaul, line 412: see general brittle ness comments for similar tests

tests/unit_tests/hyperion/supervisor/test_supervisor.py
```diff
@@ -380,3 +389,68 @@ def test_sample_error_skips_subsequent_instructions(
...
+        BlueskyRemoteControlError(),
+        NoContentError(WorkerTask),
+        UnauthorisedAccessError(),
+        UnknownPlanError(),
+        InvalidParametersError([]),
+        BlueskyRequestError(422, "Test message"),
+    ],
+)
+@patch("mx_bluesky.hyperion.supervisor._supervisor.time.sleep")
+def test_supervisor_hands_back_baton_if_any_other_error(
```
CoePaul: any_other is not a clear name ( it's other in a sense only the writer understands 
... the average reader wonders "other wrt to what exactly?" )

CoePaul: line 455 suggests "Unexpected error" is clearer than "Other error"

CoePaul (changes requested):
A few suggested places for improvement

src/mx_bluesky/common/external_interaction/alerting/_service.py
rtuck99: I think I want to retain the raise an alert phrasing - it's something that might have more meaning in future - various alerting paradigms such as AlertManager have the concept of alerts as things that can be raised and continue alerting until a subsequent action is taken.

Just using "alert" as a verb is weaker and could just mean sending a message, as opposed to something more substantial.

In the long run we will probably plug in to such systems via the alerting backend.

src/mx_bluesky/common/external_interaction/alerting/_service.py
rtuck99: I think the function has changed since this was originally authored and I can now get rid of the parameter.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
rtuck99: python gives you tuple assignment, it's the python way. They are related values so it makes sense to me.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
rtuck99: Unfortunately there is not a nice function in the Python requests library that returns the category for an HTTP status code.

I can create an is_server_error function but I think it's fairly common knowledge that 50x codes are all server error codes, 200 are ok, 400 are request errors etc.

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
rtuck99: Well it's a good job we always do only 1 try at a time then

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
rtuck99: Are we ever likely to set MAX_TRIES to a large enough number to worry about this?

tests/unit_tests/common/external_interaction/alerting/alerting/test_alerting.py
rtuck99: why on earth would you mock a string?

Also I don't really see how that connects to this test - we are trying to determine that the function populates the expected metadata fields in the log record. We are not testing whether it passes through the full range of arbitrary values - that is a given.

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
rtuck99: `external_interaction` package has an `autouse` fixture in `conftest.py` that sets the beamline to i03.

In any case i03 is currently the only user of hyperion so I think it can be assumed that it is an appropriate beamline to use in tests.

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
rtuck99: That's because we specify `tries = 3` not `retries = 3`
I doubt we are going to change it TBH. Agamemnon is supposed to be a reliable service, we might see the odd network burp that merits a brief retry, but if it's so flaky that we are frequently bumping up against the max retry limit then that's a problem for agamemnon / scicomp to figure out, and I think in that case we are right to raise an alert.

This code has already saved me a callout in production!

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
rtuck99: `_get_parameters_from_url()` uses `requests` library's `response.raise_for_status()` method to separate error- from non-error statuses.

TBH with HTTP client libraries it is pretty pointless to try to test their internal error handling functions, they do all sorts of stuff that probably isn't even fully documented, like following redirects on 30x codes, retrying requests on 401 Unauthorized (because some client libraries don't like to send credentials if they are not needed). All you can do is trust that it mostly works as advertised.

tests/unit_tests/hyperion/supervisor/test_supervisor.py
rtuck99: I have renamed it to non_retryable_error

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
CoePaul: > I think it's fairly common knowledge

 - I disagree - most people have no idea which code goes with what situation and have to look them up.  That the concept exists ( of such codes ) is not obscure - what they individually represent is obscure.
It's the sort of detailed nonsense nobody should HAVE to know - that was my basic point

src/mx_bluesky/hyperion/external_interaction/agamemnon.py
CoePaul: You've been warned

src/mx_bluesky/hyperion/supervisor/_supervisor.py
CoePaul: How is this resolved ?

tests/unit_tests/common/external_interaction/alerting/alerting/test_alerting.py
CoePaul: Apparently for primitives in python nobody mocks all the primitives.
I'm pretty sure in Java unit testing if You want to know if a A calls B 
You put struts in for all the args - rather than realistic args

(Still Java context) Separately another test then checks if realistic arguments are handled correctly.
But we're in the bonkers world of python so meh

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
CoePaul: I don't think tests should ever know anything about a real URL - unless the 
test is a system level test to see if it can genuinely see that ( but then You would never want such a test in CI )

Here it is a misleading realistic URL and creates the bad smell that the test might be dependent on the real IO3 beamline for a CI suite to pass - nobody wants to have read all the details of the code learn whether or not that is true

"... i00" ( instead of "... i03") would have made that independence plausible

Something like "http:/not_a_url/not_a_beamline" - even more so

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
CoePaul: I think You've been asked two questions 
the first You answered (that retries is not the same word as tries)

the second ( that the test is brittle for no reason) was ignored and You answered questions about the production code.  My second question has nothing to do with production code - it's saying why write a test which is clearly self advertising as brittle in nature

It's perfectly possible to write a test for the same behaviour without any brittleness
else all the "saved callout time" will be payed back in "why is the CI failing" 

Avoiding brittleness should be a general and constant discipline - with the nature and content of the production code being irrelevant to that discipline

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
CoePaul: If You're determined to only ever test 400 - I don't quibble with that
I do quibble with doing only 400 inside a tin misleading labelled 40x
change the label to say 400

A bottle of sulphuric acid isn't usually labelled "some acid or other"

tests/unit_tests/hyperion/external_interaction/test_agamemnon.py
CoePaul: yeah this

CoePaul (commented):
You can read my replies - which I'm mostly letting slide 
- and anyway it looks like it's got a merge conflict to repair

You seem to take the view that a test being brittle 
can rely on the specifics of the existing production code 
(or the specifics of the use cases that the production code is coupled to) 
to rescue the test from the consequences of its own brittleness

Whereas another school of thought simply says 
test brittleness is unnecessary and should be defensively 
( defense through offense ) annihilated unless there's an extremely 
good excuse for it being unavoidable / strictly necessary

I'm basically approving once the merge conflict irons out

CoePaul (approved):
Approving - given our agreement to disagree
and that some of the requested improvements have been made

PR #1692, opened 2026-04-08
1998 fix smargon combined moves

rtuck99:
Fixes
* DiamondLightSource/dodal#1998

Requires:
* #1640

* DiamondLightSource/dodal#1999

This removes omega from `CombinedMove` because a) omega moves cannot be reliably parallelised with other axes and b) following mod-360 changes, we wish to apply a phase angle and not an absolute angle which is inconsistent with the other axes in `CombinedMove`
Instead, when we perform a combined move, the plan explicitly moves the omega axis beforehand

1. Tests pass
2. Functionality as described

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.16.0",
     "bluesky >= 1.14.6",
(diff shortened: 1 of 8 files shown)


noemifrisina (approved):
lgtm, thanks

PR #1698, opened 2026-04-16
Support configuring the config server URL

rtuck99:
This enables the config server URL to be specified in the external callbacks, which allows system tests to be fixed.

Requires:

* DiamondLightSource/dodal#2008

Adds additional documentation for environment variables.
The external callbacks must now be launched with the CONFIG_SERVER_URL parameter specified in order to start

### Instructions to reviewer on how to test:

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

PR #1699, opened 2026-04-16
Simplify config file params

rtuck99:
This simplifies setup of various config file configuration points by removing redundant points of configuration, so that it is easier to fix the system tests.

Requires:

* #1698 
* #DiamondLightSource/dodal#2008

This introduces some potentially breaking changes in the parameter model but I don't believe these parameters are actually being used anywhere and are left at their defaults.

Breaking changes:
* Remove `det_dist_to_beam_converter_path` and `insertion_prefix` from `WithVisit` 

Other changes:
* Remove `load_panda_from_yaml()` in favour of the identical dodal function
* Remove TEST_MODE from `mx_bluesky.hyperion.parameters.constants` and support for `HYPERION_TEST_MODE` env var
* Remove unused BEAMLINE and INSERTION_PREFIX from `I03Constants`
* Make `DetectorParamConstants.BEAM_XY_LUT_PATH` always use the production value instead of depending on TEST_MODE
* Remove GDA_DOMAIN_PROPERTIES_PATH

1. Tests pass

### Checks for reviewer


src/mx_bluesky/hyperion/external_interaction/callbacks/__main__.py
```diff
@@ -146,10 +145,10 @@ def setup_logging(dev_mode: bool):
 
 
 def create_config_client() -> ConfigClient:
```
jacob-williamson, line 154: Do we need this because callbacks run in a seperate process and we can't just use `get_config_client` from dodal?

jacob-williamson (approved):
Looks good, just a question

src/mx_bluesky/hyperion/external_interaction/callbacks/__main__.py
rtuck99: Yes, in the callback process, we don't have any devices or fixtures as the beamline module is not loaded, so we need some way to initialise the config client.

PR #1705, opened 2026-04-22
Set aperture according to XRC results

rtuck99:
Fixes
* #561 

For single-sample pins, the aperture is now selected according to the crystal size detected in XRC, if the crystal is larger than 1 box (20um) then the large aperture is selected, otherwise the medium aperture is selected.

The previous code in `change_aperture_then_move_plan.py` has been removed as it was never used.

For multisample pins, the large aperture is always selected.

This introduces changes to the parameter model whereby the type of `selected_aperture` is changed from `ApertureValue` (an enum defined by the dodal `ApertureScatterguard` device) to `AperturePolicy`.

These changes should be non-breaking for plans called via `blueapi`; `SMALL`, `MEDIUM` and `LARGE` values have the same json representation. However for plan implementations using this enum via `DiffractionExperiment` subclasses in the parameter model these must now explicitly map `AperturePolicy` to `ApertureValue`. There are the following changes:

* `AperturePolicy` does not have the `OUT_OF_BEAM` or `PARKED` values as these are not useful for experiment plans
* `AperturePolicy` has the additional values `AUTO` and `CURRENT_POSITION`, for plans where the aperture position is determined automatically (as in Hyperion UDC) and for plans which may opt to execute from the aperture's currently selected position (this may be useful for commissioning plans and for plans which are e.g. executed as part of GDA operation where aperture selection is managed externally)
* The default value for `selected_aperture` in `DiffractionExperiment` is `AUTO`. `None` is no longer a valid value.
* `selected_aperture` defaults in `DiffractionExperiment` are no longer overridden in subclasses. The intention is that where `AUTO` is selected, it will be resolved to one of the physical aperture sizes in the relevant part of the plan.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


phathaway (approved):
Working in production

PR #1733, opened 2026-05-07
Fix snapshot modulo

### LINKED ISSUE #1729: Snapshot modulo 180 calculation is incorrect for generated snapshots
In the snapshot callback `_smargon_omega_to_xyz_plane()` fails to evaluate the plane if the modulo produces a value slightly less than 180

Also rather than causing the collection to fail in an easily diagnosed manner, this then causes a downstream error when the callback later fails

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

rtuck99:
Fixes #1729

1. Tests pass
2. Generated grid snapshots now work if the smargon doesn't match exactly the x-y plane with omega mod(180) 

### Checks for reviewer


DominicOram (approved):
Looks good to me. Thanks.

PR #1742, opened 2026-05-11
Fix for shutter renaming

DominicOram:
Update to match https://github.com/DiamondLightSource/dodal/pull/2053

1. Confirm tests still pass on the linked dodal branch

### Checks for reviewer


src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
```diff
@@ -29,7 +29,7 @@
 from dodal.devices.undulator import UndulatorInKeV
 from dodal.devices.xbpm_feedback import XBPMFeedback
 from dodal.devices.zebra.zebra import Zebra
-from dodal.devices.zebra.zebra_controlled_shutter import ZebraShutter
+from dodal.devices.zebra.zebra_controlled_shutter import MXZebraShutter
```
CoePaul, line 32: I think MX should be Mx - but I guess here is not the place to argue that

tests/unit_tests/common/device_setup_plans/test_gonio.py
```diff
@@ -0,0 +1,34 @@
...
+from ophyd_async.core import set_mock_value
+
+from mx_bluesky.common.device_setup_plans.gonio import find_nearest_omega_360
+
+
+@pytest.mark.parametrize(
+    "current, expected",
+    [
+        # exact multiples
+        (0, 0),
```
CoePaul: if the numerical type is float rather than pure int You should have some non-integer test examples

tests/unit_tests/common/device_setup_plans/test_gonio.py
```diff
@@ -0,0 +1,34 @@
...
+        (360, 360),
+        (-360, -360),
+        # near positives
+        (10, 0),
+        (200, 360),
+        # near negatives
+        (-10, 0),
+        (-200, -360),
+        # halfway cases
+        (180, 0),
```
CoePaul: I would have made these test case close to an edge like 180.1 or 179.9 
balancing on the exact knife edge is inviting the test to go flaky and really nobody would care which way exactly 180 went ( when rounded ) in practice

src/mx_bluesky/common/device_setup_plans/gonio.py
```diff
@@ -26,3 +26,8 @@ def move_gonio_warn_on_out_of_range(
             ) from fs.__cause__
         else:
             raise fs
+
+
+def find_nearest_omega_360(gonio: XYZOmegaStage):
+    current_omega = yield from bps.rd(gonio.omega.user_readback)
+    return round(current_omega / 360) * 360
```
CoePaul: does it not make sense to put the maths into dodal ?
and test it there?

pyproject.toml
```diff
@@ -55,7 +55,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.16.0",
     "bluesky >= 1.14.6",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
+    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@cryst_bluesky_10_fast_shutter",
```
CoePaul: again we're pinning to a specific version of dodal - is that an intended side effect of this PR ?

CoePaul (commented):
Just commented on some things I noticed

src/mx_bluesky/beamlines/i04/experiment_plans/i04_grid_detect_then_xray_centre_plan.py
DominicOram: I would argue that `MX` is acronym, and as discussed in https://github.com/DiamondLightSource/dodal/pull/2033#issuecomment-4388555931, acronyms are all upper case in PEP8. I don't mind having the discussion but:
* It should be in a wider group
* If we change away from PEP8 we should then document the exception and fix it everywhere at once

src/mx_bluesky/common/device_setup_plans/gonio.py
DominicOram: Apologies, this was accidentally commited by me. Not sure why it made it made it here

tests/unit_tests/common/device_setup_plans/test_gonio.py
DominicOram: As above, this was an accidental change, I will remove

tests/unit_tests/common/device_setup_plans/test_gonio.py
DominicOram: As above, this was an accidental change, I will remove

pyproject.toml
DominicOram: See https://github.com/DiamondLightSource/mx-bluesky/pull/1743#discussion_r3228522450

rtuck99 (approved):
Approved

PR #1743, opened 2026-05-11
Fix renamed interlocks

DominicOram:
Updates to handle dodal changes https://github.com/DiamondLightSource/dodal/pull/2033

1. Confirm tests pass against new dodal branch

### Checks for reviewer

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.16.0",
     "bluesky >= 1.14.6",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
+    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@2024_decouple_interlock_from_shutter",
 ]
 
 
--- a/tests/unit_tests/beamlines/i24/conftest.py
+++ b/tests/unit_tests/beamlines/i24/conftest.py
@@ -1,3 +1,5 @@
+from unittest.mock import MagicMock
+
 import pytest
 from dodal.beamlines import i24
 from dodal.devices.beamlines.i24.aperture import Aperture
@@ -11,7 +13,6 @@ from dodal.devices.beamlines.i24.focus_mirrors import (
 )
 from dodal.devices.beamlines.i24.pmac import PMAC
 from dodal.devices.hutch_shutter import (
-    HUTCH_SAFE_FOR_OPERATIONS,
     InterlockedHutchShutter,
     ShutterDemand,
     ShutterState,
@@ -23,10 +24,7 @@ from ophyd_async.core import callback_on_mock_put, set_mock_value
 @pytest.fixture
 def shutter() -> InterlockedHutchShutter:
     shutter = i24.shutter.build(connect_immediately=True, mock=True)
-    set_mock_value(
-        shutter.interlock.status,  # type: ignore
-        HUTCH_SAFE_FOR_OPERATIONS,
-    )
+    shutter.interlock._safe_to_operate = MagicMock(return_value=True)
 
     def set_status(value: ShutterDemand, *args, **kwargs):
         value_sta = ShutterState.OPEN if value == "Open" else ShutterState.CLOSED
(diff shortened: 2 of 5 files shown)


tests/unit_tests/hyperion/experiment_plans/test_udc_default_state.py
```diff
@@ -68,7 +68,7 @@ async def default_devices(
         cryostream = OxfordCryoStream("")
         cryojet = OxfordCryoJet("")
         fluo = FluorescenceDetector("")
-        hutch_shutter = InterlockedHutchShutter("", interlock=HutchInterlock(""))
+        hutch_shutter = InterlockedHutchShutter("", interlock=PSSInterlock(""))
```
CoePaul, line 71: I'm suprised we're letting this deviation from camel case go through - given how much - validly so - effort we put in to getting standards into the codebase ( ruff etc )

pyproject.toml
```diff
@@ -55,7 +55,7 @@ dependencies = [
     "ophyd >= 1.10.5",
     "ophyd-async >= 0.16.0",
     "bluesky >= 1.14.6",
-    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@main",
+    "dls-dodal @ git+https://github.com/DiamondLightSource/dodal.git@2024_decouple_interlock_from_shutter",
```
CoePaul, line 58: I note the pin - rather than continuing to depend on main - is that a strategy against vulnerability to further renaming?

tests/unit_tests/beamlines/i24/conftest.py
```diff
@@ -23,10 +24,7 @@
 @pytest.fixture
 def shutter() -> InterlockedHutchShutter:
     shutter = i24.shutter.build(connect_immediately=True, mock=True)
-    set_mock_value(
-        shutter.interlock.status,  # type: ignore
-        HUTCH_SAFE_FOR_OPERATIONS,
-    )
+    shutter.interlock._safe_to_operate = MagicMock(return_value=True)
```
CoePaul, line 27: I naively assumed set_mock_value was a canonical way of doing things within this project - I don't understand the apparent row back to not using it - presumably it wouldn't work ... but then I don't know why it would not work: Is it the leading underscore before the method name _safe_to.. ?

CoePaul (commented):
Just commented on some things I noticed

tests/unit_tests/hyperion/experiment_plans/test_udc_default_state.py
DominicOram: As discussed https://github.com/DiamondLightSource/dodal/pull/2033#issuecomment-4388555931, this is the standard at the moment.

pyproject.toml
DominicOram: Changing the pin allows the tests to run against the correct version. If we kept the pin to `main` in this PR then the tests on this PR would fail. Normal procedure is to change this for the PR, to make sure the tests are correct then swap back to main once the `dodal` side is merged and just before merging this. (Unless that procedure is now different @rtuck99?)

tests/unit_tests/beamlines/i24/conftest.py
DominicOram: No, the problem is that we would ideally do `set_mock_value(is_safe, True)` but you can't 'cos `is_safe` is a derived signal, which doesn't support mocked values. The choice here was to either:
1. Patch out the method that the derived signal is based on (`_safe_to_operate`)
2. Set the mock values of the underlying signals that `_safe_to_operate` reads

Both are pretty ugly as it relies on knowing more logic about the device than we should need to at this level but I switched to doing 1 as with the refactor it is a layer less of introspection than going to the underlying signal

pyproject.toml
rtuck99: I am tending not to do that, at the moment - CI seems to do a good job of nagging you to fix uv.lock anyway.

There is also a pre-release step to do a uv sync which should sort out any lingering lockfile issues
https://diamondlightsource.github.io/mx-bluesky/main/developer/general/how-to/create-a-release.html

rtuck99 (approved):
Approved

Issue #1739, opened 2026-05-08
Restarting GDA server appears to kill off hyperion-blueapi


rtuck99:
During an overnight collection, GDA server was restarted and this seemed to shut down blueapi. It's not clear what the mechanism is for this. Shutting down GDA server should have no effect on blueapi.

graylog reports for hyperion-blueapi:
```
2026-05-07 19:51:16.339	i03-control.diamond.ac.uk	1101050
Shutting down
2026-05-07 19:51:16.439	i03-control.diamond.ac.uk	1101050
Waiting for application shutdown.
2026-05-07 19:51:16.440	i03-control.diamond.ac.uk	1101222
Attempting to stop worker
2026-05-07 19:51:46.454	i03-control.diamond.ac.uk	1101050
Worker did not stop within 30.0 seconds
2026-05-07 19:51:46.455	i03-control.diamond.ac.uk	1101050
Application shutdown complete.
2026-05-07 19:51:46.455	i03-control.diamond.ac.uk	1101050
Finished server process [1101050]
```
at the exact same time as I see in the GDA logs
```
2026-05-07 20:51:16,259 INFO  [server/4108907@i03-control.diamond.ac.uk] gda.util.logging.LogbackUtils - Shutting down logging
2026-05-07 20:51:16,259 DEBUG [server/4108907@i03-control.diamond.ac.uk] gda.jython.server.GdaSshServer - Shutting down SSH server
2026-05-07 20:51:16,259 INFO  [server/4108907@i03-control.diamond.ac.uk] uk.ac.diamond.daq.server.GDAServerApplication - GDA application stopping
2026-05-07 20:51:16,259 INFO  [server/4108907@i03-control.diamond.ac.uk] uk.ac.diamond.daq.concurrent.Async - Shutting down common thread pools
2026-05-07 20:51:16,259 DEBUG [server/4108907@i03-control.diamond.ac.uk] uk.ac.diamond.daq.server.GDAServerApplication - Closed status port
2026-05-07 20:51:16,259 DEBUG [server/4108907@i03-control.diamond.ac.uk] uk.ac.diamond.daq.server.GDAServerApplication - Stopping accepting status port connections
2026-05-07 20:51:16,259 INFO  [server/4108907@i03-control.diamond.ac.uk] uk.ac.diamond.daq.server.GDAServerApplication - GDA server application ended
```

## Acceptance Criteria
- Specific criteria that will be used to judge if the issue is fixed

Issue #1737, opened 2026-05-08
simplify `hyperion_restart()` for blueapi


rtuck99:
Currently `hyperion_restart()` is hotfixed for blueapi operation with options `mode="blueapi"`, `mode="udc"`, `mode="supervisor"` with `udc` being the default.

We should simplify this in order to reduce instances where beamline staff accidentally start the wrong process.

## Acceptance Criteria
* introduce a new GDA property that switches between `blueapi` and `udc` mode
* remove the parameters to `hyperion_restart()`
* remove supervisor mode, make blueapi mode restart the supervisor as well
* change `run_hyperion.sh` so that all invocations always kill all processes regardless of what kind they are.

PR #1744, opened 2026-05-11
simplify hyperion restart

rtuck99:
Fixes 
* #1737
* #1739

Also requires

* https://gerrit.diamond.ac.uk/c/gda/gda-mx/+/44987

As used from GDA, `hyperion_restart()` should always kill all hyperion processes, however separate launching of supervisor and blueapi is retained for use direct from the command line via `run_hyperion.sh`.
Blueapi is now started in a separate process group to avoid being sent the SIGINT by `gda_launcher` which was terminating the process on GDA server shutdown.

### Checks for reviewer


adaudon (approved):
Looks good. Is it not possible or unnecessary to add testing for bash scripts like this?

rtuck99:
> Looks good. Is it not possible or unnecessary to add testing for bash scripts like this?

Traditionally we haven't had unit tests for bash scripts, I suppose in theory it might be possible to test bash scripts, although I haven't looked into what frameworks exist for it. In general it's fairly complicated as we don't currently have any infrastructure for it.

Long term we are looking to move away from this though as once containerised we won't need complex bash scripts like this.

PR #1747, opened 2026-05-12
Change the process name for hyperion for better graylog output

rtuck99:
This sets the process name for Hyperion so that graylog will show it in the `process_name` field instead of the unhelpful `MainProcess`, allowing the supervisor and callbacks to be distinguished.

Most blueapi activity occurs within child worker processes, which already have their own name.

### Checks for reviewer


phathaway (approved):
Meets issue spec

PR #1755, opened 2026-05-19
Added log dir and debug log dir paths

adaudon:
Fixes small bug where log_dir was a missing environment variable so declared it before blueapi is ran.


rtuck99 (approved):
Approved

Issue #1770, opened 2026-06-16
mx-bluesky-blueapi Dockerfile installs the wrong mx-bluesky version


rtuck99:
During a release build, Dockerfile.blueapi runs

RUN uv pip install --no-cache-dir --no-compile mx-bluesky

which installs the previous version of mx-bluesky by downloading it instead of the current version from sources.

## Acceptance Criteria
- Dockerfile runs uv pip install for the current mx-bluesky

PR #1772, opened 2026-06-16
Update Dockerfile.blueapi to generate the .venv correctly

rtuck99:
* Update Dockerfile.blueapi to generate the .venv for the current build instead of downloading a previous version of mx-bluesky

Fixes 
* #1770

1. mx-bluesky-blueapi image is built with the correct dependencies as defined in the mx-bluesky uv.lock file

### Checks for reviewer


phathaway (approved):
As per run-through, looks correct

PR #1773, opened 2026-06-17
1770 fix dockerfile multiple pythons

rtuck99:
Part 2 of fix for

* #1770

This addresses

* scratch init was unable to read the .venv files
* scratch init was unable to resolve python because `uv sync` in the Dockerfile installed python 3.14 in `/home/ubuntu/.local/share/uv/python`, overwriting the already installed python 3.11

### DIFF
--- a/Dockerfile.blueapi
+++ b/Dockerfile.blueapi
@@ -34,5 +34,5 @@ USER ubuntu
 RUN --mount=type=bind,from=work,source=/work,target=/tmp/work,rw --mount=type=tmpfs,target=/tmp/uvcache \
     cd /tmp/work && \
     . /app/.venv/bin/activate && \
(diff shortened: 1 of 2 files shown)


Dockerfile.blueapi
adaudon:
```suggestion
FROM ghcr.io/diamondlightsource/blueapi:latest AS base
```

adaudon:
```suggestion
FROM base AS work
```

adaudon:
```suggestion
FROM base AS build
```

adaudon (changes requested):

adaudon (approved):
Looks good

Issue #1587, opened 2026-01-29
hyperion should honour the ROI mode specified in the LoadCentreCollect parameters instead of overwriting it with hard coded values


rtuck99:
A brief description of the issue, including specific stakeholders and the business case where appropriate

## Acceptance Criteria
- Hyperion honours `use_roi_mode` in `robot_load_then_centre` in LoadCentreCollect parameters, instead of ignoring it
- Hyperion supervisor specifies `use_roi_mode` as false in the parameters instead of true in `robot_load_then_centre` (but keeping it as true in the rotations
- The hard-coded defaults are only used if the value is not specified

PR #1776, opened 2026-06-19
1587 Fix use_roi_mode option in mx-bluesky

rtuck99:
Fixes
* #1587 

Requires:
* DiamondLightSource/daq-config-server#193

1. Hyperion honours the `gda.mx.hyperion.xrc.use_roi_mode` property in `domain.properties` for Xray centring. For rotations behaviour is unchanged and uses the full frame.

### Checks for reviewer


phathaway (approved):
Meets issue spec

Issue #1756, opened 2026-05-21
Nexus file generates wrong omega for multipin when alternating direction


rtuck99:
See /dls/i03/data/2026/mx23694-160/auto/TestInsulin/20250428-mp1601/

It seems nexus files contain omega 360 -> 720 "even" rotation sweeps instead of 360 -> 0, this manifest in xia2-dials failing for every other sample in multipin.

```
[yqf46943@ws502 20250428-mp1601]$ h5ls -d 20250428-mp1601_4.nxs/entry/sample/sample_omega
omega                    Dataset {3600}
    Data:
        (0) 360, 360.1, 360.2, 360.3, 360.4, 360.5, 360.6, 360.7, 360.8, 360.9, 361, 361.1, 361.2, 361.3, 361.4, 361.5, 361.6, 361.7, 361.8, 361.9, 362, 362.1, 362.2, 362.3, 362.4, 362.5, 362.6, 362.7, 362.8, 362.9, 363,
        (31) 363.1, 363.2, 363.3, 363.4, 363.5, 363.6, 363.7, 363.8, 363.9, 364, 364.1, 364.2, 364.3, 364.4, 364.5, 364.6, 364.7, 364.8, 364.9, 365, 365.1, 365.2, 365.3, 365.4, 365.5, 365.6, 365.7, 365.8, 365.9, 366,

```

Suspect this is due to scanspec being generated incorrectly for the even sweeps as it doesn't appear to account for rotation direction parameter. It appears that rotation likely works correctly because it does not use the scanspec to program the zebra

rtuck99:
Currently waiting on info wrt possible imginfo issues reading the generated files

rtuck99:
After discussion with Neil, have decided to run multipin collections with this and see what the fallout is wrt downstream processing and autoproc

PR #1784, opened 2026-06-29
1756 nexus generates wrong omega multipin

rtuck99:
Fixes

* #1756 

Note: system tests are currently skipped (it is currently not certain whether changes to imginfo are required or whether the warnings will be fixed at some point in future), imginfo raises warnings due to it not liking decreasing omega.

The fix should now result in the `rotation_direction` parameter in the hyperion request being applied to the omega values in the generated nexus files instead of them always increasing.

1. Nexus files are generated with correct omega metadata in forwards and backwards directions
2. ~System tests pass (when above `imginfo` fix is applied)~

### Checks for reviewer

### DIFF
--- a/tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
+++ b/tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
@@ -1445,12 +1445,18 @@ def test_full_multi_rotation_plan_nexus_files_written_correctly(
             expected_omega_starts = np.linspace(
                 scan.omega_start_deg,
                 scan.omega_start_deg
-                + ((scan.num_images - 1) * multi_params.rotation_increment_deg),
+                + (
+                    (scan.num_images - 1)
+                    * multi_params.rotation_increment_deg
+                    * scan.rotation_direction.multiplier
+                ),
                 scan.num_images,
             )
             assert np.allclose(omega, expected_omega_starts)
             expected_omega_ends = (
-                expected_omega_starts + multi_params.rotation_increment_deg
+                expected_omega_starts
+                + multi_params.rotation_increment_deg
+                * scan.rotation_direction.multiplier
             )
             assert np.allclose(omega_end, expected_omega_ends)
             assert isinstance(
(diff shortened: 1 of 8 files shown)


tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
```diff
@@ -1445,12 +1445,18 @@ def _expected_dset_number(image_number: int):
             expected_omega_starts = np.linspace(
                 scan.omega_start_deg,
                 scan.omega_start_deg
-                + ((scan.num_images - 1) * multi_params.rotation_increment_deg),
+                + (
```
CoePaul: Nit / Should : 

This has been ( originally ) written to cause the reader max double vision 

a much neater version would be something like 

assert len(omega) = scan.num_images   # same ie just to place this snippet
delta_deg = (scan.num_images - 1)
                    * multi_params.rotation_increment_deg
                    * scan.rotation_direction.multiplier
omeg_end_deg = delta_deg + scan.omega_start_deg 
expected = np.linspace(scan.omega_start_deg, omega_end_deg, scan.num_images)

tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
```diff
@@ -1445,12 +1445,18 @@ def _expected_dset_number(image_number: int):
...
+                    (scan.num_images - 1)
+                    * multi_params.rotation_increment_deg
+                    * scan.rotation_direction.multiplier
+                ),
                 scan.num_images,
             )
             assert np.allclose(omega, expected_omega_starts)
             expected_omega_ends = (
-                expected_omega_starts + multi_params.rotation_increment_deg
+                expected_omega_starts
```
CoePaul, line 1458: ditto delta comment a few lines up

tests/unit_tests/hyperion/experiment_plans/test_rotation_scan_plan.py
rtuck99: I've made it and the following slightly clearer

CoePaul (approved):

PR #1785, opened 2026-07-01
Improve the PR template

rtuck99:
Fixes:

* #N/A no specific issue

Requires:

* #N/A

Updates the PR template from the existing one to give better guidance for submitters and reviewers

_Specific advice for testing, if any_
* Are there any other things that are missing from this checklist, suggestions please.

### PR Checklist

- [ ] Would the PR title make sense to a user on a set of release notes?
- [ ] Has the documentation been updated accordingly, if appropriate?
- [ ] Docstrings added/updated on all exposed interfaces and any other significant points of interest.
- [ ] Has the PR been given an appropriate label?
- [ ] If updated versions of other dependencies are required, has pyproject.toml been updated?
- [ ] If changes to entrypoints have been made, have helm charts been updated?
- [ ] Are additional system tests appropriate and do they pass?
- [ ] If this change introduces additional dependencies, are they really necessary?

### DIFF
--- a/pull_request_template.md
+++ b/pull_request_template.md
@@ -1,13 +1,25 @@
-Fixes #ISSUE
+Fixes:
 
-Link to dodal PR (if required): #XXX
+* #ISSUE
+
+Requires:
+
+* #other PRs/Issues in this project or elsewhere which this PR depends on
+
 (remember to update `pyproject.toml` with the dodal commit tag if you need it for tests to pass!)
 
-### Instructions to reviewer on how to test:
+_Insert description of functionality added or changed, changes to any external interfaces such as
+exposed experiment plans, REST interfaces, etc_
 
-1. Do thing x
-2. Confirm thing y happens
+_Specific advice for testing, if any_
 
-### Checks for reviewer
+### PR Checklist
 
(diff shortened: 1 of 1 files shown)


pull_request_template.md
```diff
@@ -1,13 +1,25 @@
-Fixes #ISSUE
+Fixes:
+
+* #ISSUE
```
DominicOram, line 3: Obviously up to you on how you want to run your workflow but this will mean that you no longer get tickets auto-closed when related PRs are merged but maybe that's deliberate?

rtuck99: TBH I hadn't realised that was a side-effect of this change - I've been deliberately reformatting my PRs for ages like this, and had never noticed it.

rtuck99: Seems like I can either have nice formatting via tasklists, or have auto-linked issues but not both? That's a bit pants. 
I think I would rather have the nicer formatting, sometimes I have multiple PRs relating to a single issue but only the last of them fixes it.

Not sure I can be bothered to create a github action to get this functionality back, which might be a possibility of having my cake and eating it.

DominicOram: As I said, up to you, just letting you know

Issue #1765, opened 2026-06-10
box_size_um is required by GenericGrid even though x + y box dimensions can be specified independently


rtuck99:
Vmxm only run the gridscan and do not use the grid detect, however still have to populate `box_size_um` despite it not being used in gridscans.

We should probably separate out the grid specifications for grid detect and grid scan and not require the unneeded fields to be populated especially if the different axes could potentially be specified differently.

## Acceptance Criteria
- Gridscans don't need to specify box_size_um or alternatively,
- Vmxm gridscans specify only box_size_um and compute x/y dimensions from this

PR #1786, opened 2026-07-01
1765 Simplify gridscan parameter hierarchy

rtuck99:
Part 2 of a fix for #1765

Requires:

* #1787
* DiamondLightSource/dodal#2112

Below is the parameter class Hierarchy for `GenericGrid` before this refactor:
image

Here is the parameter class Hierarchy of `DiffractionExperimentWithSample` after the refactor (`GenericGrid` and related subclasses are now completely gone)
image

### A note on changes

The internal parameter model has changed significantly and is now simpler.

The `GenericGrid` and subclasses portion of the parameter hierarchy is being removed and replaced with `GridScanParams` and `GridDetectionParams`.

`detect_grid_and_do_gridscan` now only requires `DiffractionExperimentWithSample` and `GridDetectionParams`
the grid detection plan now return `GridScanParams`
`DiffractionExperimentWithSample` and `GridScanParams` is all that is needed to perform a grid scan.

The grid detection and grid scan plans are now properly genericised with type parameter `TParameter` denoting the concrete parameter type extending `DiffractionExperimentWithSample`, so that it can be passed to the BeamlineSpecific portions and also deserialized in the callbacks.

`DiffractionExperimentWithSample` hierarchy now concentrates mainly on general-purpose diffraction parameters such as exposure timings, transmission fraction, and sample metadata.

`DetectorParams` is now decoupled from the gridscan parameter model, meaning it is no longer constructed when a property on the parameter model. Instead it is explicitly constructed in the experiment plan from the parameters, and passed in explicitly into the gridscan plan and to the callbacks.

This means that many subclasses can be removed from the parameter model, and it is possible to run 2d, 3d, gridscans and grid detection in any combination without requiring modifications to the parameter model.

The rotation scan parameter model has not been changed; it still has a `detector_params` property.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
```diff
@@ -158,9 +167,21 @@ def robot_load_then_xray_centre(
...
 
+    # TODO this is no longer used in production since r_l_t_x_c is now only
+    # ever called via agamemnon, so energy is always specified
+    # https://github.com/DiamondLightSource/mx-bluesky/issues/1792
+    detector_params = (
+        create_detector_params_for_grid_scan_with_hyperion_feature_settings(parameters)
+    )
+    detector_params = yield from fill_in_energy_if_not_supplied(
+        composite.dcm, detector_params
+    )
```
DominicOram, line 178: Must: Doesn't this risk filling the detector params with the old energy before we change it below?

DominicOram (commented):
Generally this all looks a lot cleaner than before. I only noticed the one glaring thing but obviously should be looked at properly by someone in MX

src/mx_bluesky/hyperion/experiment_plans/robot_load_then_centre_plan.py
rtuck99: This should be ok, since this will only get populated if we didn't specify it, in which case it will be the current energy, and set_energy_plan won't set the energy (since it takes it from the original parameters demand_energy_ev, which will be None in this case)

In practice as the comment says we always specify an energy. This whole workflow only exists for the standalone robot load plan which is no longer exposed. I have to say I don't really like it as it makes the whole plan much more complicated, and am tempted to remove it.

PR #1787, opened 2026-07-01
1765 Refactor grid scan params

rtuck99:
Part 1 of a fix for
* #1765 

This PR introduces a refactoring of the grid scan parameters.

Currently, `GenericGrid` inherits from `DiffractionExperimentWithSample`, thus all plans use it regardless of whether they are executing grid detection or grid scans.

This is confusing for the following reasons:
* Grid detection and grid scans require different information - the former only requires broad parameters of the grid (box size and grid width and potentially also a tip offset for pin tip detection), whereas the latter requires a fully specified grid(s) with all spacings and offsets specified and for potentially multiple grids.
* GenericGrid is also used to return the detected grid parameters via `specified_grids_params`
* `SpecifiedGrids` also inherit from `GenericGrid` despite XRC potentially being called without doing grid detection by directly specifying the grid
* GenericGrid inherits from `DiffractionExperimentWithSample` which also contains information irrelevant to grid detection, such as exposure times,  apertures and other information

The purpose of this refactoring is to separate these concerns by introducing the following parameter classes:

* `GridDetectionParams` - defines the grid parameters necessary for performing grid detection
* `GridScanParams` - defines the grid parameters necessary for performing a gridscan (note that this is plan-agnostic and doesn't contain e.g. panda or zebra settings, or anything not related to the grid
* These parameter classes inherit from `BaseModel` and are intended to be used either as mixins or standalone parameter objects.

A subsequent refactoring will simplify the concrete parameter gridscan parameter classes that derive from SpecifiedGrids by decoupling them from `DiffractionExperiment`, so that they can be used as mixins into the main parameter model.

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

### DIFF
--- a/src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
+++ b/src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
@@ -212,7 +212,7 @@ def create_parameters_for_flyscan_xray_centre(
     xrc_params_type: type[SpecifiedThreeDGridScan],
 ) -> tuple[SpecifiedThreeDGridScan, GridScanParams]:
     grid_scan_params = GridScanParams(
-        omega_starts_deg=parameters.omega_starts_deg,
+        omega_starts_deg=[0, 90],
         x_start_um=grid_parameters["x_start_um"],
         y_starts_um=grid_parameters["y_starts_um"],
         z_starts_um=grid_parameters["z_starts_um"],
--- a/src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
+++ b/src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
@@ -43,7 +43,12 @@ from mx_bluesky.common.parameters.device_composites import (
     FlyScanEssentialDevices,
     GridDetectThenXRayCentreComposite,
 )
-from mx_bluesky.common.parameters.gridscan import GenericGrid, SpecifiedThreeDGridScan
+from mx_bluesky.common.parameters.gridscan import (
+    GenericGrid,
+    GridDetectionParams,
+    GridScanParams,
+    SpecifiedThreeDGridScan,
+)
 from mx_bluesky.common.utils.log import LOGGER
 
 TFlyScanEssentialDevices = TypeVar(
@@ -102,6 +107,9 @@ def detect_grid_and_do_gridscan(
     xrc_params_type: type[SpecifiedThreeDGridScan],
     construct_beamline_specific: ConstructBeamlineSpecificFeatures,
 ):
+    grid_detect_params = GridDetectionParams(
+        box_size_um=parameters.box_size_um, grid_width_um=parameters.grid_width_um
+    )
     snapshot_template = f"{parameters.detector_params.prefix}_{parameters.detector_params.run_number}_{{angle}}"
 
     grid_params_callback = GridDetectionCallback()
@@ -113,28 +121,6 @@ def detect_grid_and_do_gridscan(
         wait=True,
     )
 
-    @bpp.subs_decorator([grid_params_callback])
-    def run_grid_detection_plan(
-        oav_params,
-        snapshot_template,
-        snapshot_dir: Path,
-    ):
-        grid_detect_composite = OavGridDetectionComposite(
-            backlight=composite.backlight,
-            oav=composite.oav,
-            gonio=composite.gonio,
-            pin_tip_detection=composite.pin_tip_detection,
-        )
-
-        yield from grid_detection_plan(
-            grid_detect_composite,
-            oav_params,
-            snapshot_template,
-            str(snapshot_dir),
-            parameters.grid_width_um,
-            parameters.box_size_um,
-        )
-
     if parameters.selected_aperture:
         # Start moving the aperture/scatterguard into position without moving it in
         yield from bps.prepare(
@@ -143,10 +129,15 @@ def detect_grid_and_do_gridscan(
             group=PlanGroupCheckpointConstants.PREPARE_APERTURE,
         )
 
-    yield from run_grid_detection_plan(
-        oav_params,
-        snapshot_template,
-        parameters.snapshot_directory,
+    yield from bpp.subs_wrapper(
+        _run_grid_detection_plan(
+            composite,
+            grid_detect_params,
+            oav_params,
+            snapshot_template,
+            parameters.snapshot_directory,
+        ),
+        grid_params_callback,
     )
 
     yield from bps.abs_set(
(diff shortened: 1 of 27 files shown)


src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
```diff
@@ -95,6 +101,9 @@ def plan_to_perform():
         group=PlanGroupCheckpointConstants.GRID_READY_FOR_DC,
     )
 
+    assert grid_scan_params
```
CoePaul: using assert for runtime checks or validation is heavily discouraged in Python

CoePaul: Can't we check this instead in unit testing ?
or at least a comment explaining why we want to assert here

src/mx_bluesky/common/external_interaction/callbacks/common/grid_detection_callback.py
```diff
@@ -12,21 +12,23 @@
 
 class GridParamUpdate(TypedDict):
     """
-    Grid parameters extracted from the grid detection.
+    Generate grid parameters extracted from the grid detection for two 2D gridscans, one in xy-plane and the
+    second in the xz-plane.
+
+    Note, in this class y refers to both the y plane in absolute coordinate space, and to the secondary axis
```
CoePaul: Can we find some way to use distinct directions like u, v, w instead of ambiguous dual-use of the same letter(s).

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -44,6 +43,106 @@
...
+class GridDetectionParams(BaseModel):
+    """
+    Parameters that specify the characteristic dimensions of one or more grids to be detected.
+
+    Attributes:
+        box_size_um: The dimensions of the boxes in the grid in microns.
+        grid_width_um: The suggested width of the grid to be detected.
+    """
+
+    box_size_um: float = Field(default=GridscanParamConstants.BOX_WIDTH_UM)
```
CoePaul: should we not add checks that these are > 0 in the field validation ?

e.g. either
box_size_um: PositiveFloat = Field(default=blah)

or 
box_size_um: float = Field(gt=0.0, default=blah)

CoePaul: and that comment of mine implies we ought to add unit tests 
to prove that sending in values of 0 of minus something are rejected as expected

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -44,6 +43,106 @@
...
+
+        name_and_length = {name: len(value) for name, value in fields.items()}
+        lengths = name_and_length.values()
+        if len(set(lengths)) != 1:
+            details = "\n".join(
+                f"  {name}: length={len(value)}, value={value}"
+                for name, value in fields.items()
+            )
+
+            raise ValueError("Fields must all have the same length:\n" + details)
```
CoePaul: this is very opaque
You are checking that the string representation of the values has been padded to the same numerical representation width? I think?

the statement "Fields must all have same length" 

if I saw that as an error in a support incident - I would be checking my variable name strings and not knowing what the heck is wrong, nor why

There is probably some way to make this clearer - but I don't know how

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -44,6 +43,106 @@
...
+    @property
+    def num_grids(self):
+        return len(self.y_steps)
+
+    @property
+    def grid_specs(self) -> list[Product[str]]:
+        _grid_specs = []
+        for idx in range(self.num_grids):
+            x_end = self.x_start_um + self.x_step_size_um * (self.x_steps - 1)
+            y_end = self.y_starts_um[idx] + self.y_step_sizes_um[idx] * (
```
CoePaul: this code using idx as index is better than other examples ( higher up in this )
where 0, 1 and 2 were "hard wired" can we not do something similar to this 
in those locations?

CoePaul: also it feels like 
def terminus_of(a,b,c) 
    return a + b * (c -1 )
is warranted here or somewhere

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -44,6 +43,106 @@
...
+                self.y_steps[idx] - 1
+            )
+            grid_x = Line("sam_x", self.x_start_um, x_end, self.x_steps)
+            grid_y = Line("sam_y", self.y_starts_um[idx], y_end, self.y_steps[idx])
+            grid_z = Static("sam_z", self.z_starts_um[idx])
+            _grid_specs.append(grid_y.zip(grid_z) * ~grid_x)
```
CoePaul: this magic tilde is 'orrible - but I guess it is too late now 

maybe it could be mitigated with an explanatory comment at the end of the line 
snake alternates x direction ... or similar

Scarily enough our silicon overlords already know what this ~ means and why it is used

src/mx_bluesky/common/parameters/gridscan.py
```diff
@@ -44,6 +43,106 @@
...
+        return _grid_specs
+
+    @property
+    def scan_points(self) -> list[AxesPoints[str]]:
+        """A list of all the points in the scan_spec for each grid."""
+        _scan_points = []
+        for grid in range(self.num_grids):
```
CoePaul: it feels like there's a python comprehension way of doing this more succinctly

src/mx_bluesky/hyperion/parameters/gridscan.py
```diff
@@ -38,53 +40,55 @@ def detector_params(self):
...
+    """During 3D grid scans, there is an omega rotation before the second grid,
+    transforming Y -> Z axes, so use the second element of the Y params to set
+    Z params on the 3D grid scan device.
+    """
+    return ZebraGridScanParamsThreeD(
+        x_steps=grid_scan_params.x_steps,
+        y_steps=grid_scan_params.y_steps[0],
+        z_steps=grid_scan_params.y_steps[1],
+        x_step_size_mm=grid_scan_params.x_step_size_um / 1000,
+        y_step_size_mm=grid_scan_params.y_step_sizes_um[0] / 1000,
```
CoePaul: we have in general maths in dodal common some conversion functions for this - we should be using them

tests/unit_tests/common/experiment_plans/test_common_flyscan_xray_centre_plan.py
```diff
@@ -336,6 +368,8 @@ def test_when_grid_scan_ran_then_eiger_disarmed_before_zocalo_end(
     ):
         test_three_d_grid_params.x_steps = 9
         test_three_d_grid_params.y_steps = [10, 12]
+        grid_scan_params.x_steps = 9
```
CoePaul: got to about here

src/mx_bluesky/common/parameters/gridscan.py
rtuck99: I think you are misinterpreting the code.
The check ensures that all of the attributes of list type are of the same length. 
First we construct a map of field name to list length
Then we take the values of that list (i.e the list of lengths) and put those values into a set.
If the set is of size != 1, then there is more than one length.
There is a test for all this in `test_gridscan.py`
Although, thinking about this, we could just create the set directly, perhaps that would  be clearer.

src/mx_bluesky/common/parameters/gridscan.py
rtuck99: The indices in this function relate to the index of the grid, whereas in the 0,1,2 case, the indices generally relate to how x,y,z map to x', y' (or u, v as you would have it)

src/mx_bluesky/common/parameters/gridscan.py
rtuck99: I always have to look up the API for `scanspec` whenever I read this code. It works though and it's been like this for ages. Maybe this wouldn't be quite so hard to understand if it was in a library of other similar code, as I imagine it might be one day, when we support loads of weird scan styles. Then I'd have to internalise the scanspec API any maybe it would feel more natural. The closest I can get at the moment is it's basically a drawing API for scans.

src/mx_bluesky/common/experiment_plans/common_grid_detect_then_xray_centre_plan.py
rtuck99: The assert is there because we assign `None` to `grid_scan_params` however the type-checker can't determine that grid_scan_params is always set to a non-null value. I could insert a `#  type: ignore` here although I don't think that would be any clearer. I have added a comment

src/mx_bluesky/common/external_interaction/callbacks/common/grid_detection_callback.py
rtuck99: Hmm. I think that ship may have sailed - there is a lot of code elsewhere that has already decided that x and y are the coordinates for the grid plane, without rewriting it all might be just as confusing.

tests/unit_tests/common/experiment_plans/test_common_flyscan_xray_centre_plan.py
CoePaul: if the 9,10,12 triple here for gsp doesn't have to match the 3d params then making them different might help debug any failures more rapidly ?

rtuck99:
See
* #1823 

for code changes arising from this PR review

tests/unit_tests/common/experiment_plans/test_common_flyscan_xray_centre_plan.py
rtuck99: This disappears in part 2

CoePaul (commented):
I mean it looks ok I guess - and as it is a bit opaque and only part of a bigger change 
I don't see what else to say

CoePaul (approved):

PR #1806, opened 2026-07-29
Added two parameters and a change to chip collect i24

adaudon:
Notes on some things that might also be worth fixing/changing or not, mentioned to simply confirm.
```Removed use_control_machine: bool = true from deploy_mx_bluesky.py, is that fine?

Values are different so worth checking?

cs_maker.json
DiamondExtruder-I24-py3v1.edl (beginScreenProperties x and y)
DiamondChipI24-py3v1.edl (several x and y values through the file)
pumpprobe-py3v1.edl (two x and y values)
QiamondChipI24-py3v1.edl, qumpprobe-py3v1.edl, qumpprobe-py3v2.edl, qumpprobe-py3v3.edl seem to be old files that are not needed anymore but maybe double check.
i24/serial/setup_beamline/pv_abstract.py has different values
test_ispyb_dev_connection.py
It had focalspotsizeatsamplex and y to be missing but can't find commit that removes them

Setup_panda is now load_panda_yaml.py now it seems, just want to confirm if the refactor is true or not because I wasn't too sure.```


rtuck99 (approved):
Approved!

PR #1653, opened 2026-03-03
Ispyb integration vmxm grid scan

olliesilvester:
Add ispyb integration for vmxm grid scans

Requires 
* https://github.com/DiamondLightSource/mx-bluesky/pull/1636
* DiamondLightSource/dodal#2057

Also requires additional changes from beamline hotfixes
* DiamondLightSource/dodal#1925

### Instructions to reviewer on how to test:

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer

- [ ] Would the PR title make sense to a user on a set of release notes

Issue #1555, opened 2026-01-19
Change s4 slits to use common slits


DominicOram:
We now have a set of common, configurable slits in `dodal/devices/slits`. We should use them for the s4 slits on i03 and i04

## Acceptance Criteria
- Use the common slits

PR #1814, opened 2026-08-05
1555 convert i03, i04 to use common slits

rtuck99:
Fixes
* #1555

Also fixes a regression caused by
* #1653

where SynchrotronMode could not always be read in the event.

Requires
* DiamondLightSource/dodal#2160

1. Do thing x
2. Confirm thing y happens

### Checks for reviewer


CoePaul (approved):
Looks broadly correct

